PR #178, opened 2026-04-09
fix(vestad): fallback to export/import when docker commit fails

elyxlz:
## Summary
- When `docker commit` fails (e.g. content digest corruption), fall back to `docker export | docker import` which streams the container filesystem as a flat tarball, bypassing the content store entirely
- Fixes auto-backup failures caused by Docker storage layer corruption

## Test plan
- [ ] Trigger a backup on an agent with corrupted storage layers — should succeed via fallback
- [ ] Trigger a backup on a healthy agent — should still use `docker commit` as before


elyxlz:
Superseded by #180 which includes this fix plus 9 additional backup system improvements.

PR #233, opened 2026-04-13
ci: add vite base path check for Cloudflare

elyxlz:
## Summary
- Adds a CI check on PRs that fails if `app/vite.config.ts` has `base: "./"` instead of `base: "/"`
- Relative base path breaks Cloudflare Pages deployments on vesta.run
- Added to the merge gate so PRs can't land with the wrong value

## Test plan
- [ ] Verify the check passes on this PR (base is currently `"/"`)
- [ ] Confirm merge-gate includes the new job


emipasca:
i added a test for this already

PR #414, opened 2026-04-22
fix(vestad): disable TLS 1.3 post-handshake tickets to unbreak LAN clients

### LINKED ISSUE #341: vestad TLS renegotiation breaks non-curl clients on LAN
## Problem

vestad's HTTPS listener triggers a mid-handshake TLS renegotiation, which blocks most modern HTTP clients from connecting on the LAN.

## Symptoms

From a Windows client (`curl.exe 8.18.0` via schannel) connecting to `https://<host>:<vestad-port>/`:

```
* Trying 192.168.4.5:41343...
* schannel: disabled automatic use of client certificate
* schannel: using IP address, SNI is not supported by OS.
* ALPN: curl offers http/1.1
* ALPN: server accepted http/1.1
* Established connection to 192.168.4.5
> GET / HTTP/1.1
* Request completely sent off
* schannel: remote party requests renegotiation
* schannel: renegotiating SSL/TLS connection
* schannel: SSL/TLS connection renegotiated
< HTTP/1.1 404 Not Found
```

curl tolerates renegotiation and gets through. The Vesta desktop/GUI app on the same machine cannot reach the host with the same URL ("could not reach server"), presumably because the embedded HTTP client (Electron/Node or similar) rejects renegotiation by default, as do TLS 1.3 clients generally.

## Environment

- okami container, vestad v0.1.140 (container code still at 0.1.134, host daemon at 0.1.140)
- Two clients on the same LAN 192.168.4.0/22
- TCP reachable (`Test-NetConnection` succeeds)
- No firewall on host, confirmed via `ufw`, `nft`, `iptables`
- Tunnel via `mamba-okami.vesta.run` works fine (no renegotiation there, Cloudflare terminates TLS)

## Why it happens (suspected)

Looks like vestad may be initiating a post-handshake client certificate request (for mTLS or some per-request auth context), which triggers renegotiation on TLS 1.2 clients and outright failure on strict/TLS 1.3 clients.

## Ask

- Avoid TLS renegotiation entirely. If client-cert auth is wanted, request it in the initial handshake via `CertificateRequest` in the first ServerHello, not via mid-stream renegotiation.
- Or drop per-request cert auth on the listener and rely on bearer-token (`X-Agent-Token`) which is already in place.
- Make sure the server negotiates TLS 1.3 cleanly; TLS 1.3 forbids renegotiation, so any client hitting 1.3 will fail if vestad tries it.

## Workaround today

Use the Cloudflare tunnel URL (`https://<agent>.vesta.run`) instead of the direct LAN IP. Fine for us, bad for anyone who wants offline/no-tunnel LAN access.

elyxlz:
## Summary
- `vestad`'s HTTPS listener was breaking direct-LAN connections from the desktop app's webview (WebView2 / schannel). curl worked with `remote party requests renegotiation` visible in its verbose log; WebView2 couldn't recover and reported "could not reach server". Cloudflare-tunneled URLs were fine because Cloudflare terminates TLS.
- Root cause: rustls' default `ServerConfig` sends post-handshake TLS 1.3 `NewSessionTicket` messages, which schannel interprets as a renegotiation request. TLS 1.3 forbids renegotiation, so strict clients drop the connection.
- Switched to a hand-built `rustls::ServerConfig` with `send_tls13_tickets = 0`, explicit `with_no_client_auth()`, and ALPN `h2 + http/1.1` in the initial ServerHello. Newly-generated self-signed certs also get `EKU=ServerAuth` + `KeyUsage=DigitalSignature,KeyEncipherment` so strict clients don't infer mTLS capability from an extension-less cert.
- Existing certs on disk are reused as-is so current fingerprint pinning keeps working; only new cert generations pick up the EKU/KeyUsage.

## Test plan
- [x] `cargo clippy -p vestad -- -D warnings`
- [x] `cargo test -p vestad --bin vestad` (includes new `tls_config_tests::disables_tls13_tickets_and_sets_alpn`)
- [ ] Desktop app connects to a LAN-hosted vestad over HTTPS without the "could not reach server" failure
- [ ] `curl.exe` (schannel) verbose output no longer shows `remote party requests renegotiation`
- [ ] Cloudflare tunnel URL still works end-to-end

Fixes #341

### DIFF
--- a/vestad/Cargo.lock
+++ b/vestad/Cargo.lock
@@ -1850,6 +1850,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustls-pemfile"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dce314e5fee3f39953d46bb63bb8a46d40c2f8fb7cc5a3b6cab2bde9721d6e50"
+dependencies = [
+ "rustls-pki-types",
+]
+
+[[package]]
 name = "rustls-pki-types"
 version = "1.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2558,6 +2567,7 @@ dependencies = [
  "ring",
  "rust-embed",
  "rustls",
+ "rustls-pemfile",
  "self-replace",
  "serde",
  "serde_json",
--- a/vestad/Cargo.toml
+++ b/vestad/Cargo.toml
@@ -25,6 +25,7 @@ tower-http = { version = "0.6", features = ["cors", "trace"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 rustls = { version = "0.23", default-features = false, features = ["ring", "logging", "std", "tls12"] }
+rustls-pemfile = "2"
 rcgen = "0.14"
 ring = "0.17"
 serde = { version = "1", features = ["derive"] }
--- a/vestad/src/serve.rs
+++ b/vestad/src/serve.rs
@@ -80,6 +80,18 @@ pub fn ensure_tls(config_dir: &std::path::Path) -> (String, String, String) {
             }
         }
     }
+    // Mark as a TLS server cert only; without EKU, some clients (schannel /
+    // WebView2) treat the cert as potentially mTLS-capable and get confused
+    // by post-handshake messages. See issue #341.
+    params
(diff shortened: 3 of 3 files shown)


elyxlz:
Closing — fix was speculative (couldn't reproduce schannel renegotiation behavior locally, no Windows/LAN test environment to validate). Will revisit once we can actually confirm the root cause.

PR #443, opened 2026-04-24
feat(vestad): authenticate dashboard iframe via per-service path key

### LINKED ISSUE #264: feat(infra): add authed dashboard service to avoid public registration
## Context

PR #260 introduced public/non-public service registration. The dashboard currently registers as `public: true` because non-public services require auth, and iframe sub-resource loads (JS/CSS) can't carry Bearer tokens or `?token=` query params.

The old approach (token injection into HTML) was removed because it was dead code — the app passes auth via `postMessage`, not URL tokens.

## Problem

Registering the dashboard as public means anyone with the tunnel URL can access it without authentication. Ideally the dashboard would be a non-public (authed) service, but we need a mechanism for iframe sub-resources to authenticate.

## Options to explore

- **Cookie-based auth**: On the first authed request to a non-public service, set an `HttpOnly` cookie scoped to that service path. Subsequent sub-resource loads from the iframe would send the cookie automatically.
- **Service worker**: The dashboard could install a service worker that intercepts fetch requests and injects the auth header (token received via `postMessage`).
- **Proxy-level session**: vestad generates a short-lived session ID for each iframe load, embedded in a cookie, and validates it on sub-resource requests.

## Acceptance criteria

- Dashboard service can be registered without `"public": true`
- Dashboard iframe loads correctly (HTML + all JS/CSS assets) with authentication
- External visitors without auth get 401

elyxlz:
## Summary

Makes the embedded agent dashboard private without relying on third-party iframe cookies.

- registers the dashboard as a private service
- gives every registered service a stable random 256-bit access key
- serves keyed service URLs at `/agents/{name}/{service}/k/{key}/...`
- carries the key through the current `/sync` state tree
- builds dashboard iframe URLs with the key in the path, so relative JS/CSS/font requests inherit authentication
- preserves the existing Bearer/query-token path for authenticated API clients
- aligns the mobile app's Expo patch dependencies with Expo 57's current compatibility requirements so the shared app CI matrix remains green

Closes #264.

## How it works

`ServiceEntry` now persists a per-service key in `settings.json`. New registrations generate a 32-byte random key, re-registrations keep the existing key, and legacy numeric/full service entries receive a key when they are loaded.

The service proxy recognizes a leading `/k/{key}` segment. A matching key authenticates the request, the proxy strips the key prefix before forwarding upstream, and both HTTP and WebSocket forwarding avoid injecting the internal agent token. A missing or incorrect key on a private service still returns `401` unless normal API authentication is present.

The current sync protocol exposes the key on `ServiceInfo`. The TypeScript field is optional so a new app can still consume an older daemon response; when a key is present, the Dashboard uses it in the iframe path. Service equality also includes the key so a rotated/migrated value reaches consumers.

## Why the key is in the path

The desktop and mobile apps load the dashboard cross-origin in a WebView, where cookie behavior is not reliable. A query parameter authenticates only the iframe document and is not inherited by relative asset requests. Because the dashboard uses a relative Vite base, a path prefix is inherited by the document and all of its subresources.

## Compatibility and security

- New app with old daemon: `key` is absent, so the app falls back to the legacy dashboard URL.
- Old app with new daemon: public services remain compatible; the newly private dashboard requires an updated client after it re-registers.
- Service-list and sync endpoints remain API/agent-token authenticated; the path key is not exposed through a public metadata endpoint.
- Random keys are persisted and kept stable across ordinary service re-registration.

## Testing

- `cargo clippy --manifest-path vestad/Cargo.toml -p vestad -- -D warnings`
- `cargo test --manifest-path vestad/Cargo.toml -p vestad --bin vestad` — 255 passed, 15 Docker-only tests ignored
- `cargo test --manifest-path vestad/Cargo.toml -p vestad api_contract_fixtures_up_to_date`
- `cargo test --manifest-path vestad/Cargo.toml -p vestad --bin vestad service_entry`
- `./check.sh app-core app-web` — 174 core tests and 179 web tests passed
- `./check.sh app-mobile` — Expo compatibility check, lint, typecheck, 111 tests, and clean native prebuild passed
- `./check.sh guards`
- Docker-gated `non_public_service_requires_path_key` coverage verifies valid key → 200, incorrect/missing key → 401, and API token fallback → 200 in CI


elyxlz:
fix merge conflicts

elyxlz:
Resolved the merge conflicts against master (which had moved a lot since this PR opened).

Key conflicts and how they were resolved:
- `vestad/src/serve.rs`: master extracted the whole settings schema into a new `vestad/src/settings.rs` module, so I moved this PR's `ServiceEntry.key` field, `gen_service_key()`, and the key-preserving deserializer into that new module, and kept the per-service key logic in `register_service_handler` (stable key across re-registrations).
- `vestad/src/agent_proxy.rs`: kept the path-key auth (`split_key_subpath`, `skip_auth = is_public || via_key`) over master's pure reformatting.
- `vestad/tests/server/mod.rs`: took master's module list and added `service_proxy_auth` in order (the PR side had stale duplicate mod lines).
- `agent/skills/dashboard/SKILL.md` and `SETUP.md`: master rewrote both (subagent dispatch + `daemon` script), which supersedes this PR's edits there. The PR's real intent, dropping `--public` from `register-service dashboard`, is already carried in `scripts/serve` (auto-merged) so the dashboard registers non-public and authenticates via the path key.

Checks run from the worktree:
- `./check.sh vestad`: pass (191 tests, clippy clean).
- `./check.sh web`: pass (167 tests; API contract fixtures regenerate with no diff, so `key`/`public:false` are in sync across Rust and TS).
- `vesta-tests` integration crate compiles (the new `service_proxy_auth` e2e is Docker-gated and will run in CI).

PR is now MERGEABLE.

🤖 Generated with Claude Code

elyxlz:
Resolved merge conflicts against current master.

**Conflict:** `agent/skills/dashboard/scripts/serve` (one line). Master renamed the `service` skill to `vestad`; this PR dropped `--public` so the dashboard is served through the new per-service path-key auth instead of a public token-free tunnel. Combined both: `register-service dashboard` on the renamed `vestad` path, no `--public`.

**Silent (semantic) merge fixups:** master added a new `ServiceEntry` test constructor (`resident_service_keeps_its_port_across_reregistration`) and updated the API-contract fixture; both needed the PR's new `key` field. Added `key` to that constructor and regenerated `cli/tests/fixtures/vestad-api.json` (dashboard sample now `public: false` + `key`). Master's `public` field and the PR's `key` field coexist cleanly.

**Checks:** `./check.sh vestad` green (203 passed), `./check.sh web` green (185 passed), `./check.sh guards` all passed. Docker-gated suites (integration, vestad-docker, live) not run here.

🤖 Generated with Claude Code
Do NOT merge this PR.

PR #537, opened 2026-05-05
refactor(agent): unified state.json, notification-driven system prompts, agent-marked completion

### LINKED ISSUE #515: fix(dream): retry rate-limited runs instead of silent skip
## Pattern

The nightly dreamer runs at ~03:00 local time. When the Anthropic API rate limit is exhausted around that time (e.g., from a long-running background dispatch chain that overlapped with the dream window), the dreamer fails silently. The next morning, no dream summary file appears in `~/agent/dreamer/`, no MEMORY.md update happens, and the agent's session is wiped without any retroactive consolidation.

## Impact

- Yesterday's work is lost from working memory: tasks reverted to old state, lessons un-captured, sensitive data not redacted.
- Compounds quickly: a skipped dream + a fresh-start session = full re-derivation of yesterday's context from WhatsApp/JSONL transcripts (expensive and lossy).

## Suggested fix

Detect a skipped dream on session resume (no `~/agent/dreamer/<yesterday>*.md` file) and either:
1. Auto-trigger a retroactive dream as the first action after services come up, OR
2. Surface a one-line warning to the user in the morning notification ("dreamer didn't run, want me to retro-dream now?")
3. Retry mechanism: if rate limit, sleep until reset window and retry the dream cycle.

Either path beats silent failure.

## Reproduction

Trigger a long Agent dispatch shortly before 03:00 that would consume rate limit, then observe that the next morning's session has no dream summary and a stale-feeling MEMORY.md.

elyxlz:
## Summary

Restructures how the agent boots and gets its system prompts. Five tightly-coupled changes:

### 1. Restart prompt moved into a `restart` skill

- `agent/prompts/restart.md` (user-mutable) replaced by `agent/skills/restart/SKILL.md`, which owns both the workflow and per-skill service-startup commands.
- `agent/core/prompts/restart.md` shrinks to a one-line pointer at the skill (mirrors `nightly_dream.md`).
- `agent/prompts/` directory + `prompts_dir` config property removed; `load_prompt` drops its fallback. Prompts live in core, customization happens via MEMORY.md and skills.

### 2. Single `state.json` replaces six legacy marker files

- New `agent/core/state_store.py` with a `PersistedState` Pydantic model + atomic tmp+rename writes.
- Consolidates `first_start_done`, `restart_reason`, `last_dreamer_run`, `show_dreamer_summary`, `session_id`, `migrations.applied` into one inspectable file.
- First boot under this code imports legacy markers and removes them; subsequent boots take the fast path.
- Corrupt or schema-incompatible `state.json` is logged and replaced with a fresh state instead of crash-looping the container.

### 3. System prompts flow through ` ~/agent/notifications/`

- Migrations, first-start setup, restart greetings, proactive checks, and the nightly dream all drop JSON files (`source: "core"`) instead of going through the in-memory ` asyncio.Queue`.
- `process_batch` partitions notifications by source: external sources keep the "reply if appropriate" suffix; system sources skip it.
- Same shape vestad already uses (per #545's rename notification): unified ingestion path, on-disk durability, observability via `ls notifications/`.

### 4. Agent-marked completion via three new MCP tools

- `mark_setup_done` — records first-start completion AND brings the WebSocket server online (the readiness signal vestad polls).
- `mark_migration_applied(name)` — records a migration as applied.
- `mark_dreamer_complete` — records today's dream, clears the session, queues the nightly restart.

If the agent never calls a `mark_*` tool (rate limit, crash, hallucinated success), the work re-runs on the next boot/tick. Migration prompts re-state idempotency framing since they may genuinely retry. **Fixes #515** — `process_nightly_memory` keeps firing past the configured hour until `last_dreamer_run.date() == today`, bounded to ` DREAMER_CATCHUP_HOURS = 6` so a silently-failed `mark_dreamer_complete` cannot preempt the agent every hour for the rest of the day.

### 5. First-start setup + greeting unified into one prompt

- Old: two-stage flow (setup → bind WS in `run_vesta` → drop greeting notification).
- New: one prompt covering both. The agent calls `mark_setup_done` mid-prompt (binds WS, satisfies vestad's readiness poll), then continues the same turn to greet via app-chat.
- Restart boots bind WS immediately at boot — no readiness-poll deadlock (preserves #540's fix).
- Fail-closed by design: if first-start crashes before `mark_setup_done`, the WS port stays unbound. A half-configured agent shouldn't look ready.

## Code-organization changes

- `agent/core/client.py` (670 lines) split into `client.py` + `diagnostics.py` + `sdk_parsing.py` + `tools.py`.
- `vm.State` methods (`touch_activity`, `sdk_idle_seconds`, `longest_running_tool`) moved to free functions in `diagnostics.py` per the "no methods on dataclasses" convention.
- `_process_message_safely` inlined as a nested function inside `_run_messages_with_interrupts`. Two module-level functions become one.
- Restart-reason constants `PROCESSOR_CANCELLED_RESTART`, `PROCESSOR_SILENT_EXIT_RESTART`, `PROCESSING_CANCELLED_ERROR` deleted; their strings inlined at the three use sites.
- Notification type strings (`first_start_setup`, `restart_greeting`, `proactive_check`, `nightly_dream`, `migration`) promoted to constants in `models.py`; filename stems match for cheap on-disk identification.
- New `body: str | None` field on `Notification` for multi-line system content; renders as inner XML text. External notifications keep their `key=value` rendering.

## Personality refactor

- Personality registers the voice-loading line into the restart skill via `agent/core/skills/personality/SETUP.md`, instead of the restart skill hardcoding it.
- `AGENT_SEED_PERSONALITY` env var (already injected by vestad) is the single pointer to the active preset; the `## Personality` block in restart is gone.
- "Pick one / swap" workflow dropped — drift happens in place on the active preset file.

## Migration: `2026-05-legacy-catchup`

One-shot catch-up for legacy agents:

1. Install `restart` and `timezone` skills via the skills registry.
2. Strip legacy `## Personality` block, run `personality/SETUP.md`.
3. Fold `~/agent/prompts/restart.md` body into the new skill.
4. Remove the legacy file + empty directory.
5. Run `upstream-sync` end-to-end (sparse-checkout narrow, refresh skills, restore `.gitignore`).
6. Manual-merge `MEMORY.md` (upstream layout + local profile/learned patterns).
7. Reconcile installed skills against `default-skills.txt` and the registry.
8. Call `mark_migration_applied("2026-05-legacy-catchup")`.

Fresh agents skip migrations entirely (pre-marked applied at first start).

## Code-review fixes folded into this PR

Following a multi-angle review of the diff, the following correctness bugs were addressed in-branch:

- **stripe-pay channel**: WhatsApp + Telegram JSON-key mismatches (`from_me` → `is_from_me`, `text` → `content`). Without this, the agent's own approval prompt was read back as the user's reply, or Telegram replies were dropped entirely.
- **agentmail webhook**: re-read secret per request (rotation took no effect on a running daemon); drop `body_text`/`body_html`, set `interrupt=false` so inbound mail doesn't preempt the agent and dump full bodies into context.
- **agentmail config**: `shlex.quote` values in `bashrc_set` so API keys containing `$`/`#`/whitespace survive the next shell start.
- **email-client poll daemon**: tmp+rename notification writes; cap `mb.fetch` batch to 500 messages (long offline → unbounded fetch could time out the IMAP transaction).
- **email-client account_profile**: keep per-account config overrides on unknown-provider `KeyError` fallback (was silently dropping `imap_host`/`smtp_host`/etc).
- **vestad serve.rs**: `realpath`-based symlink guard on `read_file_handler` and `write_file_handler` — an agent-controlled symlink under `/root/agent` could otherwise expose `/etc/shadow` via the FilesTab.
- **FilesTab**: cancellation guard on the file-load `useEffect` so a late `readFile` for an old selection can't clobber the user's in-flight edits.
- **CLI**: `server not reachable` error now includes the host being attempted plus the `vesta connect` recovery hint, instead of reading like "you have no setup".

## Other

- `set_timezone` MCP tool removed; `first_start_setup` writes `export TZ=` to `~/.bashrc` directly via the new `timezone` skill, and ends with the `restart` skill so the new TZ + registered services take effect together.
- `dream` skill: new "Consolidate" guidance for single-source-of-truth and splitting `MEMORY.md` into pointer files.
- `classic` personality preset: 😂 reserved for genuinely funny moments, not the default reaction.

Fixes #515

## Test plan

- [x] `uv run ruff check` clean
- [x] `uv run ty check` clean
- [x] `uv run pytest tests/ --ignore=tests/test_e2e.py` — 162 passed, 1 skipped
- [x] `cargo clippy -p vestad` clean, 85 vestad unit tests pass
- [x] Web build + 35 vitest pass
- [x] New `tests/test_state_store.py` covers defaults, round-trip, legacy import + cleanup, atomic write, "state.json wins over legacy markers"
- [x] `tests/test_migrations.py` rewritten for the file-drop API
- [x] `tests/test_dreamer.py` rewritten for sync `process_nightly_memory` + notification output + retry-until-done semantics
- [x] `tests/test_processor.py`, `tests/test_crash_recovery.py`, `tests/test_interrupts.py`, `tests/test_watchdog.py`, `tests/test_formatting.py` updated for the new state shape, module split, and inlined message handler
- [x] Verified on a fresh agent (`charbel`): no migration queued, pre-marked applied at first start, first-start setup notification drops, agent ran upstream-sync + personality SETUP + tasks/app-chat/dashboard install, called `mark_setup_done` to bring WS online, sent greeting via app-chat. Total ~3 min, $1.56, 6% context.
- [ ] Verify on an existing agent: first boot under new code imports `~/agent/data/{first_start_done, restart_reason, last_dreamer_run, show_dreamer_summary, session_id, migrations.applied}` into `state.json` and removes the originals; the legacy-catchup migration drops as a notification and the agent calls `mark_migration_applied` when done


elyxlz:
🤖 babysit-prs

CI was failing on:
- `test-integration`: a layout test asserted on `/root/agent/prompts` after the restart-prompt refactor moved that file
- `merge-gate-ci`: cascaded from `test-integration`

Pushed [`d569046`](https://github.com/elyxlz/vesta/pull/537/commits/d569046d8eec0d96b5c70a73fb81d6b675d9ea05):
- dropped the obsolete `/root/agent/prompts` assertion from the layout test

CI is now green ✅

PR #609, opened 2026-05-30
feat(web): add in-app gateway logs viewer

### LINKED ISSUE #476: feat(app): view gateway (vestad) logs from inside the app
## Problem

When the gateway misbehaves (won't start, dropping connections, agent fails to come up, update fails), the only way to see what happened is to SSH into the host and tail vestad's logs. Non-host users (mobile, remote desktop) have no path at all.

## Proposal

Add a "Gateway logs" view in the app, reachable from the agent/gateway menu or from the proposed disconnection overlay (#X above).

Behavior:

1. `vestad` exposes a tailing endpoint, e.g. `GET /logs?since=<cursor>` (JSON lines) and/or `GET /logs/stream` over WebSocket for live tail.
2. Authenticated with the existing API key (Bearer / `?token=`), same as other vestad routes.
3. App renders a simple log viewer: scrollback, follow-tail toggle, filter by level, copy-to-clipboard for support, optional download as file.
4. Cap the in-memory buffer so a long-running session does not eat unbounded RAM (e.g. last 5k lines, with "load older" pagination).

## Why it matters

- Diagnoses real incidents from any device, not only the host.
- Materially improves bug reports: users can copy the log slice straight into a GitHub issue.
- Pairs naturally with the "block UI when gateway disconnected" issue: when reconnect keeps failing, the overlay can offer a "view gateway logs" link.

## Notes

- Be careful about secrets in logs. vestad already redacts agent tokens / API keys from log output, but worth re-auditing before exposing the stream broadly.
- Consider per-agent container log access too (separate followup), but the gateway is the higher-priority target since it owns the failure modes that hide the most context today.

elyxlz:
## Summary

There was no way to read vestad (gateway) logs from inside the app, so mobile and remote users had to SSH into the host. vestad already exposes an authed `GET /gateway/logs` SSE route (backed by `journalctl --user -u vestad`, behind the existing API-key/JWT auth), but nothing in the app called it. This PR adds the missing in-app viewer.

- New `api/gateway.ts` streams `/gateway/logs` over SSE using the existing token auth (`?token=`), mirroring the existing `api/logs.ts` agent-log streamer. It satisfies the pre-existing `streamGatewayLogs`/`stopGatewayLogs` export contract style used across the api module.
- New `GatewayLogsViewer` component (its own folder with `index.tsx` + `styles.ts`) renders scrollback with a follow-tail toggle, refresh, and copy-to-clipboard, reusing the shadcn `Dialog`, `Button`, and `Switch` primitives. Auto-scroll only sticks when the user is near the bottom, matching the existing `Console` log viewer behavior.
- Wired into the existing Gateway section of the settings dialog via a "View logs" button.

Scope was intentionally kept small: no WebSocket changes, no new backend route (one already exists), no version bumps.

## Redaction

No new secrets enter the logs. The route reads systemd's journal rather than an in-process buffer, and vestad never logs raw API keys or agent tokens (`auth.rs` logs only short token fingerprints). Redaction posture is unchanged.

## Files

- web (TS): `apps/web/src/api/gateway.ts` (new), `apps/web/src/components/GatewayLogsViewer/{index.tsx,styles.ts}` (new), `apps/web/src/components/Settings/index.tsx` (View logs button).
- vestad (Rust): no changes (the `GET /gateway/logs` route already exists).

## Verification

- `npm -w @vesta/web run check`: pass (0 errors)
- `npm -w @vesta/web run lint`: pass (eslint exit 0; warnings are all pre-existing, and the one in the new viewer matches the existing Console pattern)
- `cargo clippy -p vestad`: pass (0 warnings)
- `cargo test -p vestad`: pass (162 tests, 0 failed)

Fixes #476

### DIFF
--- a/agent/MEMORY.md
+++ b/agent/MEMORY.md
@@ -97,6 +97,10 @@ The user's important people are [agent_name]'s important people too. Keeps track
 - To add a new server: register with vestad to get a port, start it in a screen session, and add the command to the `## Services` section of `~/agent/skills/restart/SKILL.md`.
 - **Public services**: pass `"public": true` in the registration body to make a service accessible without authentication (e.g. hosting a website). Public services are fully open, no auth token needed. Default is `false` (requires auth).
 
+### Self-diagnosis
+- Read the gateway (vestad) logs to debug gateway or container issues: `curl -sk "https://localhost:$VESTAD_PORT/gateway/logs?tail=200" -H "X-Agent-Token: $AGENT_TOKEN"`
+- This returns the last N gateway log lines as Server-Sent Events, so parse the `data:` lines. It closes after the tail. Add `&follow=true` to keep streaming live.
+
 ### Self-Modification
 - Edit skills, prompts, MEMORY.md freely
 - **To change a config setting**: read `core/config.py` for all options and their env var names; set the env var in `~/.bashrc`, then call the `restart_vesta` MCP tool
--- a/apps/web/src/api/gateway.ts
+++ b/apps/web/src/api/gateway.ts
@@ -1,0 +1,55 @@
+import { getConnection } from "@/lib/connection";
+import type { LogEvent } from "@/lib/types";
+
+let gatewayLogSource: EventSource | null = null;
+
+export function streamGatewayLogs(
+  follow: boolean,
+  onEvent: (event: LogEvent) => void,
+): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const conn = getConnection();
+    if (!conn) {
+      reject(new Error("not connected"));
+      return;
+    }
+
+    const params = new URLSearchParams({ token: conn.accessToken });
+    if (follow) params.set("follow", "true");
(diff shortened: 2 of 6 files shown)


elyxlz:
Update: `GET /gateway/logs` now accepts either the vestad API key or an agent token. Pulled the route into its own small router layered with `auth::auth_middleware_api_or_agent_token` (mirroring `agents_services_read`), leaving the rest of the `vestad_protected` group API-key-only. SSE handler behavior is unchanged.

Also documented agent self-diagnosis in `agent/MEMORY.md` (SYSTEM CONFIGURATION): the agent can read the last N gateway log lines via

```
curl -sk "https://localhost:$VESTAD_PORT/gateway/logs?tail=200" -H "X-Agent-Token: $AGENT_TOKEN"
```

which returns Server-Sent Events (parse the `data:` lines) and closes after the tail; add `&follow=true` to stream live.

Verified locally: `cargo clippy -p vestad` clean, `cargo test -p vestad` passing.

PR #615, opened 2026-05-31
feat(app): notify when an agent loses Claude credentials

### LINKED ISSUE #481: feat(app): notify when an agent loses auth (Claude credentials expired/revoked)
## Problem

When an agent loses Claude auth (token expired, OAuth refresh failed, credentials wiped, plan downgraded), the failure surfaces silently from the user's point of view: the agent stops processing messages, the chat UI just sits there, no notification, no banner. The user only finds out when they notice the agent has not replied for a while or by digging into vestad logs.

## Proposed behavior

When the app detects an auth-loss state for any agent it has registered, push a notification:

1. **In-app**: prominent banner / toast on the agent's chat (and on the home screen if multiple agents) saying "Vesta agent <name> has lost its Claude credentials. Tap to re-authenticate."
2. **OS-level**: native push (mobile + desktop) so the user sees it even when the app is closed. Reuse the existing notification permission flow.
3. **Tap-through**: the notification opens the agent's onboarding / re-auth step directly, not the home screen.

The notification should debounce — one push per auth-loss event, not one per failed message.

## Detection

Two signals to consider:

- vestad / agent reports a structured event when the SDK returns an auth error (401 from Anthropic, missing/expired credentials at boot). Today this likely shows up only as a log line.
- The control-WS could carry an `agent_state: \"auth_lost\"` event so the app does not have to poll.

A small server-side change to surface the auth-error class as a typed event is probably required. Path-of-least-resistance is for the agent's `core/loops.py` to emit an `auth_error` event on the EventBus when it sees an Anthropic 401, and for vestad to forward that on the control-WS.

## Why it matters

Auth loss is the single most disruptive failure mode for a long-running agent: the user thinks Vesta is broken when actually it just needs to re-login. A clear push notification turns a confusing dead-agent experience into a one-tap fix.

## Related

- StatusPill / GatewayProvider already drive other status badges in the app, so the rendering surface exists.
- Pairs naturally with #475 (block UI when gateway disconnected) — both are \"Vesta cannot serve you, here is why\" UX.

elyxlz:
Fixes #481

## Problem

When an agent loses its Claude auth, it silently stops processing and the user gets no proactive heads-up.

## Approach

Reuse the machinery that already exists. vestad already computes a `not_authenticated` agent status from the credentials-file expiry check (`is_authenticated` in `vestad/src/docker.rs`), and the app already renders it as the orb's "not signed in" state plus the navbar **reauthenticate** button. The only thing missing was a proactive notification when that status flips.

So this PR is a single, additive web change: `NotificationProvider` watches each agent's status and, on a transition into `not_authenticated`, fires one OS notification (deduped by `tag`) that routes to the agent's chat on tap via the existing `focusAndNavigate`. No new event type, no agent-side detection, no vestad change.

## What this intentionally does NOT do

An earlier draft added agent-side runtime detection (string-matching SDK 401s and deleting the credentials file). That was dropped to keep the change minimal. Consequence: a token **revoked server-side before its `expiresAt`** won't surface as `not_authenticated` until it actually expires, so that narrow case is still covered only on expiry. Normal expiry and wiped/missing credentials are caught immediately by the existing check.

## Verification

- `apps/`: `npm -w @vesta/web run check` pass, `npm -w @vesta/web run lint` -> 0 errors (pre-existing warnings only, none in the changed file).

### DIFF
--- a/apps/web/src/providers/NotificationProvider/index.tsx
+++ b/apps/web/src/providers/NotificationProvider/index.tsx
@@ -14,7 +14,7 @@ import { setAppBadge } from "@/lib/app-badge";
 import { setFaviconUnseen } from "@/lib/favicon";
 import { useWindowFocus } from "@/hooks/use-window-focus";
 import { router } from "@/router";
-import type { VestaEvent } from "@/lib/types";
+import type { AgentInfo, VestaEvent } from "@/lib/types";
 
 const RECONNECT_BASE_MS = 1000;
 const RECONNECT_MAX_MS = 30000;
@@ -89,6 +89,7 @@ export function NotificationProvider({ children }: { children: ReactNode }) {
   const permissionRef = useRef<boolean>(false);
   const tapsRef = useRef<Map<string, TapEntry>>(new Map());
   const chattingAgentRef = useRef<string | null>(null);
+  const prevStatusRef = useRef<Map<string, AgentInfo["status"]>>(new Map());
 
   useEffect(() => {
     const onVisible = () => {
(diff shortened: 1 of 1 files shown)


elyxlz:
Reworked to reuse the existing not_authenticated agent status instead of a parallel auth-lost system.

New approach:
- On a detected Claude auth error the agent now deletes its credentials file (/root/.claude/.credentials.json) instead of emitting a custom event. vestad's existing 3s status poll then sees no creds and reports the existing not_authenticated status, so the existing in-app reauthenticate UI (AgentNavbar button, Orb, ActionsCard via AuthFlow) handles recovery. No new vestad code, no new agent to vestad endpoint.
- The OS push now fires off the gateway agent status: NotificationProvider tracks each agent's previous status and fires one debounced OS notification when an agent transitions into not_authenticated, reusing the existing notification permission flow and the existing focusAndNavigate click routing.

Removed from the previous version:
- the custom auth_error EventBus event (AuthErrorEvent in agent/core/events.py and the StreamEvent union)
- the AuthLostBanner component
- ChatProvider.authLost tracking and the Chat banner render
- the auth_error entry in types.ts

Kept: _looks_like_auth_error() plus its tests, and the auth_error_notified debounce flag (now gates the file delete plus log, re-armed after a successful turn).

Verification: agent ruff, ty, and pytest tests/test_notifications.py (4 passed) plus web lint and check all green. git diff origin/master for apps/web/src/components, apps/web/src/lib, and apps/web/src/providers/ChatProvider is empty; the only remaining web change is NotificationProvider.

PR #617, opened 2026-05-31
fix(migrations): re-anchor-aware workspace resync that also repairs already-migrated agents

elyxlz:
## Problem

The shipped `2026-05-legacy-catchup` migration tells legacy agents to run upstream-sync end to end, but gives no guidance for the case where the agent's local branch and upstream **share no common ancestor** (which happens because the repo was deleted and recreated, so previously-merged upstream commits have new SHAs). A normal `git merge` then forces `--allow-unrelated-histories` and floods with false conflicts.

In production this caused one agent to grind through **171 conflicts** by hand over ~16 minutes, commit vestad-managed `agent/core/` onto its branch, and leave skill spillover on disk.

## Why a new filename instead of editing in place

Migrations are keyed by **filename stem**, tracked in `state.json` `applied_migrations` (persists across updates). A same-name edit can never re-run for agents that already applied the broken `2026-05-legacy-catchup` (the name is already in their applied set). To repair that group, the corrected migration ships under a **new name**, and the old file is removed so nobody runs the broken body twice.

Coverage:
- `<154` → `155` (majority, skipped the broken release): new name not applied → fires with the fixed body.
- `154` → `155` (already ran the broken one): new name not applied → fires and repairs.
- Fresh `155` installs: `first_start` → pre-marked, never runs.

## What the new body changes

- Adds the **re-anchor escape hatch** for the no-common-ancestor case: take the upstream tree as the new base and re-apply only owned files (`MEMORY.md`, installed skills, `.gitignore`). Still pulls all upstream content, skips the false conflicts.
- States that `agent/core`, `pyproject.toml`, `uv.lock` are vestad-managed and must **never be committed to the branch**, and to repair it if a prior run did.
- Trims skills to: container defaults + `skills-install`'d + **self-authored** (kept because they aren't in `index.json`); drops only registry spillover.
- `MEMORY.md` reconcile stated as a principle, not a step-by-step recipe.
- Idempotent and repair-oriented, so re-running on a half-converged agent is safe.

## Notes

- No code changes; `agent/tests/test_migrations.py` uses synthetic fixtures, not the real files. Migration tests pass locally (10 passed).
- No version bump (handled by `release.sh` at release time).


elyxlz:
Folded into #618, which reworks upstream-sync and includes this migration (updated to drive the new sync.sh). Closing in favor of the unified PR.

PR #623, opened 2026-06-01
feat(agent): drive Claude Code in tmux instead of the Claude Agent SDK; harden test suite + CI

elyxlz:
## What

Replaces the official **`claude_agent_sdk`** (which drives `claude` headlessly over the control protocol) with **`agent/core/cc_sdk/`** — an in-repo package that exposes the *same public surface* but instead runs the **interactive `claude` CLI programmatically inside a private tmux server**. `core/` just imports `cc_sdk` where it used to import `claude_agent_sdk`; everything else is unchanged. cc_sdk lives under `core/` so it ships and versions with the embedded agent code.

On top of that, this PR hardens the entire test and CI setup: a single `check.sh` entry point shared by CI and local dev, hermetic agent images, sleep-free synchronization, per-test timeouts, property-based tests, cross-language contract tests, one unified CI workflow, and a no-retry nightly run.

## How the transport works

```
ClaudeSDKClient ──(tmux paste + Enter)──▶ claude TUI (own tmux -L server)
       ▲                                      │ native hooks   │ mcp stdio
       │ tail transcript.jsonl                ▼                ▼
   AssistantMessage / ResultMessage     _forward.py       _mcp_stdio.py
                                              │                │
                                              ▼                ▼
                                     unix-socket bridge (in agent process)
                                       → HookMatcher callbacks / tool handlers
```

- **Prompts** are submitted by bracketed-pasting text into the pane + `Enter` (multi-line stays in the box; a single Enter submits; leading `/` runs a slash command). **Interrupt** sends `Escape`.
- **Responses** are reconstructed by tailing the session transcript JSONL (`~/.claude/projects/<munged-cwd>/<session-id>.jsonl`). Each main-agent `assistant` line becomes an `AssistantMessage` with the same `TextBlock`/`ThinkingBlock`/`ToolUseBlock` shape. A turn ends on the native **Stop** hook, after which a synthetic `ResultMessage` (session id + usage) is yielded. Subagent (`isSidechain`) lines are skipped — they surface via SubagentStart/Stop hooks.
- **Hooks** are native command hooks (`--settings`) that relay event JSON to an in-process **unix-socket bridge**, dispatched to the existing `make_hooks()` `HookMatcher`s — unchanged.
- **MCP tools** from `create_sdk_mcp_server` are exposed to `claude` as a real **stdio MCP server** (`--mcp-config`) proxied back to the bridge, so handlers still run in-process and mutate live `State`.
- **Pinned binary**: cc_sdk owns a pinned `claude` binary (`_claude_bin.py`) so the transport is reproducible regardless of what's globally installed.
- **Startup** pre-seeds `~/.claude.json` (onboarding + per-project trust) so a fresh container goes straight to the input box, then waits for the `SessionStart` hook (which also hands over the transcript path).

See `agent/core/cc_sdk/README.md` for the full design.

### The MCP first-start bug (root cause)

First-start originally couldn't call `mark_setup_done` and had to drive the bridge socket by hand. Root cause, captured from claude's real MCP handshake: tools declared with `inputSchema: {}` (`restart_vesta`, `mark_setup_done`, `mark_dreamer_complete`) are invalid JSON Schema, and Claude Code 2.1.x rejects the **entire** MCP server when any tool schema is malformed — so every vesta control tool became invisible to the model. Fix: `ToolDef` normalizes a missing/`{}` `inputSchema` to `{"type": "object"}` in one place, covering all tools. The symptom-chasing workarounds (static tools/list file, `ENABLE_TOOL_SEARCH=false`, bridge health checks) were reverted once the real cause was found.

## cc_sdk changes

- **New:** `agent/core/cc_sdk/` — types, messages, mcp, transcript tailing, tmux control, bridge, client, pinned-binary management, two stdlib-only helper scripts (`_forward.py`, `_mcp_stdio.py`), README.
- `core/` + tests: `claude_agent_sdk` → `cc_sdk` imports (same symbols); removed two now-unused `ty: ignore`s that proper typing made redundant.
- `agent/pyproject.toml`: drop `claude-agent-sdk`, add `cc_sdk` to ty; relocked `uv.lock`.
- `vestad/Dockerfile`: install `tmux`, recreate `/root/.claude` (the pinned binary no longer runs `claude install`).
- `vestad/build.rs` + `agent_code.rs`: hash the rust-embedded `agent/core` inputs into `VESTAD_EMBED_HASH` so any agent-code change forces a vestad recompile → fresh embed snapshot (previously `rerun-if-changed` alone could ship a stale embed and crash-loop the agent with ModuleNotFound).
- `CLAUDE.md`: document the new transport.

### `check.sh` — single entry point (CI/local parity)
CI jobs run `./check.sh <suite>` (`agent`, `cli`, `vestad`, `vestad-docker`, `web`, `integration`, `live`, `all`). Local and CI checks are identical by construction, so they can't drift (formatting checks, `-D warnings`, etc.).

### Hermetic agent images in CI
`VESTAD_AGENT_IMAGE` override in vestad (`resolve_image()` + `#[ignore]` Docker tests). `test-integration`, `check-vesta`, and `test-live` build the agent image **from the checkout** with GHA layer caching and test against it. Previously: integration pulled the *previously released* `ghcr.io :latest`, the Docker unit tests ran against the old release, and release-time live tests validated the **previous** release's image instead of the one being shipped.

### Integration/live suites moved into `vestad/tests/`
The server/multi_user/live/oauth/migrations suites moved out of the standalone `vesta-tests` crate into vestad's own `tests/`, so `cargo test -p vestad --test <name>` builds the vestad binary and hands tests its path via `CARGO_BIN_EXE_vestad` — no `find_vestad()` heuristic, no separate build, no stale-binary risk. `vesta-tests` is now a harness-only lib that vestad dev-depends on; unit tests are scoped with `--bins` so the heavy Docker/live suites don't run in the unit job. New live suites: `interrupt.rs`, `mcp_tools.rs`.

### Sleep-based synchronization eliminated
- `agent/tests/wait_util.py`: `wait_for_condition()` deadline polling
- `test_interrupts.py` rewritten: fake SDK streams coordinate via consumed-message handshakes instead of `asyncio.sleep` guesses
- `test_api`/`test_processor`/`test_watchdog`/`test_crash_recovery`: every "sleep then assert" replaced with explicit conditions
- `tests/server/websocket.rs`: blind 2s sleep replaced with a deadline poll
- `pytest-timeout` (120s/test): hung async tests fail fast instead of stalling CI

### Property-based tests (found + fixed 2 real bugs)
- **hypothesis** (agent): EventBus round-trip with adversarial unicode, FTS5 search robustness
- **proptest** (vestad + cli): `normalize_name`/`validate_name`, `normalize_url`/`version_less_than` invariants
- **fast-check** (web): `stripAnsi` + `linkify` properties, which caught:
  - `stripAnsi` was not idempotent — nested escapes left live ANSI codes after one pass
  - `linkify` crashed (TypeError) on text containing literal `\x00URL<n>\x00`

### Cross-language API contract tests
- Fixed `test_contract.py`'s stale `app/` path — the Python↔TS contract check had been **silently skipping** since the web app moved to `apps/web/`
- New: vestad generates `apps/web/src/lib/vestad-api-fixtures.ts` from production serialization code (control WS agents message, backups, auth, personalities, tree, version); `api-contract.test.ts` `satisfies`-checks them against the web's TS types at compile time

### One CI workflow + nightly
- `pr-checks.yml` merged into `ci.yml`: one `detect-changes`, one merge gate (`merge-gate-ci`, also fails on cancelled jobs)
- cli tests now actually run on macOS and Windows (previously clippy-only)
- `build-vesta` is a pure build job (clippy/tests deduplicated into `check-vesta`)
- Integration retries are now **visible**: warning annotation + job summary when >1 attempt was needed
- Every job has `timeout-minutes`
- New `nightly.yml`: full suite + Docker tests with **no retry** (flakes surface there); failures open/update a `nightly-failure` issue

### Hygiene
- CLAUDE.md: fixed stale root `tests/` crate reference, documented check.sh, updated CI section
- tests-integration README: real counts and file lists
- Removed dead `--ignore=tests/test_e2e.py` flag, dead `e2e.py` pytest pattern, empty `finance/cli/tests/` dir

## Verification

- `uv run pytest` (green), `ruff`, `ty` — all pass
- Rust: vestad unit 85, integration 62, live 5/5 locally
- Real end-to-end against `claude` v2.1.x: streaming text, `PreToolUse` hooks firing for both builtin (Bash) and in-process MCP tools, session resume across turns, `get_context_usage`, and **interrupt → drain → recovery** through the real `core.client.converse` path

## Notes

- Helper scripts run with `PYTHONSAFEPATH=1` so `cc_sdk/types.py` can't shadow the stdlib `types` module
- Requires `tmux` + `claude` on `PATH` (both in the agent image)
- No version bumps (handled by `release.sh`)

> **Branch protection note**: merging requires dropping the `merge-gate-pr` required check (that job no longer exists; `merge-gate-ci` is the single gate).


elyxlz:
## Added: e2e regression coverage + two interrupt bug fixes

**New e2e test layers** (commit f2aa3805):

| Layer | What it covers | Where it runs |
|---|---|---|
| `agent/tests/test_e2e_transport.py` (13 tests, ~9s) | Full cc_sdk machinery against a fake claude TUI in **real tmux**: bracketed paste/Enter, `_forward.py` hooks → unix-socket bridge, `_mcp_stdio.py` MCP round trips, transcript tailing, interrupts, crash detection, startup-failure cleanup, resume, CLI arg contract | `agent-tests` (every PR, no credentials) |
| `vestad/tests-integration/tests/live/mcp_tools.rs` | Real claude must invoke an in-process MCP tool mid-conversation (stdio proxy → bridge → events.db handler), verified via the session transcript | new `test-live` CI job + locally |
| `test-live` CI job | The whole live suite (file ops, migration, MCP) with a real agent + real Claude via the existing `CLAUDE_CREDENTIALS` secret; in `merge-gate-ci` | agent-touching PRs + master + releases |

The fake TUI reproduces behaviors **verified empirically against real claude v2.1.159 in tmux** (hook payload shapes, transcript line shapes, escape parsing, bracketed-paste handshake), so the transport tests fail if cc_sdk drifts from how the real TUI actually behaves.

**Two real bugs found by that verification, fixed in `cc_sdk/client.py`:**

1. **Interrupt never worked.** `interrupt()` sent a single `Escape`, but the TUI's escape parser buffers a lone ESC waiting for a follow-up byte — generation ran to completion, the Escape was never seen. Fix: double Escape (verified working), plus an idle guard (a double Escape at idle opens the rewind dialog).
2. **Every turn after an interrupt would hang.** An interrupted turn never fires its `Stop` hook (verified — no late Stop arrives either), so the count-based turn matching could never reach the next turn's threshold. Fix: `interrupt()` credits the abandoned turn's missing Stop.

Both are covered by new unit tests and the e2e interrupt test, which fails if either fix regresses.

Local verification: ruff/ty/pytest all green (204 passed), live MCP test passed against real claude in 615s, no leaked containers.

PR #632, opened 2026-06-03
refactor(agent): tighten exception handling and prune dead abstractions

elyxlz:
No behavior change. A focused quality pass over the agent core that tightens exception handling and removes a few dead abstractions.

## Changes

- **sdk_parsing.py**: Replace the banned dict `.get()` fallback style in usage-token parsing with the codebase idiom `key if key in d else default`. Values only feed a log string.
- **api.py**: Stop swallowing the real error in the `/search` handler. It now catches only `sqlite3.OperationalError` (the FTS5 parse failure case), logs it, and keeps the user-facing 400. Any other exception type propagates to aiohttp so genuine bugs surface as 500 instead of a misleading "invalid search query".
- **loops.py (monitor_loop)**: Let cancellation propagate through the existing `finally` (which still cancels and awaits the watcher task) instead of catching `CancelledError` and breaking silently. This restores correct task-cancelled semantics for the caller in main.py.
- **loops.py (run_one)**: Drop the dead `session_id` try/except in the error path. `session_id` is a plain property and cannot raise `AttributeError` or `TypeError` once `state.client` is truthy.
- **events.py**: Remove the unused `EventBus.tool_started` and `tool_finished` methods and the `_active_tools` counter. `State.active_tools` is the single source of truth for active-tool tracking.
- **cc_sdk/client.py + diagnostics.py**: Drop the leaky `_Proc`/`_Transport` mirror dataclasses and expose a public `returncode` property on the client. `diagnostics._check_sdk_subprocess_alive` now reads `client.returncode` directly instead of reaching across two private boundaries with type-ignore comments. The exit code is no longer mirrored into two places.

## Notes

Two test files asserted against the removed internal `_transport._process` structure. They were adapted to the new public `returncode` surface with the same behavioral assertions, no assertions weakened or removed.

`./check.sh agent` is green (ruff, format, ty, 220 tests pass).


elyxlz:
🤖 babysit-prs

Was failing on:
- `merge conflict in agent/core/api.py`: import-block clash. master's #602 moved `CREDENTIALS_PATH` into `core/provider.py`, so the PR's local `import pathlib as pl` (only used to build that path) became dead, while its new `import sqlite3` (for `except sqlite3.OperationalError` in the search handler) was still needed.

Pushed [`54b72d36`](https://github.com/elyxlz/vesta/commit/54b72d36):
- `agent/core/api.py:14-22`: kept `import sqlite3`, dropped `import pathlib as pl` (now unused since `CREDENTIALS_PATH` is imported from `.provider`).

`./check.sh agent` is green locally (ruff + format + ty + 265 pytest). CI is now green and the branch is mergeable ✅

PR #685, opened 2026-06-04
feat(agent): add events.db schema-version migration seam

### LINKED ISSUE #641: Establish an events.db schema-version and migration seam before the first schema change
## Problem
The SQLite schema is applied with `CREATE ... IF NOT EXISTS` on every boot and there is no `PRAGMA user_version` or migration runner for events.db. The prompt-based migration system converges agent prose/state, not the events.db SQL schema. If a future image needs to add a column, change the FTS tokenizer, or alter a trigger, existing databases silently keep the old schema (IF NOT EXISTS no-ops) and new code reads/writes assuming the new shape against an old-shaped table. The schema is stable today, so nothing is broken, but the first schema change becomes a silent data-corruption incident across all existing agents.

## Evidence
- agent/core/events.py:106-130 (`_EVENTS_SCHEMA` uses only `CREATE ... IF NOT EXISTS`)
- agent/core/events.py:144-145 (`executescript` on every EventBus construction, no version check)
- agent/core/migrations.py (migration system is prompt/notification based, converges agent state, not SQL schema)

## Proposed direction
Introduce `PRAGMA user_version` with a small ordered list of in-process migration steps applied at EventBus construction (each idempotent and version-gated), mirroring how state_store handles legacy import. Document the strategy in events.py and CLAUDE.md as the decision record. Do not add speculative migrations now; just establish the versioned seam plus a test that an old-schema db is upgraded in place.

## Alternatives considered
Keep IF NOT EXISTS and handle changes ad hoc when needed: rejected, that is the exact silent-divergence trap. A heavyweight migration framework: overkill for one in-process SQLite file.

## Effort and risk
Medium. The seam itself is small; the value is establishing it before it is needed. Risk is low since no behavior changes until a real migration is added.

elyxlz:
## What

Establishes a `PRAGMA user_version` based migration seam for `events.db` before the first real schema change, so future column/trigger/FTS changes migrate explicitly instead of silently diverging (the `IF NOT EXISTS` no-op trap described in the issue).

## How

- `events.py` now owns an ordered, version-gated `_MIGRATIONS` list applied by `_migrate()` at `EventBus` construction. Each step is idempotent.
- Version 1 is the baseline: it runs the existing `_EVENTS_SCHEMA` (all `CREATE ... IF NOT EXISTS`). A fresh db and a pre-versioned db (tables present, `user_version=0`) both converge to v1 with no data loss; the latter is simply stamped.
- No speculative future migrations added. Mirrors the legacy-import spirit of `state_store.py`.
- Decision record added to `CLAUDE.md`'s session-persistence section.

## Tests

`agent/tests/test_eventbus.py`:
- `test_fresh_db_stamped_to_current_version` — fresh db ends at `_SCHEMA_VERSION`.
- `test_pre_versioned_db_upgraded_in_place` — old-schema db (tables present, `user_version=0`) is upgraded in place: stamped to v1, prior row preserved, no error.

`./check.sh agent` passes (276 passed).

Fixes #641


elyxlz:
🤖 babysit-prs

Was failing on:
- merge conflict in `agent/tests/test_eventbus.py`: both sides edited the `from core.events import ...` line (this PR added `_EVENTS_SCHEMA`/`_SCHEMA_VERSION`, master added `SUBSCRIBER_QUEUE_MAXSIZE`)

Pushed [`8196ae88`](https://github.com/elyxlz/vesta/commit/8196ae88):
- `agent/tests/test_eventbus.py:6`: combined both imports into one line so the migration-seam symbols and master's queue-maxsize symbol are all imported (all three are used in the merged tests)

The rest of the merge (events.py and other core/web files) auto-resolved. Full agent suite passes locally (278 passed).

PR #720, opened 2026-06-05
fix(agent): coerce legacy THINKING json-dict env into the typed config

elyxlz:
## Problem

Part 2 of #713. The `THINKING` env var is documented as a plain string (`adaptive|enabled|disabled`), but older Vesta versions accepted a JSON object form, e.g. `THINKING='{"type":"adaptive"}'`, back when `ThinkingConfigAdaptive` did not require a `display` field.

`_parse_thinking` only coerced the string form and passed dicts straight through to pydantic union validation. Now that `display` is required, a legacy dict fails against all three branches and the agent dies before it even logs its config:

```
ValidationError: 4 validation errors for VestaConfig
thinking.ThinkingConfigAdaptive.display  Field required [input_value={'type': 'adaptive'}]
thinking.ThinkingConfigEnabled.type      Input should be 'enabled' [input_value='adaptive']
...
```

Only deployments whose env file predates the string form are affected; new installs using `THINKING=adaptive` are fine.

## Fix

Add a dict branch to `_parse_thinking` that fills in the same defaults the string form already uses (`adaptive` → `display="summarized"`, `enabled` → default budget). Unknown `type` values still fall through to pydantic's normal error, so genuinely malformed configs are not silently accepted.

One concern, one validator, no behavior change downstream: `_thinking_env` in cc_sdk already consumes the resulting typed dict regardless of which env form produced it.

## Tests

`test_config.py`:
- `test_thinking_legacy_json_dict_coerces_with_defaults` — the three legacy dict forms coerce to complete typed configs.
- `test_thinking_string_form_still_parses` — the documented string form is unchanged.

### DIFF
--- a/agent/core/config.py
+++ b/agent/core/config.py
@@ -59,7 +59,7 @@ class VestaConfig(pyd_settings.BaseSettings):
     @classmethod
     def _parse_thinking(cls, value: object) -> object:
         # The THINKING env var is documented as a plain string (adaptive|enabled|disabled);
-        # coerce it into the SDK's config dict. JSON object values pass through untouched.
+        # coerce it into the SDK's config dict.
         if isinstance(value, str):
             mode = value.strip().lower()
             if mode in ("", "adaptive"):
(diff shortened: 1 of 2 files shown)


elyxlz:
Combined into #721 (both config-layer fixes ship together there). The THINKING coercion is included in #721 unchanged.

PR #749, opened 2026-06-08
fix(app): virtualize chat and console message lists (#474)

### LINKED ISSUE #474: fix(app): chat UI slows once tool calls are rendered
## Symptom

Once a conversation has accumulated rendered tool calls (the `ToolCallLabel` rows in `apps/web/src/components/Chat/ToolCallLabel/`), the whole chat UI becomes noticeably sluggish: scroll lags, typing in the composer feels heavy, new message render is delayed. Effect compounds with the number of tool calls in the visible history.

## Repro

1. Open a busy conversation that has dozens of tool calls (e.g. an active workday session).
2. Scroll up and down, type in the composer, send a new message.
3. Notice frame drops and input lag that scale with the number of tool calls on screen.

## Hypothesis

`ChatMessageArea` likely re-renders the full message list on every event, with each `ToolCallLabel` re-rendering as well. Possible culprits to check:

- No memoization on `ToolCallLabel` or per-message components, so every event-bus tick re-renders the entire list.
- Tool-call payload pretty-printing or syntax highlighting running on every render.
- Markdown rendering (react-markdown) per message on every render rather than memoized output.
- No virtualization on the message list.

## Suggested investigation

1. Profile with React DevTools: confirm where the wasted render time goes.
2. Memoize `ChatBubble` and `ToolCallLabel` on stable message id/version.
3. If markdown/highlight is the cost, memoize the rendered node per message.
4. If list size dominates, add windowing (react-virtual / react-window).

## Impact

Makes the app feel broken in exactly the kinds of long sessions where Vesta is most useful.

emipasca:
## Summary

Fixes #474. Long conversations (dozens of tool calls) and busy log streams made the whole UI sluggish — scroll lag, heavy typing, delayed renders — because the message/log lists were **not virtualized**: every WebSocket event re-rendered the entire list (up to 5000 items), re-parsing markdown per bubble, re-rendering every `ToolCallLabel`, and re-running tag-coloring + linkify for every console line.

This windows both lists with **react-virtuoso** so only the visible rows render. Applied to the **Chat** message area and the **Console** log viewer.

## Chat (`ChatMessageArea`)
- Replaced the `flex flex-col-reverse` scroll container with `<Virtuoso>` (`alignToBottom` + `followOutput` + `initialTopMostItemIndex: LAST`).
- New pure, unit-tested helpers in `virtual.ts`:
  - `buildDecorated` — memoized day-stamp/gap decoration (moved out of the render path), with collision-safe unique row keys.
  - `computeFirstIndexShift` / `useStableFirstItemIndex` — keep scroll position stable across load-more **prepend** and the `MAX_MESSAGES` front-drop.
- `startReached` → `loadMore`; `atBottomStateChange` → new-message banner; `scrollToBottom` via the Virtuoso handle (replacing the IntersectionObserver + reverse-scroll math).
- Memoized `ChatBubble` (now takes `isMobile` as a prop) and `ToolCallLabel`.

## Console
- Replaced `lines.map()` with `<Virtuoso>`; each line gets a stable monotonic `id`, and `firstItemIndex` is derived from it so the cap's front-drop doesn't desync keying.
- Precompute color class + linkified HTML **once on arrival** instead of per render.
- `followOutput` tails the stream; `atBottomThreshold={80}` preserves the old "near-bottom keeps following" band. Removed the now-unused `useAutoScroll` hook.

## Verification
- Web `lint` (0 errors), `tsc`, `prettier`, and `vitest` all pass — including the new `virtual.test.ts` (firstItemIndex shift cases, key uniqueness, day-stamp/gap decoration).
- Production `vite build` is clean.
- **Manual QA still recommended** (jsdom can't measure Virtuoso layout): verify in a running app that scroll/load-more doesn't jump, the new-message banner works, tool-call expansion stays pinned, and the console tails correctly. The console empty-state bottom-pinning in particular is worth an eyeball.


elyxlz:
🤖 babysit-prs

Was failing on:
- merge conflict in `apps/web/package.json` / `apps/package-lock.json`: master bumped `react-hook-form` (^7.78.0) and `react-router-dom` (^7.17.0) while this PR added `react-virtuoso` and pinned the older versions

Pushed [`06eaf5a9`](https://github.com/elyxlz/vesta/commit/06eaf5a9):
- merged `origin/master`, took union of deps in `package.json` (kept master's higher `react-hook-form`/`react-router-dom` versions plus this PR's `react-virtuoso ^4.18.7`), regenerated `package-lock.json` from scratch via `npm install`
- verified `./check.sh web` (eslint 0 errors, prettier, tsc, 69 vitest tests) all green after the merge

PR #811, opened 2026-06-11
fix(vestad,web): recover from expired sessions instead of looping forever

emipasca:
## Problem

A user who hadn't opened the app in a while got stuck on the loading skeleton behind the "disconnected from gateway / attempting to reconnect…" overlay, forever. Manually disconnecting and re-pairing fixed it.

Two compounding bugs:

1. **The refresh-token family expiry was absolute, not idle-based.** `RefreshFamily.exp` was set once at login (`now + 7d`) and never extended on rotation, so every session died exactly 7 days after pairing regardless of activity — even though each re-minted refresh JWT claimed `exp = now + 7d`.
2. **The client had no terminal-failure path.** `doConnect` ignored `ensureFreshToken()`'s result and reconnected with the dead access token; vestad rejected the WS handshake, `onclose` scheduled a retry, and the loop never ended. `doRefresh` also returned the same `false` for "vestad said 401" and "network blip", so callers couldn't tell a dead session from a transient outage.

## Fix

**vestad**
- `rotate()` slides `family.exp = now + REFRESH_TOKEN_TTL` on every successful rotation (including the retry-grace branch, whose re-minted token must be outlived by the family).
- `REFRESH_TOKEN_TTL` raised 7 → 30 days and documented as an idle window: active clients never expire; an idle one re-auths after 30 days.

**web**
- `ensureFreshToken` returns `"ok" | "transient" | "expired"`: 401 from `/auth/refresh` is terminal, network/5xx is transient, the hosted PKCE bounce stays transient (navigation takes over).
- On `"expired"`, the gateway connect loop clears the stored connection and bails instead of retrying forever; `NavigationGuard` routes to the connect screen, which now explains "your session expired — sign in again". Transient failures keep the existing backoff loop.

The vestad-served web bundle (`VITE_VESTAD_HOSTED=true`) has no host field, so recovery there is just re-pasting the API key; hosted (vesta.run) boxes bounce through PKCE as before.

## Tests

- `rotation_slides_the_family_expiry` + `retry_grace_slides_the_family_expiry` (vestad); existing rotation/revocation/pruning tests unchanged and still pass.
- New `token-refresh.test.ts` covering all classification paths: fresh skip, successful rotation persists tokens, 401 → expired, 5xx → transient, network error → transient, hosted → PKCE bounce.
- `./check.sh web` fully green (83 tests). vestad is Linux-only and this machine has no Docker/cross toolchain, so the Rust suite rides on CI; the pure rotation logic + full test module were verified locally in a scratch crate (6/6 pass).


elyxlz:
🤖 babysit-prs

Was failing on:
- merge conflict in `apps/web/src/components/Connect/index.tsx`: master's #807 (recompose connect front door) replaced the heading text with `<LogoText>` branding in both the hosted sign-in and key-entry forms, exactly where this PR added its session-expired messaging.

Pushed [`e17c3bb8`](https://github.com/elyxlz/vesta/commit/e17c3bb85cd20c730440d9f695ad9aa1e3e81ec2):
- `Connect/index.tsx`: kept master's `<LogoText>` in both forms and layered the PR's session-expired notice underneath, rendered only when `sessionExpired` is true. Non-expired flows match master's redesign exactly; expired flows still get the "your session expired, sign in/connect again" affordance. Used a comma instead of an em dash per repo prose style.

PR #840, opened 2026-06-19
fix(vestad): backfill required env keys into legacy agent env files

elyxlz:
## Problem

After updating to **0.1.160**, agents created by an older vestad crash-loop on boot:

```
AttributeError: 'VestaConfig' object has no attribute 'agent_model'
  at core/provider.py → return "claude", config.agent_model
```

Commit #825 made `AGENT_MODEL` / `AGENT_PROVIDER` *required-from-env* (the agent stopped re-defining their Python defaults; vestad seeds them at creation). But vestad writes the agent env file **only at creation**, and reconcile only *recreates* it when it's missing, never backfills an existing one. So a legacy env file written before #825 has neither key, and the new core reads `config.agent_model` unconditionally and crashes. Upgrading the daemon swaps the binary and the mounted core but never recreates the container, so it never self-heals.

Found this debugging a real remote agent stuck in `Restarting` (multiple agents on the host were affected).

## Fix

`reconcile_containers` (Phase 1) now backfills any required key missing from an **existing** env file, appending the canonical default. It:

- leaves every existing value, the generated `AGENT_TOKEN`, the allocated `WS_PORT`, and legacy aliases untouched (no token rotation, no silent reset of a user's choice),
- treats a present legacy alias (`AGENT_SEED_PERSONALITY`) as satisfying `AGENT_PERSONALITY`, so a non-default personality is never overwritten,
- relies on the existing Phase 3 restart to re-source the updated file, so legacy agents self-heal on the next vestad start.

This is a deliberate **backfill-merge**, not a full overwrite: a blind rewrite-from-defaults would rotate the token and reset user-chosen values (personality/timezone), trading a loud, instantly recoverable crash for a quiet, late-noticed regression.

## Single source of truth for defaults

Keys, defaults, and accepted aliases now live once in `defaults::REQUIRED_AGENT_ENV_DEFAULTS`, consumed by **both** the creation seed in `write_agent_env_file` and the new backfill. Adding a future required var to that one list makes it both seed for new agents and backfill for old ones.

## Fleet impact

This *is* the fleet-healing mechanism: on the next daemon start, every managed agent with a legacy env file gets the missing keys, additively. No separate migration needed (the change touches on-disk config state and heals it in place).

## Tests

- `backfill_adds_required_keys_missing_from_legacy_env` — model+provider added; personality preserved via alias (non-default `warm` kept); token/port untouched
- `backfill_is_noop_when_all_required_keys_present` — byte-for-byte unchanged
- `backfill_adds_personality_when_no_alias_present` — ancient file gets the default
- `write_agent_env_file_seeds_all_required_keys` — refactor regression + creation-time personality override

`./check.sh vestad` green (139 passed).

### DIFF
--- a/vestad/src/defaults.rs
+++ b/vestad/src/defaults.rs
@@ -13,6 +13,28 @@ pub const DEFAULT_PROVIDER: &str = "claude";
 pub const DEFAULT_MODEL: &str = "opus";
 pub const DEFAULT_CONTEXT_TOKENS: u32 = 1_000_000;
 
+/// An env key the Python agent now requires, its canonical default, and the legacy env
+/// names the agent still accepts for it.
+///
+/// This is the single list both env paths read from: `write_agent_env_file` seeds these
+/// into a brand-new agent's env, and `backfill_agent_env_file` adds any that are missing
+/// from a legacy env file written before the key became required (notably AGENT_MODEL /
+/// AGENT_PROVIDER, added when the agent stopped re-defining their defaults). If any
+/// `alias` is already present, the key counts as satisfied and is not backfilled, so a
+/// non-default legacy value (e.g. AGENT_SEED_PERSONALITY) is never silently reset.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct RequiredEnvDefault {
+    pub key: &'static str,
+    pub default: &'static str,
+    pub aliases: &'static [&'static str],
+}
+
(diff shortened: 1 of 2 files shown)


elyxlz:
Superseded by #844, which unifies config into three buckets (identity / preferences / provider-auth). The crash this PR's backfill addressed is now impossible by construction: agent config defaults come from the shipped defaults.json, so nothing is required-from-env. Closing in favor of the deeper fix.

PR #864, opened 2026-06-24
feat(agent): drive Claude through the official Claude Agent SDK; keep cc_sdk dormant

elyxlz:
## What

Brings the agent back to the official **Claude Agent SDK** (`claude-agent-sdk`), which `core/` used before #623. cc_sdk was deliberately built as a drop-in for the SDK's public surface, so this is mostly an import flip (`core.cc_sdk` → `claude_agent_sdk`) plus a few fixups. **cc_sdk is kept in the tree, dormant**, as a ready alternative driver.

## How it works now

- **Imports flipped** in the 6 consumer files (`client.py`, `loops.py`, `models.py`, `tools.py`, `config.py`, `sdk_parsing.py`). cc_sdk's directory is untouched.
- **Options**: dropped the three cc_sdk-only fields (`max_context_tokens`, `context_window`, `expanded_context_window`). The official `ClaudeSDKClient.get_context_usage()` returns the same `{percentage, totalTokens, maxTokens}` shape, so context-usage logging is unchanged; the OpenRouter window + 1M beta still flow via `CLAUDE_CODE_MAX_CONTEXT_TOKENS` env + `betas` (both native to the official options).
- **cc_sdk-only client members** (`is_alive()`, `session_id`, `compact()`) are guarded with `try/except AttributeError` at their three call sites so the watchdog, error-path session-id capture, and dreamer compaction degrade gracefully on the headless SDK (sanctioned over the banned `getattr`).
- **Removed tmux-coupled features** that can't work on the headless SDK: the diagnostics **pane-tail** and the **`vestad attach`** command (+ their tests).
- **Image**: no longer pins/symlinks a `claude` binary — the SDK resolves its own bundled Claude Code CLI. `tmux` stays installed so cc_sdk remains runnable if reactivated.

## cc_sdk stays — leave it alone

Nothing in `core/` imports it. It's retained as a drop-in CLI-driving alternative; `CLAUDE.md` and `agent/core/cc_sdk/README.md` both say not to delete/refactor/rewire it without an explicit ask. Its transport tests (`test_cc_sdk`, `test_e2e_transport`, `test_stop_race`) still run and keep it healthy.

## Verification

- `./check.sh agent` — green (321 passed; the 3 removed tests are the pane-tail ones).
- `./check.sh vestad` — green (133 passed; `attach` command + tests removed).
- Official-SDK surface parity confirmed by introspecting `claude-agent-sdk` 0.2.108: every top-level name and `types` symbol `core/` imports is present, `ClaudeAgentOptions` has `betas`/`env`/`skills`/`setting_sources`/`stderr`/`can_use_tool`, and `ClaudeSDKClient.get_context_usage()` exists.
- The real interactive contract (transcript/hooks/MCP over a live agent) is gated by the release-only `test-live` e2e, as before.

## Fleet impact

No on-disk state or config format change, so **no migration**. `pyproject.toml` + `uv.lock` are bind-mounted into agents and `uv sync --frozen` runs at boot (`docker.rs`), so existing fleet agents install `claude-agent-sdk` on their next restart.

## Risk note

The SDK relies on its bundled Claude Code CLI being present/arch-correct in the container; the release `test-live` job boots a real agent from the freshly built image and gates publishing, which is the backstop here.


elyxlz:
Update: addressed the dreamer compaction question. The headless SDK has no `client.compact()`, so I reimplemented it via the documented slash-command path — `compact_session()` sends `query("/compact")` and drains `receive_response()` until the `SystemMessage(subtype="compact_boundary")`. **Verified live against real Claude**: a built-up conversation compacted 22469→2377 tokens (trigger=manual) and the session retained its earlier context across the boundary. The dreamer's compact-then-restart-resuming-the-compacted-session behavior is preserved (not dropped).

emipasca:
Check if the provider state specifically auth has not regressed. Updates automatically when claude auth expires/failed etc

PR #879, opened 2026-06-24
fix(vestad): confirm container removal before recreating in rebuild_agent

elyxlz:
## Problem

A cross-version upgrade makes vestad **rebuild** the agent container (`needs_rebuild` → spec changed between versions): `rebuild_agent` does `[1/4] stop → [2/4] snapshot → [3/4] remove old → [4/4] create new`, recreating under the **same name**.

`[3/4]` was best-effort:
```rust
remove_container_force(docker, &cname).await.ok();
```
`docker rm --force` can return before the name is actually released, and a transient daemon error leaves the container present. Either way `[4/4] create` then **collides on the name** and `rebuild_agent` errors. `reconcile_containers` treats a rebuild error as non-fatal (logs + continues), so **vestad comes up healthy while the agent stays stopped on the old image** — an intermittently broken upgrade.

Surfaced by the upgrade e2e (#872): the agent container was left `Exited (0)` on the old image with a fresh `vesta-rebuild:` snapshot present — exactly the stop+snapshot-but-not-recreated signature.

## Fix (one change, at the root)

`ensure_container_removed`: retry the force-remove and **poll until docker no longer reports the container**, erroring loudly if it never disappears. `rebuild_agent` `[3/4]` uses it and **propagates** the error, so `[4/4] create` only runs once the name is free — no silent agent-down, no confusing name conflict. No redundant retry added to create (the removal is the root cause).

`rename_agent` removes the OLD name and creates a NEW one, so it can't self-collide — left as-is.

## Tests

- Structural: `rebuild_agent` confirms removal before create and no longer uses the swallowing `remove_container_force` (matches the existing `rebuild_agent_stops_running_container_before_snapshot` convention).
- Docker-gated (`#[ignore]`): `ensure_container_removed` removes a real container and is idempotent when already absent.

All 118 vestad unit tests pass; clippy clean.

### DIFF
--- a/vestad/src/docker.rs
+++ b/vestad/src/docker.rs
@@ -121,6 +121,11 @@ pub(crate) fn agent_container_entrypoint_cmd() -> Vec<String> {
 
 const CONTAINER_STOP_TIMEOUT_SECS: i32 = 10;
 const CONTAINER_RESTART_TIMEOUT_SECS: i32 = 10;
+/// `docker rm --force` can return before the container name is actually released, and a transient
+/// daemon error can leave it present; poll-and-retry until it's gone so a follow-up create under the
+/// same name (rebuild_agent) can't collide. Bounded so a genuinely stuck removal fails loudly.
+const CONTAINER_REMOVE_MAX_ATTEMPTS: u32 = 5;
+const CONTAINER_REMOVE_POLL_MS: u64 = 200;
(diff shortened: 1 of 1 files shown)


elyxlz:
Folded into #872 so the upgrade e2e ships together with the reliability fix it surfaced (the test exercises the rebuild path the fix repairs). Re-verifying with repeated runs there.

PR #921, opened 2026-06-29
feat(whatsapp): managed WhatsApp via vesta.run — foolproof daemon + anti-ban hardening

elyxlz:
# Managed WhatsApp on the agent side: foolproof daemon, residential egress, anti-ban hardening

This is the agent half of headless "managed" WhatsApp (server half: elyxlz/vesta-cloud#120; they merge together). It lets a hosted vesta.run box run WhatsApp on its own pooled number with no phone farm, no eSIM, no QR scan, and no user step, while keeping the pre-existing self-hosted path where the user links their own WhatsApp by QR. In every mode Vesta drives the account from an embedded whatsmeow client. All of it lives under `agent/skills/whatsapp/cli/` (a Go CLI that self-manages its own daemon), with the operator and design notes in `SKILL.md`, `MANAGED_AUTH.md`, `SETUP.md`, and `DEVELOPING.md`, plus a rewritten `birth` onboarding arc and a WhatsApp-first hand-off in `onboard`.

The design goal throughout is that the agent runs one idempotent verb (`whatsapp connect`) and the daemon owns every step: claim, pair, link, egress, and recovery. Nothing is orchestrated turn by turn by the model.

---

## What changed

- A second auth paradigm (managed pooled number) alongside the existing QR self-link, selected once at construction.
- A single `connect` setup verb that drives the whole claim to pair to link handshake synchronously inside the daemon.
- A pool-API client (`managed_auth.go`) with typed terminal-state classification, talking either directly to a self-hosted home box or through the vesta.run control plane.
- Residential-proxy egress for the companion, with an IP-intelligence check that hard-fails a datacenter or hosting exit before WhatsApp ever sees it.
- A structural ban-avoidance layer on the client: reply-first send gate, cold-call gate, and sliding-window pairing caps.
- A foolproof single-instance daemon: exclusive flock, screen supervision, park-on-conflict, device-preservation reconnect, corrupt-state salvage, and one consolidated `state.json`.
- A bring-your-own-number flow where the user owns the primary on their own phone and the agent links only as a companion.
- Configuration renamed to `DOUBLETICK_*` with a temporary legacy fallback, and an agent-authored `--opener` greeting carried into the click-to-chat link.
- A rewritten `birth` onboarding conversation (name first, then Vesta introduces themselves and their purpose, then capabilities, then the user's goals).

---

### 1. One paradigm decision, made once

The skill picks between two auth paradigms exactly once, at construction, so nothing branches on "which mode" inline afterward.

- `linker` is the box's pairing strategy. `qrLinker` links the user's OWN WhatsApp (QR or phone code); `managedLinker` links the agent's own pooled number (its credential selection hidden inside `managedAuth`). Each serves only the operations its paradigm supports and rejects the rest with a clear message.
- `chooseLinker(cfg, state)` is the single decision. It reconciles direct-mode pool credentials against persisted state first (env creds win and are persisted, otherwise persisted ones backfill, so an env scrub still selects managed), then returns `managedLinker` when `isHosted()` is true, else `qrLinker`.
- `isHosted()` is true for a direct home-box key (`DOUBLETICK_API_URL` + `DOUBLETICK_API_KEY`) OR a genuine cloud tenant. The cloud signal is `cloudManaged`, derived from `VESTA_CLOUD_CONTROL_URL`, which the control plane's cloud-init sets only on managed VMs. Every agent container carries `VESTAD_PORT` / `AGENT_NAME` / `AGENT_TOKEN`, so their presence alone cannot tell a paid tenant from a plain self-hosted box; without the cloud signal a plain box falls back to QR rather than dead-ending on a managed path whose token mint would 404.

Two ways reach the same pool API, differing only in base URL and credential. Direct (self-hosted) sends a static per-account `wak_` key straight to the home box. Cloud mints a short-lived server-identity token from the box's own vestad and sends it to `vesta.run/api/integrations/whatsapp`, which forwards to the home box. Both hit the identical native paths.

### 2. managedAuth: the pool-API client and wire contract

`managed_auth.go` is pure transport plus terminal-state classification, holding no disk state.

- `mintToken()` POSTs `https://localhost:$VESTAD_PORT/agents/$AGENT_NAME/account-token` with `X-Agent-Token`; vestad signs it locally from the box api_key (no network hop, no standing WhatsApp secret on a cloud box). `authorize()` resolves base and header once and reuses the token across a multi-request command.
- `do()` marshals the body, decodes a 2xx, and turns any non-2xx into a typed `*httpError{Status, Key, Body}`, so callers branch on the status and machine key rather than an opaque string.
- `classifyBlock()` maps the control plane's real signals: 403 `account_banned` to `errBlocked` (unusable, re-run for a fresh one), 409 `account_restricted` to `errRestricted` (self-clearing, wait then re-run). The reauth `/pair` error path maps the same way, so a number banned AFTER linking re-homes rather than looping.
- `claim()` POSTs the idempotent `/provision`. A dry pool answers 409 with no number, and claim re-POSTs until one binds, bounded by wall clock (not iteration count) so a slow control plane cannot surface a spurious daemon timeout. Never binding returns `errPoolFilling`; a banned number returns `errBlocked`.
- `reauth(st, pairPhone)` mints a fresh pairing code for the existing number and re-links the same companion: no new number, no OTP, no user action. This is the reconnect path after the companion-unlink window lapses.

### 3. The single `connect` verb

`whatsapp connect` is the one setup verb (hidden `provision` / `link` aliases route to it), idempotent and safe to re-run until `whatsapp status` shows linked. It makes the same paradigm choice `chooseLinker` does and dispatches to the hosted or QR flow. `connect [--opener <text>]` lets the agent author the first-contact greeting; `connect --own-number` selects the bring-your-own path.

- `runProvision` cold-starts the daemon (waiting on the socket through the first-boot recompile, up to 5m) then runs the synchronous `provision` socket command, which drives the WHOLE claim to pair to link handshake inside the daemon and returns a terminal result.
- `managedLinker.provision` claims the number FIRST, off the pairing socket (a dry-pool poll can run a while and whatsmeow closes the login websocket about 160s after Connect, so claiming first keeps the pair well inside that window). It always re-consults the idempotent `/provision`, which also auto-heals a post-link ban onto a fresh number. It routes through the verified residential egress, opens the pairing WS (draining the unused QR channel so the emitter never blocks during the long server-synchronous `/pair`), mints the code through `guardedPairPhone`, posts it, and waits for the companion to finish. Every failure path `Disconnect`s so the next attempt starts clean.
- **Record-on-dispatch**: `guardedPairPhone` checks the ban cap BEFORE minting, verifies the socket is up, then records the attempt on dispatch (before awaiting the response), so a request that reaches WhatsApp but errors mid-handshake still burns a slot while a pre-dispatch failure burns none.
- **Fresh vs resume onboarding**: `cmdProvisionManaged` captures the prior number before provisioning. A first-ever claim (or a heal onto a different number) returns `managedLinkedResult` with the reply-first onboarding copy and the `wa.me` link; re-linking the same established number returns `managedResumeResult`, which omits onboarding (telling Vesta to stop initiating and re-share a link mid-relationship would be wrong).
- Terminal statuses are clean, never raw errors: `provisioning` (pool filling), `blocked` (banned), `rate_limited` (restricted or capped), `linked`, each with a `next:` step.

### 4. The agent-authored opener

The click-to-chat link the agent surfaces to the user carries a prefilled first message. `whatsapp connect --opener "<text>"` lets the personality author a warm, name-bearing greeting for that link; `connect` threads it through to the daemon's `provision` command and into `waMeLink`. Empty or whitespace falls back to the default `"Hey! It's me, connecting here on WhatsApp."`. Only the fresh-link result embeds the opener; the resume result never does.

### 5. Companion residential egress (fail-closed, datacenter hardfail)

A companion egressing from a datacenter IP is itself a ban signal, so the client verifies its egress before every Connect and refuses to run WhatsApp over a hosting exit.

- Precedence: an explicit bring-your-own `WHATSAPP_PROXY_URL` wins in every mode; otherwise a genuine cloud tenant leases a residential proxy from doubletick (`/proxy/lease`, reusing the same server-identity token); otherwise a direct connection is used as-is.
- Whatever path is selected, `ensureManagedProxy` inspects the real egress through that path (an IP-intelligence lookup) and treats a `hosting` exit as a hard datacenter signal. A residential lease may legitimately report `proxy=true`, so hosting is the discriminator, not proxy.
- If a direct connection is detected as a datacenter exit and the box is hosted, it upgrades to a lease and re-verifies through it. If egress still resolves to a datacenter or hosting exit, the connect fails closed with an actionable message (obtain a residential proxy and set `WHATSAPP_PROXY_URL`). The lease is cached under a mutex for the process lifetime (sticky per account) and applied via whatsmeow `SetProxyAddress` BEFORE Connect, since it must precede Connect to take effect.
- Fail-closed means the daemon refuses to connect rather than leaking the datacenter IP; the socket server still comes up for `whatsapp status`.

### 6. User-owned (BYO) number

`whatsapp connect --own-number` is a distinct flow where the USER owns the primary on their own phone and the agent links only as a companion. `provisionSelf()` reserves a pool number (`POST /byo`); `selfNumberOTP()` GETs `/byo/otp`, blocking until the SMS code arrives. The command relays both to the user (register the number on your own phone, then enter this code), then links the agent as a companion through the normal QR page path. Because the user owns the primary and keeps their phone online, they reauth when it drops, and normal messaging applies: no `wa.me` onboarding and no reply-first restriction (the number is theirs, not a fresh pooled one).

### 7. Ban-avoidance on the client

A live run proved a freshly provisioned account can be banned within about 15 minutes; the trigger is a minutes-old number cold-messaging a non-contact. The client enforces structural guards.

- **Reply-first send gate**: `requireSendAllowed(jid)` gates every outbound (text, file, audio, and voice call). It layers the saved-contact check with `requireReplyFirst`, which on a MANAGED number additionally requires a prior inbound from that chat (`hasInboundMessage` checks the OLDEST stored message is inbound, proving the peer messaged first). A self-hosted number carries no such rule.
- **Voice-call cold-initiate gate**: `CallManager.Place` passes the same gate before dialing, so a managed number never cold-calls (the strongest ban trigger).
- **Pairing rate-limit cap**: `checkPairAttempt` enforces a sliding window, widest first: hard caps of 6/7d and 3/24h that `--acknowledge-ban-risk` does NOT bypass, plus a soft 2/h cap that only a self-hosted link can override for a legitimate retry (a managed number has no override). Combined with record-on-dispatch, a mid-handshake failure cannot bypass the caps.
- **@lid preservation** on reply keeps the full sender JID (including a hidden-user `@lid`) so replies actually deliver (rebuilding a phone JID from a LID's digits is accepted locally but never delivered).

### 8. Daemon lifecycle

- **Single-instance lock**: `runServe` takes an exclusive non-blocking flock on the per-instance `daemon.lock` BEFORE opening the whatsmeow store; a held lock exits `already_running`. Two connected daemons on one device store (the root of the device-session conflict) become structurally impossible, and the lock is held process-wide so the GC never releases it.
- **screen supervision**: `startDaemonProcess` launches `whatsapp serve` under `screen -dmS` and waits for the socket, distinguishing a slow first-boot compile from a dead session.
- **Single-JSON restart**: `stopDaemon()` does the stop work and returns a status silently, so `daemonRestart` composes stop and start and emits exactly one JSON object. Restart replays the last run's serve flags from `state.json`, so `--read-only` is never dropped. A `stop-requested` marker tells the dying process the stop was intentional so it suppresses the `daemon_died` notification.
- **Post-link sync window**: after a fresh link, stop/restart are refused for 5m, because restarting inside the fragile history-sync window logs the device out and forces a re-pair. `--force` overrides with explicit acceptance.
- **Park on conflict**: a `StreamReplaced` (another device took over) parks the daemon, recording `ConnParked=true` and staying up without reconnecting. Every reconnect path reads the park and refuses, and `Connect` hydrates the flag at boot, so an auto-restarted daemon cannot steal the session back and ping-pong. Only a deliberate `whatsapp connect` clears it.
- **Device-preservation reconnect**: whatsmeow deletes the local device store the instant the server sends a (frequently spurious) `device_removed`, so recovery leans on a periodically refreshed `.good` snapshot (`VACUUM INTO` on a read-only connection, throttled). On removal, if a snapshot exists and a single-retry guard is clear, the daemon re-execs in place (same PID, so the screen session survives) and restores the snapshot BEFORE opening the store: a reconnect, not a re-pair, so no ban risk. A genuinely dead device re-drops fast, the guard blocks a second attempt, and it falls through to record logout and exit for a deliberate connect.
- **Corrupt-state salvage**: if `state.json` is unreadable it is preserved as `state.json.corrupt` and the daemon starts from a safe base (not the legacy migration, which would clobber the damaged file and lose the number and creds). `salvageCorruptState` leniently decodes just the ban-avoidance caps, and if even that fails returns a cap-exhausted window, so a corrupt file can never silently reopen the hard pairing rate limit into a clean slate.
- **State consolidation**: the former per-concern files (managed-auth, auth-status, last-exit, daemon-info, pairing-attempts, linked-at) fold into one atomically written `state.json` per instance, owned solely by the serve process. The legacy migration carries a grep-able `LEGACY(remove-when: ...)` marker.

### 9. Onboarding: the birth arc and WhatsApp-first hand-off

- **`birth` rewrite**: the first conversation now opens by asking the user's name and acknowledging it warmly (a genuine compliment when it fits, never forced for a common name). Vesta then introduces themselves as a Vesta, a guardian angel sent to give people back their time and help them reach their goals, paints what life with a Vesta looks like from the breadth in MEMORY.md, and only then segues into drawing out the user's specific pain points and goals. The arc keeps the same silent housekeeping and channel/email/capability follow-ups as before. Brand voice is explicit in the prompt: a Vesta is never "it" and never gendered.
- **`onboard` WhatsApp-first hand-off**: when an existing Vesta walks a stranger through signup inside a WhatsApp chat, the new Vesta cannot message the user first, so the new Vesta sets up their own WhatsApp, writes the `wa.me` connect link to a `file-host` path, and the onboarding Vesta polls that public URL and relays the tappable link. The whole coordination rides the seed context written at create-agent; no new mechanism is needed.

### 10. Notifications, status, migrations

- **Paradigm-aware copy**: `WriteUnpairedNotification` / `WriteLoggedOutNotification` re-resolve the paradigm (env then state, since they run without a live client). Managed copy says Vesta reclaims the number autonomously with no user step and to run `whatsapp connect` now; self-hosted copy gates on the user being ready to scan. Both say NOT to retry-loop pairing.
- **Status vocabulary**: the agent-facing surface speaks `linked` / `not linked`, and connect returns `provisioning` / `blocked` / `rate_limited` / `linked`, each with a `next:` step. The persisted auth cache carries `logged_out` so status stays truthful after the daemon goes quiescent.
- **Migrations** (`agent/core/migrations/`): one repoints the restart skill at the idempotent `whatsapp start` front door; one lets the daemon fold legacy files into `state.json` on its next natural restart (explicitly not inside the sync window). Both are idempotent and skip cleanly when WhatsApp is not installed.

---

## Testing

Go unit tests are hermetic (fake vestad plus control-plane HTTP servers, in-memory state, injected `pairPhone` / `connected` / `lookupEgress` seams) so the whole managed stack runs offline: the full claim/pair/save handshake, queued-then-bound, dry-pool to `errPoolFilling`, the wall-clock poll bound, the direct-key native path, reauth, the 403/409 to `errBlocked`/`errRestricted` mapping, `leaseProxy` success plus 503 plus the direct-vs-cloud base and account-header split, `ensureManagedProxy` fail-closed and residential-direct pass, the BYO `provisionSelf`/`selfNumberOTP` paths, paradigm selection, the pairing caps (hard caps un-overridable, record-on-dispatch), the managed cold-initiate and cold-call blocks, the park-respecting reconnect behaviors, corrupt-state salvage, the preserve decision plus snapshot round-trip, second-daemon rejection, restart flag replay, the `DOUBLETICK_*`-preferred and legacy-fallback config resolution, and the `--opener` fallback. The Python `test_whatsapp_launcher.py` exercises the real launcher shell script against a fake `go` toolchain (the mtime build cache in both directions, the serve-time staleness check, and the `update-deps` path).

---

## Migration and back-compat notes

- **Env rename**: managed config now reads `DOUBLETICK_API_URL` / `DOUBLETICK_API_KEY`, falling back to the legacy `WHATSAPP_API_URL` / `WHATSAPP_API_KEY` when the new names are unset, so an existing direct-key box or seed context keeps working. The fallbacks carry `LEGACY(remove-when: all seed-contexts/env use DOUBLETICK_*)` markers.
- **State files**: existing per-concern state files fold into `state.json` on the daemon's next restart via a marked legacy migration; a corrupt file is preserved, not clobbered.
- **Restart front door**: a migration repoints the restart skill at `whatsapp start`, which is idempotent, so a box mid-fleet-update converges without a manual step.


elyxlz:
fix ci and conflicts

elyxlz:
Rebased against master: `git merge origin/master` merged clean, no conflicts, so the branch is simply refreshed.

CI (agent-tests) fixed at the root cause. Two failures, both stale assertions this PR's launcher change deliberately outdated:

- `test_whatsapp_launcher.py`: `serve` no longer floats whatsmeow to `@latest` via `go get` (that could invalidate a live multi-device session mid-turn and forced a rebuild churn loop). It now runs a non-blocking `go list -m -u` check that only warns, and deliberate bumps go through the new `update-deps` subcommand. Updated the two tests to assert the check-and-warn path and the current-pin serve, and taught the fake `go` a `list` case.
- `test_deployment.py` brand-voice guard: dropped an em dash in `whatsapp/SKILL.md` line 71.

Verified locally: full `agent` suite green (597 passed), `ty check` clean, ruff clean on the edited test. cc_sdk tmux transport tests run in CI. No test weakened or skipped.

🤖 Generated with Claude Code

elyxlz:
fix conflicts

elyxlz:
Re-resolved the merge conflicts against master (edit/delete work #1289, CLI help fix #1294, contact-naming #1292/#1293, strict-linting #1291 all now land alongside the managed-auth daemon work).

Conflicted files this round and how each was resolved:
- **SKILL.md**: kept the PR's simplified provision/status/send/messages model and the hands-off daemon framing (advanced daemon/serve docs stay in SETUP.md, per the PR). Master's #1289 "Edited and deleted messages" section merged cleanly above and survives; the PR's condensed voice section subsumes master's voice reword.
- **cli/events.go**: combined both. The PR's worker offload (`enqueueWork`) now wraps master's three-way reaction/protocol/message split, so #1289's `handleProtocolMessage` (edit/revoke) runs on the offloaded worker.
- **cli/cli.go**: kept BOTH struct fields (master's `hidden` + the PR's `timeout`); took the PR's single blocking `link` (`cmdLink`) that replaces the old link-start/status/stop trio; marked `provision`/`link`/`daemon-status` `hidden` so master's #1294 registry-driven usage listing does not double-list the lifecycle commands the header already documents.
- **cli/main.go**: `main()` auto-merged carrying both master's #1294 per-command `--help` path and the PR's provision/start/status dispatch switch. `printUsage` keeps master's registry-driven command loop under the PR's "Setup / health" + "Internal" headers.
- **cli/client.go**: kept the PR's `recoverOrRestart` (force-drop-then-reconnect) body and comment, folding in master's StreamReplaced note.
- **tests/test_whatsapp_launcher.py**: kept the PR's warn-on-newer-whatsmeow test, which matches the merged warn-only launcher (`_run` takes `whatsmeow_update`; master's `go_get_exit` path no longer exists).
- Also trimmed two PR comments (guards.go, managed_auth.go) to satisfy master's new 8-line comment-block guard (#1291), and regenerated `skills/index.json` (it had lost the `wx` entry in the auto-merge).

Checks: `./check.sh guards` passes (shellcheck absent locally, runs in CI), `ty check` passes, launcher pytest 4 passed. The whatsapp CLI is Go and I have no Go toolchain here, so the Go build/tests were not run locally; CI will compile them. Conflicts are resolved (`mergeable: MERGEABLE`).

🤖 Generated with Claude Code

PR #935, opened 2026-06-30
feat(notifications): match interrupt rules on any notification field

emipasca:
## Summary

Notification interrupt rules could only match `source`/`type`/`sender`/`keyword`, where `sender` and `keyword` searched **hardcoded synonym lists** (`_IDENTITY_FIELDS`, `_BODY_EXTRA_FIELDS`). But the notification schema is open (`Notification` is `extra="allow"`) — each source attaches its own fields — so any field outside those lists was unaddressable. A WhatsApp/Telegram group's `chat_name`, `chat_type`, `is_group`, a mail `label` — none could be targeted. **Closes #933**: you couldn't pool one busy group chat without leaky per-member `sender` rules that also catch those contacts in other chats.

## Design — a general, scalable matcher

Replace the closed allowlist with one open primitive. A rule keeps the two universal declared fields (`source`, `type`, exact/ci) and gains a `match` list of **field predicates**, ANDed together:

```python
class FieldPredicate(BaseModel):      # extra="forbid"
    field: str                         # any notification key, OR an alias
    op: Literal["contains", "regex"] = "contains"   # ci substring | ci re.search
    value: str
    negate: bool = False

class NotificationInterruptRule(BaseModel):
    id: str = ""
    source: str | None = None
    type:   str | None = None
    match: list[FieldPredicate] = []
    action: Literal["interrupt", "pool"]
```

- **Any field is targetable by name** — `{field: "chat_name", value: "Bride squad"}`, `{field: "chat_type", value: "group"}`. A new source's new field works with **zero code changes**.
- **Operators cover the space**: `contains` + `regex` + `negate`. Values are str-coerced, so `is_group`/`chat_id` (bool/int) match too.
- **Aliases keep the ergonomics**: `sender` = the identity fields, `text` = body/message — so you needn't know each source's field name. Synonym knowledge lives in one extensible table.
- **`source`/`type` stay dedicated** because the whole subsystem (default overrides, facets, observed-pairs) keys on the `(source, type)` pair and exact-match is the right default for them.

## No data model redundancy, no migration

`sender`/`keyword` are **removed from the data model** and kept only as **edge sugar** that compiles to predicates (CLI `--sender`/`--keyword`, web pickers), so the common case stays terse. Back-compat is additive: a `model_validator(mode="before")` (core) and an on-read normalizer (skill CLI) fold legacy flat `sender`/`keyword` keys into `match`, and writes emit canonical shape — so existing fleet rules keep working and the file converges on the next edit with **no migration pass**. Both shims carry `LEGACY(remove-when: …)` markers.

## Discoverability (facets)

A general matcher is useless if you can't see what's there. Notifications now emit a `fields` map of their targetable structured extras (`{chat_name: "Bride squad", chat_type: "group"}`); the skill's `facets` command surfaces them and the web rule editor offers field-name + value suggestions.

## Surfaces touched (one PR)

- **Core** (`notification_interrupt_policy.py`): `FieldPredicate`, generalized `_matches`, declared+extra field reads, str-coercion, aliases, legacy normalization; `notif_facet_fields` extractor. `events.py`/`loops.py` emit the `fields` map.
- **Agent CLI** (`notif-interrupt-rules.py`): `--match FIELD<op>VALUE` (`=`, `~=`, `!=`, `!~=`), `--sender`/`--keyword` sugar, legacy normalization, `facets` surfaces extra fields.
- **Web** (`NotificationInterruptRulesCard`): custom field-condition rows (field + is/matches + value + not), generalized match preview, `FieldPredicate` types.
- **SKILL.md**: rewritten matching docs with concrete examples (group-chat pooling, `chat_type`, negate).

## Usage example (the issue's case)

```bash
# Snooze one busy group, leave 1:1s and other groups interrupting
notif-interrupt-rules.py add --source whatsapp --match 'chat_name=Bride squad' --action pool
```

## Tests

New coverage for predicate matching (contains/regex/negate/missing-field/coercion/aliases/AND), legacy normalization + on-disk convergence, the `--match` CLI grammar, the `fields` facet extractor and CLI surfacing, the API round-trip, and the web custom-field flow. Full agent suite green (the only related failures pre-exist on a clean tree: cc_sdk env + upstream-sync sparse-checkout, unrelated). Web tsc + eslint + prettier + vitest green.

Closes #933


emipasca:
Combined into #938 (single PR) — this branch's commits are included there. Closing in favor of #938.

PR #937, opened 2026-06-30
feat(notifications): control + reason about interrupt-rule precedence

emipasca:
## Summary

Interrupt rules are evaluated **first-match-wins** (ordered list, top = highest priority) — the right, predictable model (firewalls, ACLs, `.gitignore`). But both surfaces were **append-only**: a newly-added specific rule landed at the *bottom*, exactly the wrong place to override a broad rule above it. Now that #935 makes negation and many complex rules expressible, controlling and reasoning about order is the bottleneck. This keeps the engine unchanged and adds **control + visibility** (Option 1 from discussion).

Stacked on #935 (`feat/notification-match-predicates`) — review/merge that first.

## What it adds

**1. Smart auto-placement (predictable, position-only).** A new rule is inserted above any *strictly broader* rule (fewer conditions), so a narrow exception isn't shadowed by a broad rule that would match first. Specificity = condition count (`source` + `type` + #predicates). It only chooses the *new* rule's position; nothing else moves. The CLI (`_placement_index`) and web (`placementIndex`) use identical logic.

```bash
# broad pool first, then a narrow interrupt exception -> exception auto-lands ABOVE the pool
add --source whatsapp --action pool
add --source whatsapp --sender wife --action interrupt   # placed first, not shadowed
```

**2. Explicit placement + reorder.**
- CLI: `add --before <id>` / `--after <id>`, and a new `move <id> {--before <id> | --after <id> | --top | --bottom}`.
- Web: drag-to-reorder the rule list (a grip handle per row when there's more than one).

**3. Visibility.**
- CLI `list` prints "rules in priority order (first match wins; top = highest priority)" (on stderr, so stdout stays pure JSON).
- Web: the draft preview warns *"a higher-priority rule already catches these"* when the rule being added would be fully shadowed by rules above its placement, using the recent-notification history.

## Engine unchanged

`should_interrupt` / `_matches` are untouched — still pure first-match-wins. Specificity is used **only** to position a new rule at write time; it never ranks rules at evaluation. So precedence stays debuggable and order remains the single source of truth, now with the controls to set it.

## Tests

CLI: auto-placement (specific-above-broad both insertion orders), `--before`/`--after`, unknown-id errors, `move` (`--top`/`--bottom`/`--before`/`--after`), unknown-id move error. Web: specificity-insert places the new rule above a catch-all; drag-drop reorders and persists. Agent + web suites green (pre-existing unrelated failures only).

## SKILL.md

Documents the precedence model (first-match-wins, OR-across-rules, AND-within-a-rule), auto-placement, and the reorder commands, with examples.

### DIFF
--- a/agent/skills/notifications/SKILL.md
+++ b/agent/skills/notifications/SKILL.md
@@ -43,7 +43,14 @@ notification's other fields. Every field/condition you set must hold (AND); what
   `is_group`, `media_type`, and so on). Two aliases search across a source's synonym fields so you needn't
   know the exact name: `sender` (the identity fields) and `text` (body/message). `--sender X` and
   `--keyword RE` are shortcuts for `--match 'sender=X'` and `--match 'text~=RE'`.
-- Rules are an ordered list, **first match wins**, so put specific rules before broad ones.
+- Rules are an ordered list, **first match wins**: evaluation runs top to bottom and stops at the first
+  rule that matches, so the top rule has the highest priority. Order is the only precedence; a later,
+  more-specific rule never overrides an earlier, broader one. To OR across different fields, write
+  separate rules (a single rule's conditions are all ANDed).
+- **Placement is handled for you, but you can override it.** `add` auto-places a new rule above any
+  broader rule (one with fewer conditions), so a narrow exception is not shadowed by a broad rule that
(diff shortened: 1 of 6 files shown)


emipasca:
Combined into #938 (single PR) — this branch's commits are included there. Closing in favor of #938.

PR #941, opened 2026-06-30
Revert "feat(vestad): honor --expose-lan under systemd (#940)"

emipasca:
Reverts #940, which was merged by accident.

This restores `master` to exactly its pre-#940 state (verified: zero diff against the commit before the squash merge). No follow-up cleanup needed.

The feature work itself is preserved on branch `feat/enable-lan-systemd` and will be reopened as a fresh PR to refine before a deliberate merge.


emipasca:
Abandoning the revert — keeping the fix in master (#940). No revert needed.

PR #957, opened 2026-07-02
fix(client): resync the SDK stream when an interrupted turn outlives the drain

emipasca:
## The bug (from production logs, 2026-07-02)

Agent appears to hang mid-conversation; every `attempt_interrupt` times out; when the user sends any new message, the *previous* turn's entire output (thinking, task events, usage) floods the log in a single second, and the cycle repeats — every turn's output arrives one user-message late, forever.

## Root cause

`converse` ends a turn when `receive_response()` yields a ResultMessage — but the SDK message stream is shared across turns:

1. **Seed:** an interrupt's wind-down outlives the 5s drain window (`_INTERRUPT_DRAIN_TIMEOUT_S`). This is common: under heavy thinking/long tools the CLI acks interrupts late (18s+ observed — confirmed by a Stop hook firing 18s after the interrupt, matching the buffered result's `duration` to the second). The drain gives up and converse abandons the stream with that turn's ResultMessage still to come.
2. **Desync:** the next turn's `receive_response()` immediately yields the stale buffered output, terminates on the *stale* ResultMessage, and Vesta marks the turn complete in ~0s while the CLI is still working on the actual query. Off-by-one from then on, self-perpetuating.
3. **Why interrupts also time out:** the SDK's internal message buffer is bounded (`max_buffer_size=100`), and its single reader task processes both messages *and* interrupt acks — with nobody consuming, the buffer fills, the reader blocks, and `interrupt()` can never be acknowledged.

The current driving path (official SDK + this converse/drain code) shipped in 0.1.161; before that the tmux/cc_sdk transport had different interrupt mechanics, so this failure mode is new to recent releases — matching when it was first observed in the field.

## The fix

Enforce the real invariant — *a turn ends at its own result* — by counting: `State.results_outstanding` increments per `query()`, decrements per consumed ResultMessage.

- converse ends a turn only when a result brings the count to 0; a stale result is consumed (its late output still emitted, nothing lost), logged, and skipped via a fresh iterator
- the interrupt drain keeps draining until the count settles, the stream goes quiet, or the window expires
- the count resets when the SDK session is rebuilt

Any abandonment (slow interrupt, auth-loss break) now self-heals on the next turn instead of desyncing the agent until restart.

## Verification

- New test `test_late_result_after_drain_timeout_does_not_desync_next_turn` reproduces the production failure deterministically: **red on master** (`got responses=['late tail of the abandoned turn']` — the new turn returning the abandoned turn's output), green with the fix
- Full `tests/test_interrupts.py`: 23 passed (all 22 pre-existing behaviors intact)
- `./check.sh agent`: ruff + format + ty clean; 462 passed (3 pre-existing local-env failures in test_cc_sdk/test_upstream_sync, unrelated, fail on clean master too)

### DIFF
--- a/agent/core/client.py
+++ b/agent/core/client.py
@@ -122,8 +122,11 @@ async def converse(prompt: str, *, state: vm.State, config: vm.VestaConfig, show
     try:
         await asyncio.wait_for(client.query(query), timeout=config.query_timeout)
     except TimeoutError:
+        # No results_outstanding increment: delivery is unknown, and this raise ends in a
+        # restart (run_one -> graceful_shutdown), which rebuilds the session and resets the count.
         await attempt_interrupt(state, config=config, reason="Query timeout")
         raise
+    state.results_outstanding += 1
     diagnostics.touch_activity(state, "query_sent")
 
     responses: list[str] = []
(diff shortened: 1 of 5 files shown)


emipasca:
Closing in favor of the architectural fix tracked in #958. The counting approach here works for the seed case (interrupt-abandoned turns) but hardening it against every hole (phantom counts, unprompted CLI continuation results, chatty drains) required stacked timing heuristics — wrong layer. The right fix is a single long-lived stream consumer that never abandons the stream; see the issue. This branch is kept for its deterministic repro + regression tests.

PR #959, opened 2026-07-03
fix(client): one long-lived SDK stream consumer ends the per-turn desync

emipasca:
## Problem (#958)

`converse` consumed the SDK stream in per-turn `receive_response()` bites and abandoned it between bites — the post-interrupt drain gave up after 5s. When an interrupted turn's wind-down outlived that drain (routine under heavy thinking / long tools / 1M contexts), its remaining output and `ResultMessage` sat unconsumed:

1. the next turn's `receive_response()` terminated on the **stale** result in ~0s → every turn's output arrived one user message late, indefinitely;
2. with nobody consuming, the SDK's bounded buffer (100 msgs) filled and blocked its single reader task → interrupt acks and `get_context_usage` responses could never be processed (the `SDK interrupt timed out` / `get_context_usage hung` storm in the production logs).

Counting outstanding results was tried and abandoned (#957): stream-json has no query↔result correlation ID and self-initiated CLI continuation turns emit unprompted results, so any bookkeeping degenerates into stacked timing heuristics.

## Fix: never stop consuming

One **`consume_stream` task per client session** reads `receive_messages()` from connect to disconnect and dispatches everything on arrival (emit text/thinking, persist session ids, detect auth loss, log usage). Turn boundaries become advisory **`TurnSignals`**: `converse`/`compact_session` send their query and wait on the open turn's `done` event; a `ResultMessage` closes whichever turn is open, and one arriving with **no** open turn is dropped.

- Nothing can sit unconsumed, jam the SDK's reader, or flush late — by construction, independent of interrupt rate, context size, or load.
- Worst case under weird CLI behavior (a result landing after the bounded post-interrupt grace) is a briefly mislabeled turn boundary: every message still reaches the user in real time, and the labels self-heal at the first idle gap (the final stray result finds no open turn and is dropped).
- The interrupt path no longer abandons anything; the entire drain apparatus (`_STOP`, fresh-iterator drain, `_INTERRUPT_DRAIN_TIMEOUT_S`) is deleted. Silence timeout keeps the exact same budget (`response_timeout` measured from the last stream message).
- `converse` still returns the turn's collected text, so the dash-correction contract and `loops.py` orchestration (boot turns, compaction deferral, notification file lifecycle, idle triage) are untouched.

## Tests

New regression coverage (fast tier, fake stream + real consumer):
- the #958 seed: late result after an interrupt does **not** wedge later turns (real-time delivery + self-heal asserted)
- unprompted continuation result while idle is dropped; next turn unaffected
- wind-down inside the grace stays attributed to the interrupted turn
- stream death mid-turn surfaces through the turn (restart path); silence raises `TimeoutError`
- compaction boundary + result via the consumer

Existing converse/dreamer/watchdog/processor suites ported to the stream harness. `./check.sh agent` green (485 passed, incl. cc_sdk tmux transport suite).

## Field evidence

Both agents on the shared box hit this: `elio-vesta` had interrupt timeouts since 06/29 (the day after the new SDK path went live), `apollo` first wedged 07/02 23:01 after a WhatsApp-daemon-death notification storm (interrupts every 1–2 min) during long 1M-context turns — matching the issue's timeline exactly.

Closes #958

### DIFF
--- a/agent/core/client.py
+++ b/agent/core/client.py
@@ -152,6 +152,13 @@ async def _dispatch_message(msg: Message, *, state: vm.State, config: vm.VestaCo
         state.compacting = False
     if isinstance(msg, SystemMessage) and msg.subtype == "compact_boundary":
         logger.client("Compaction boundary reached")
+    # The CLI streams a thinking_tokens counter while the model reasons (dozens per turn, dropped
+    # from the log by parse_sdk_message). Track it on the open turn so the wait loop's liveness
+    # notes can say "thinking, ~N tokens" instead of guessing what the quiet means.
+    if turn and isinstance(msg, SystemMessage) and msg.subtype == "thinking_tokens":
+        if isinstance(msg.data, dict) and "estimated_tokens" in msg.data and isinstance(msg.data["estimated_tokens"], int):
+            turn.thinking_tokens = msg.data["estimated_tokens"]
+            turn.thinking_tokens_at = time.monotonic()
     texts, thinking_blocks, session_id = sdk_parsing.parse_sdk_message(msg)
     if session_id and session_id != state.persisted.session_id:
         if state.persisted.session_id:
@@ -162,6 +169,8 @@ async def _dispatch_message(msg: Message, *, state: vm.State, config: vm.VestaCo
         for block in thinking_blocks:
             _emit_thinking(block, state=state)
     text = "\n".join(texts) if texts else None
+    if turn and (text or thinking_blocks):
+        turn.last_visible_at = time.monotonic()
     if text:
         if turn:
             turn.texts.append(text)
@@ -261,7 +270,7 @@ async def converse(prompt: str, *, state: vm.State, config: vm.VestaConfig, show
                 if time.monotonic() - turn.last_message_at >= config.response_timeout:
                     await attempt_interrupt(state, config=config, reason="Response timeout")
                     raise TimeoutError
-                diagnostics.note_stream_silence(state, noted_at=noted_silence)
+                diagnostics.note_turn_liveness(state, turn=turn, noted_at=noted_silence)
                 continue
 
             if done_task in done:
--- a/agent/core/client.py
+++ b/agent/core/client.py
@@ -3,6 +3,7 @@
 import asyncio
 import datetime as dt
 import os
+import time
 import typing as tp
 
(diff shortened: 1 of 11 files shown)


emipasca:
Added in 08afd770: **liveness notes for quiet thinking stretches** (no partial messages — complete-message delivery is unchanged).

During a long extended-thinking turn the log used to be minutes of dead air followed by one massive thinking block. Now the turn's wait loop wakes every 15s and logs at most three lines per quiet stretch:

```
INFO  Model quiet for 60s (thinking or long generation) | idle=65s, last_activity=query_sent
INFO  Model quiet for 120s (thinking or long generation) | idle=125s, last_activity=query_sent
WARN  Model quiet for 300s (thinking or long generation) | ...   <- + error event to the app
```

A running tool downgrades the note to debug (tool lines already show liveness), and the noted set clears when the stream talks again.

This *replaces* `sdk_watchdog`: that background task has never fired once in production on any build (verified across two agents' full log history; the identical code fires in a faithful repro — root cause never found). The notes now live inside the loop that completes turns, so they structurally cannot stop running while turns still work. The watchdog task, poll machinery, and the always-unknown `process_alive` probe are deleted; net −44 lines.

emipasca:
e79e2946 makes the liveness notes **specific** — and closes the watchdog mystery.

Key discovery (h/t Emi pointing at #939): the CLI already streams a `thinking_tokens` system message throughout extended thinking — dozens per turn, dropped from the log as spam since #939. Two consequences:

1. **The stream is rarely silent during a think**, so any idle-clock mechanism never fires. That's why `sdk_watchdog` never fired once in production (my repro used a truly-silent fake stream — mystery closed), and the previous commit's notes had the same blind spot.
2. **The counter is exactly the specific signal we wanted**, no `include_partial_messages` needed.

Now: the quiet clock is time since last *visible* output, the counter is tracked on the open turn (never logged per-delta), and the note says what's actually happening:

```
INFO  Thinking for 60s (~2,340 tokens so far)     <- counter ticking: never escalates
INFO  Model quiet for 60s | idle=..., last_...    <- no counter, calm at 60/120s
WARN  No output and no stream activity for 300s   <- genuine dead air: + error event
DEBUG No output for 60s (tool in flight)          <- tool lines already show liveness

PR #965, opened 2026-07-03
feat: vestad-local workspace distribution with upgrade-driven sync

emipasca:
Distributes the agent's home (engine + skills) to boxes **from the vestad daemon itself**, over the loopback — no github branch, no external network, identical in dev/tests/prod. Replaces the original github-`agent-workspace`-branch design mid-PR (see the two specs); the pivot is why the branch is large.

Specs: [upgrade+engine (Parts 1–3)](docs/superpowers/specs/2026-07-03-upgrade-upstream-sync-design.md) · [vestad-local distribution (Part 4, supersedes the github branch)](docs/superpowers/specs/2026-07-03-vestad-local-workspace-distribution.md).

## The model

**Engine** — `agent/pyproject.toml` + `uv.lock` live in `agent/core/`, so the engine is one publishable, one-mountable unit. vestad mounts a single read-only `core/`; dev-tool configs split out and never ship.

**Distribution** — vestad embeds the full agent home. At startup, after `ensure_agent_code`, `workspace.rs` runs the embedded `vestad/scripts/build-workspace.sh` (one tested bash script owning all git logic; its suite is `agent/tests/test_build_workspace.py` — the *same file* vestad executes, so tests and prod can't drift) to append a snapshot commit + `agent-vX.Y.Z` tag to the per-host bare repo at `~/.config/vesta/vestad/workspace/workspace.git` and regenerate `workspace.bundle`. Boxes fetch that bundle via `GET /agents/{me}/workspace.bundle` (agent-token auth, self-scoped) using the `workspace-sync` core skill's `fetch-workspace.sh`.

**Upgrade → sync** — a vestad upgrade changes the running core version (`core/pyproject.toml`); `workspace_sync.py` turns the mismatch vs the persisted `last_synced_version` into a `[Workspace sync]` boot turn: the agent rebases its own delta onto `agent-v<version>` and records completion via `mark_workspace_synced` (unmarked turns re-fire). Fresh agents pre-mark and attach via the `workspace-sync` **SETUP.md** at birth; legacy boxes convert via the one-time `2026-07-workspace-conversion` migration.

**Box porcelain** — cone-mode sparse checkout (cone = installed skills; mount-owned `core/` invisible via `sparse.expectFilesOutsideOfPatterns`); `skills-install`/`skills-remove` are offline cone edits; the box's delta always rebases on top of stock, `git log agent-vX..HEAD` = "everything this box changed".

## Notable design points

- **Bundle build is fatal at startup.** Host **git is a hard runtime requirement** for vestad — if `ensure_agent_code` or `ensure_workspace` fails (git missing, disk error), vestad aborts startup (`exit 1`, like `validate_config_dir`) rather than serve a daemon whose boxes silently 404 on the bundle.
- **Per-host lineage** keeps rebases clean (always a real merge-base); cross-host restores use the documented transplant (`git rebase --onto agent-vNEW agent-vOLD`).
- `VESTA_UPSTREAM_REF`/`VESTA_WORKSPACE_REF` are gone (boxes need no branch name); vestad strips the stale env keys (LEGACY-marked). `upstream-pr` reads the version from the mounted core.
- The migration restores the legacy repo if its conversion attach fails, so an unmanaged legacy box at an unavailable version degrades gracefully instead of being left repo-less.

## Testing

- `test_build_workspace.py` — the real build script: append/no-op/tag-move lineage, fetchable bundle, fingerprint/exec-bit handling.
- `test_workspace_sync.py` — the real box flow against bundles built by the real script: worktree-safe attach (incl. a mounted constitution stays ignored), version-pinned rebase w/ + w/o conflicts, cone scoping, offline install, downgrade, legacy conversion + repo-restore-on-failure.
- `test_workspace_sync_turn.py` — the trigger: marker, boot turn, pre-mark, tool.
- vestad unit: `workspace.rs` builds against scratch content; bundle endpoint 404/bytes.
- Docker-gated `tests/server/workspace.rs` — a real container curls the real endpoint and attaches (the production path, end to end). Caught three real fresh-box bugs during review: exec-bit stripping, snapshot/image content skew, and the constitution not being ignored.
- `check.sh agent` (488) / `vestad` (148) / `cli` (24) green. (`check.sh web` is red on master already — prettier on two files this branch doesn't touch.)

Validated end-to-end on a live box (real upgrade → sync turn → fetch over loopback → rebase → mark → restart), which surfaced the migration dead-stock and no-conflict-rebase-pause fixes now included.


emipasca:
Heads-up found while reviewing CI exposure: the workspace migration could deadlock the first release.

`test-live` upgrades a previous-release agent (legacy `~/.git`) onto the candidate and waits for **all** migrations to converge — but at test-live time the candidate's `agent-v` tag can't exist yet (`publish-agent-branch` is gated on test-live), and on the very first release the `agent-workspace` branch doesn't exist at all. If the workspace migration required a successful attach, it would never mark, converge would time out, test-live would go red, and the branch would never get published (bootstrap deadlock).

Fixed in f55f0a78 at the migration's contract, not by reordering CI: the migration's job is retiring the legacy repo (always possible offline); an unreachable remote routes to the final step and the re-firing workspace-sync boot turn owns the eventual attach. Publishing stays gated on test-live; failed releases leave no tags.

The hermetic suites (`agent`/`cli`/`vestad`/`integration`) never touch the remote — fixture origins only.

PR #990, opened 2026-07-06
fix(client): preempt turns with a priority:"now" message so background subagents survive

emipasca:
## Problem (#982)

Notification- and message-driven preemption used the SDK's `interrupt()` control request. The CLI's headless handler for that request doesn't just abort the turn — it sweeps the task registry and kills **every running backgrounded subagent/workflow task** (`killedBy:"system"`). Probed live: an interrupt killed a background agent 0.8s after launch (`status:"killed"`, notification `"stopped"`). Interactive Esc has no such sweep, and the documented lifecycle (`stop_task` as the targeted kill; background subagents "run concurrently while you continue working", resumable, partial work never lost) promises the opposite. Full analysis on the issue.

## Fix: deliver the preempting prompt as the interrupt

The CLI's prompt queue accepts a `priority` field on stream-json user messages; a queued `priority:"now"` prompt aborts the running turn with the graceful `"interrupt"` reason — no sweep, the aborted turn still emits its ResultMessage, and the queued prompt runs next. Probed live on the bundled CLI (2.1.191, also verified in 2.1.201):

- mid-text/between-steps: turn ends within ~30ms, the preempting turn runs ~1.5s later
- a foreground tool call in flight finishes first; the latched abort applies at the step boundary (accepted trade-off — chosen over an observable-state matrix and timer escalation)
- background subagents run to completion straight through the preempt

`send_preempt` (`core/client.py`, the SDK seam) owns the wire shape and the may-preempt gates (idle / boot turn / compaction → no-op). Producers (`process_batch`) pre-send the first section's prompt and queue it with `pre_sent=True`; `converse` then waits on the turn without re-sending the query. A failed or gated pre-send degrades to plain queueing — delayed, never lost, including on a future CLI that drops the undocumented field.

**The interrupt system stays fully supported** behind `VestaConfig.preempt_mode = "interrupt"` (default `"message"`) — a first-class mode, not a transitional fallback: it preempts immediately (mid-tool included) at the cost of the headless CLI killing backgrounded subagents, for when preemption latency outranks in-flight background work. The control request also still fires on failure paths (silence/query timeout, auth loss).

## De-ghosting (same concern: how messages reach the model)

- CLAUDE.md's Message flow claimed `api.py`'s `UserEvent` feeds `message_processor` — it doesn't and hasn't: the event is history + broadcast only; the app-chat daemon (a WS subscriber) converts it to a notification file, so **all app chat reaches the model through the notification flow**. Rewritten to match reality, with a pointed comment at the emit site.
- Dropped `main.py input_handler` + the `aioconsole` dependency: #895 deleted them; the #890 merge accidentally resurrected them (dead code — never spawned). `uv.lock` refresh includes the stale post-release `0.1.168 → 0.1.169` pin.

## Tests

New `tests/test_preempt.py` (default mode): wire format (`priority:"now"`, no interrupt control request), gate matrix, pre-sent converse never double-queries, mixed-batch first-section-only pre-send, idle/failure fallbacks, processor never fires the interrupt path. Legacy-mode coverage retained in `test_interrupts.py`, pinned with `config.preempt_mode = "interrupt"`. `./check.sh agent` green (510 passed, incl. cc_sdk tmux suite).

Closes #982


emipasca:
Applied the multi-agent code-review findings (turn-order alignment, orphan-result banking, watcher retry, auth gate) and a 4-angle simplify pass (watcher-owned pre-send, protocol envelope moved to sdk_parsing, 5s write bound, shared is_unauthenticated predicate). `./check.sh agent` green — 514 passed.

PR #1049, opened 2026-07-07
fix(app-chat): stop silently dropping WS user messages

elyxlz:
Fixes #1008

## Root cause

The app-chat daemon deduped inbound user messages against a `last_seen_ts` watermark that was advanced by **every** event with a `ts` (assistant, tool, status, …), not just user messages. A later assistant event pushed the watermark past a user message that was never turned into a notification (dropped from the bounded live queue during a stall, or lost when the socket tore down). On reconnect, `_replay_missed` treated that message as `ts <= cutoff` → already-seen → skipped it. The message survived only in server-side history and never surfaced as a notification.

A related latent bug: the live path wrote a notification for every user event with **no** dedup guard, so the snapshot/live overlap window on connect could double-notify.

## Fix

- **Advance the dedup watermark only on user messages**, via a single `_notify_if_new(state, text, ts)` helper shared by the live and replay paths. Assistant/tool/status events can no longer move it, so replay reliably re-notifies any un-notified user message. The shared helper also makes the connect overlap idempotent.
- **WS heartbeat on both sides** (agent server in `api.py`, daemon client in `daemon.py`): a half-open socket (phone off the network with no close frame) is now detected and torn down within the heartbeat interval, so the client reconnects and replays promptly instead of the server buffering to a dead peer for minutes.

### On the suggested per-client send buffer

The issue also suggested a bounded per-client send buffer. Deliberately **not** added: aiohttp already buffers writes at the transport, and the event bus already bounds each subscriber's queue (`SUBSCRIBER_QUEUE_MAXSIZE`, oldest-drop). A third buffer duplicates both and runs against the architecture principle of not stacking bulkheads on this single-tenant daemon. With replay now reliable, a dropped socket is no longer data loss, so the buffer's purpose is moot.

### Fleet impact

No migration needed. The persisted `last_seen_ts` on existing boxes holds the last event seen while connected; new user messages have a strictly later `ts` and still notify, while already-processed ones (`ts <= watermark`) stay suppressed. The watermark self-corrects as new user messages arrive.

## Tests

New `tests/test_daemon.py` drives the real daemon functions against a tmp notifications dir:
- assistant events do not advance the watermark (regression for the core bug)
- a user message that arrives during a drop is replayed on reconnect even after a later assistant event
- replay skips already-notified messages; snapshot/live overlap does not double-notify
- watermark persists across restart; a message without a `ts` is still notified

### DIFF
--- a/agent/core/api.py
+++ b/agent/core/api.py
@@ -39,6 +39,11 @@ from .provider import ProviderAuthState, UsageError, clear_provider, get_usage, 
 
 logger = logging.getLogger("vesta.api")
 
+# Ping every connected client on this cadence so a half-open socket (a phone that dropped off the
+# network without a close frame) is detected and torn down promptly, freeing the subscriber and
+# letting the client reconnect and replay, rather than the server buffering to a dead peer for minutes.
+WS_HEARTBEAT_S = 30.0
+
 
 def _pending_notification_ids(config: VestaConfig) -> list[str]:
     """Notification file stems still on disk — received but not yet processed. Seeds the connect
@@ -60,7 +65,7 @@ async def _ws_handler(request: web.Request) -> web.WebSocketResponse:
     config: VestaConfig = request.app["config"]
     skip_history = request.query.get("skip_history", "") in ("1", "true")
 
-    ws = web.WebSocketResponse()
+    ws = web.WebSocketResponse(heartbeat=WS_HEARTBEAT_S)
     await ws.prepare(request)
     request.app["websockets"].add(ws)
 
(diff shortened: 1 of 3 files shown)


elyxlz:
Superseded by #1053, which merged just before this and fixes #1008 more fundamentally: it moves app-message intake in-process (`api.py` writes the `source=app-chat` notification file synchronously in `_recv_loop` via `asyncio.to_thread`) and **deletes the daemon's entire replay path** (`_replay_missed`, `last_seen_ts`) that this PR patched.

Both bugs this PR targeted are structurally eliminated there:
- The timestamp-watermark replay gap is gone with the code that had it.
- Intake no longer rides `_send_loop`, so a failed send can no longer lose a message.

Closing rather than force-rebasing, since the rebase would re-introduce exactly the machinery #1053 removed (−113 lines).

PR #1064, opened 2026-07-08
feat(browser): Camoufox stealth engine via raw BiDi (no Playwright)

elyxlz:
## Swap the `browser` engine to Camoufox (anti-detection Firefox), driven by raw WebDriver BiDi — no Playwright

**Status: DRAFT / WIP.** Foundational modules landed and tested; the engine swap itself is in progress (see checklist). The existing CDP/Chromium engine is untouched and still green, so this branch is in a consistent intermediate state.

### Why
Stock Chromium + CDP cannot reach Camoufox-level stealth, and it's structural: every JS-injected spoof is catchable by the CreepJS-tier `toString`/descriptor/**stack-frame** battery, which Camoufox sidesteps by spoofing in C++ below JS. Camoufox's own docs say Chromium fingerprints can't be spoofed ("behavior unique to V8 is impossible to spoof"). We also currently leave the CDP `Runtime.enable` domain on (`daemon.py:156`), the single highest-weight bot signal in 2026. Going Camoufox-only removes that whole class.

Feasibility verified: Camoufox reads its fingerprint from `CAMOU_CONFIG` env vars (no Python needed), hardcodes `navigator.webdriver=false` even under BiDi/Marionette, and exposes a raw BiDi WebSocket on `--remote-debugging-port` whose `id`/`method`/`params`/`event` shape maps ~1:1 onto our existing daemon loop.

### Decisions (locked with @elyxlz)
- **Single backend: Camoufox only** (no Chromium retained).
- **No Playwright** — raw BiDi client; Marionette is the documented fallback.
- **Road C**: raw BiDi transport + **vendored** accname/role JS for the snapshot (not hand-rolled).
- **Lazy fetch, no Docker changes** — Camoufox fetched + sha256-verified + extracted on first `browser launch`, cached.
- **Fingerprint = bundled real-Firefox presets**, seed-selected per profile dir.
- **CLI surface byte-identical**; only the raw escape hatch renames `cdp` → `bidi`.
- Single PR.

### Landed + tested (fast tier, fake BiDi server)
- [x] Seed-selected fingerprint presets + `CAMOU_CONFIG` chunking (`presets.py`, `test_presets.py`)
- [x] Raw BiDi client (id-correlation, error propagation, event fan-out) + in-process fake BiDi server (`bidi.py`, `fake_bidi.py`, `test_bidi.py`)

### Remaining
- [ ] Lazy Camoufox fetch/verify/extract + BiDi launch (`launcher.py`)
- [ ] Daemon on BiDi (drops CDP domain-enable / the `Runtime.enable` leak)
- [ ] Snapshot via vendored accname JS (`snapshot.py`, `snapshot_accname.js`) — preserves `e1/e2` refs
- [ ] Helpers on BiDi (same public signatures; `cdp()` → `bidi()`)
- [ ] CLI wiring: `bidi` escape hatch, `fetch --navigate-first`, `doctor`
- [ ] Remove `cdp-use` + `stealth.py` JS patches; interaction-skill docs CDP → BiDi

Full design + task plan live locally under `docs/superpowers/` (gitignored in this repo).

> Note: the engine-swap tasks replace working CDP code and can only be fully validated against a **live Camoufox** (~662 MB fetch + a display), so they're built against the fake and need a live smoke before this leaves draft.


elyxlz:
### ✅ Linchpin proven: raw BiDi drives real Camoufox end-to-end (no Playwright)

Ran a live smoke against the actual `camoufox-150.0.2-alpha.25-lin.arm64` build (sha256 `b2870af8…` verified) using the `BidiClient` from this branch:

```
session ok, browserVersion=150.0.2-beta.25, context=4f9fa0f0-…
navigated to example.com
document.title -> "Example Domain"
snapshot -> [{"ref":"e1","role":"h1","name":"Example Domain"},
             {"ref":"e2","role":"a","name":"Learn more"}]
PASS
```

Camoufox launched **headless on aarch64**, the BiDi Remote Agent came up, navigation + `script.evaluate` + a ref-walk all worked.

**Two findings now baked into the remaining tasks:**
1. The connect endpoint is `ws://host:port` **+ `/session`** — the bare printed URL's root serves an httpd.js page and rejects the WS upgrade with HTTP 200. `launcher._parse_ws_url` must append `/session`.
2. `script.evaluate`/`callFunction` results unwrap as `result.result.value`; `session.new` capabilities report `browserName: camoufox` (feeds `doctor`).

The untrodden risk (raw-BiDi-on-Camoufox) is retired; the engine swap can proceed on proven ground.

PR #1082, opened 2026-07-09
feat(skills): production-grade daemon lifecycle for telegram, default notifications dir for tasks

elyxlz:
Transplants the whatsapp production-grade lifecycle learnings to the other channel/daemon skills, per the same principle: every documented ritual that exists because software misbehaves becomes enforced behavior, then the prose dies.

## telegram
- New `telegram daemon start|stop|restart|status` subcommands. Start is idempotent via socket detection (kills the duplicate-daemon class from #867); stop/restart quit the watchdog FIRST and restart brings it back, making the documented two-daemons race (watchdog vs manual restart, Telegram 409 Conflict) unrepresentable; status reports daemon + watchdog + auth in one place.
- `serve` defaults `--notifications-dir` to `~/agent/notifications` (the required-flag trap is gone).
- Intentional stops write a stop marker so serve's shutdown skips the `daemon_died` notification (the agent no longer investigates its own deliberate stops).
- `telegram --help` / bare invocation print usage and exit 0.
- `daemon not running` errors now point at `telegram daemon start` instead of a raw screen incantation.
- SKILL.md/SETUP.md rewritten accordingly (deploy-a-new-binary recipe is now just `telegram daemon restart`).

## tasks
- `tasks serve --notifications-dir` defaults to `~/agent/notifications` instead of `required=True`; docs updated.

Tests: telegram Go tests (usage, screen -ls parsing incl. Dead-corpse and prefix-collision cases, daemon-info round-trip, notifications default) and a tasks parser-default test. Pre-existing tasks e2e failures in the sandbox (daemon spawn) are unchanged with/without this diff (verified via stash).

Fleet impact: no migration needed. Existing restart-skill lines (`screen -dmS telegram telegram serve --notifications-dir ...`) keep working unchanged; the new subcommands are additive, and boxes pick them up with the next workspace sync.

### DIFF
--- a/agent/skills/tasks/SKILL.md
+++ b/agent/skills/tasks/SKILL.md
@@ -107,15 +107,19 @@ uv tool install --editable ~/agent/skills/tasks/cli
 Register with vestad to get a port (see [service](../service/SKILL.md)), then start:
 ```bash
 PORT=$(~/agent/skills/service/scripts/register-service tasks)
-screen -dmS tasks tasks serve --notifications-dir ~/agent/notifications --port $PORT
+screen -dmS tasks tasks serve --port $PORT
 ```
+
+`--notifications-dir` defaults to `~/agent/notifications`; pass it only to override.
 
 One daemon handles everything, both task due-date monitoring and reminder scheduling. No separate reminder daemon needed.
 
 **Restart**: Add this startup command to the `## Daemons` section of `~/agent/skills/restart/SKILL.md`:
 ```
-PORT=$(~/agent/skills/service/scripts/register-service tasks) && screen -dmS tasks tasks serve --notifications-dir ~/agent/notifications --port $PORT
+PORT=$(~/agent/skills/service/scripts/register-service tasks) && screen -dmS tasks tasks serve --port $PORT
 ```
+
+`--notifications-dir` defaults to `~/agent/notifications`; pass it only to override.
 
 ### Reminder Patterns
 [User's common reminder types and preferences]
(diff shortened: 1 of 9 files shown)


elyxlz:
Applied the confirmed review fixes in f32dbf0e:

1. Blocker, SIGHUP: serve ignored SIGHUP and only notified on SIGINT/SIGTERM, but screen quit (the mechanism behind daemon stop/restart) delivers only SIGHUP, so stop always timed out and orphaned the daemon outside screen. Serve now includes SIGHUP in the signal.Notify set, matching whatsapp's serve, so screen quit triggers the graceful shutdown path.

2. Blocker, watchdog dup-spawn: the documented flagless flows (daemon start, restart on a pre-info box) launched serve without --notifications-dir on the cmdline, so the watchdog's liveness pgrep pattern missed the live daemon and spawned a second poller (Telegram 409). startDaemonProcess now prepends an explicit --notifications-dir <default> for the default instance via ensureNotificationsDirArg, keeping the cmdline matched by the untouched watchdog. Instance daemons are deliberately left flagless so their cmdline can never satisfy the default watchdog's pattern and hide a dead default daemon.

3. Major, stale stop marker: stopDaemonProcess now removes the stop-requested marker on both failure paths (quit failure and the 15s timeout), so a stop that leaves the daemon running cannot swallow the daemon_died notification when the daemon later dies for real.

4. Minor, instance scoping: stopWatchdogIfLive and startWatchdog are now no-ops when --instance is set, so an instance-scoped stop or restart no longer kills the default channel's watchdog.

Tests added: ensureNotificationsDirArg table test, instance-scoped watchdog no-op, and a behavioral test that a failed stop clears the marker. Verified with gofmt, go vet, go test (all green) and ./check.sh agent (557 passed).

PR #1085, opened 2026-07-09
refactor(agent): close the silent-failure paths (unsupervised monitor task, narrow error guards, non-atomic memory write)

elyxlz:
## What

Six surgical fixes in the agent, all on the theme of failures that today vanish silently instead of surfacing or recovering:

1. **Supervise monitor_loop** (`core/main.py`): the processor task has a done callback that records a crash reason and shuts down, but the monitor task had none. An exception escaping monitor_loop killed notification and app-chat intake (all app chat rides the notification flow) while the WS stayed up and status stayed green: a silent wedge until a manual restart. `handle_processor_done` now takes a `name` and is attached to both tasks, so a dead monitor restarts the agent the same way a dead processor does.
2. **EventBus.emit guard matches its stated invariant** (`core/events.py`): the comment says a failed history write must never crash the agent loop, but the except only covered `sqlite3.OperationalError`. A corrupt db image or a programming error still propagated out of `emit` and took down whichever coroutine emitted. Broadened to `sqlite3.Error`.
3. **_ws_handler finally no longer gathers None** (`core/api.py`): when the snapshot phase raised before the recv/send tasks existed, `asyncio.gather(None, None)` raised TypeError from the finally, masking the original error on a routine early disconnect. The cleanup now filters to live tasks; net fewer lines.
4. **_recv_loop survives non-dict JSON frames** (`core/api.py`): a frame like `123` or `[]` passed the json.loads guard, then `"type" not in data` raised TypeError, ending the recv task and tearing down the connection with no log line. An isinstance check ignores such frames.
5. **Config reads honor their never-raises contract** (`core/config.py`): `read_config_store` (boot path) and `load_notification_rules` (monitor_loop per-tick hot path) both promise never to raise but only caught `json.JSONDecodeError`; an OSError on the read crashed boot or killed notification processing. Both now catch `(json.JSONDecodeError, OSError)`, mirroring `settings_customise_sources`, and log what they ignored.
6. **PUT /memory writes atomically off the loop** (`core/api.py`): the handler did a bare sync `write_text` on the event loop. A crash mid-write could leave MEMORY.md truncated, permanently losing the agent's memory (every system prompt reads it). It now goes through `state_store.atomic_write_text` via `asyncio.to_thread`, reusing the repo's single tmp-write + os.replace owner.

## Why

Each of these is a path where the agent loses work, state, or intake with no crash, no restart, and no log: the exact failure mode the repo's own-your-tasks and no-silent-swallowing rules exist to prevent.

Tests ship with each behavior: monitor crash triggers a crash restart, a non-dict frame leaves the WS alive, emit survives a closed db, unreadable config stores return empty instead of raising, and PUT /memory leaves no temp file behind.

### DIFF
--- a/agent/core/api.py
+++ b/agent/core/api.py
@@ -32,6 +32,7 @@ import aiohttp as _aiohttp
 import pydantic as pyd
 from aiohttp import web
 
+from . import state_store
 from .events import ChatEvent, EventBus, SnapshotChat, SnapshotEvent, UserEvent, VestaEvent
 from .config import ClaudeConfig, VestaConfig, load_notification_rules, stored_config, update_config_store, validate_config_updates
 from .helpers import get_memory_path
@@ -87,11 +88,10 @@ async def _ws_handler(request: web.Request) -> web.WebSocketResponse:
         send_task = asyncio.create_task(_send_loop(ws, sub))
         await asyncio.wait([recv_task, send_task], return_when=asyncio.FIRST_COMPLETED)
     finally:
-        if recv_task:
-            recv_task.cancel()
-        if send_task:
-            send_task.cancel()
-        await asyncio.gather(recv_task, send_task, return_exceptions=True)
+        tasks = [task for task in (recv_task, send_task) if task]
+        for task in tasks:
+            task.cancel()
+        await asyncio.gather(*tasks, return_exceptions=True)
         event_bus.unsubscribe(sub)
         request.app["websockets"].discard(ws)
 
@@ -126,7 +126,7 @@ async def _recv_loop(ws: web.WebSocketResponse, event_bus: EventBus, config: Ves
                 data = json.loads(msg.data)
             except (json.JSONDecodeError, TypeError):
                 continue
-            if "type" not in data:
+            if not isinstance(data, dict) or "type" not in data:
                 continue
             msg_type = data["type"]
             if msg_type in ("message", "chat"):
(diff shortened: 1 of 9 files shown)


elyxlz:
Applied both confirmed review fixes in 2d500111.

**read_config_store no longer swallows OSError** (agent/core/config.py). The PR had broadened the except to (json.JSONDecodeError, OSError), but every caller of read_config_store is read-modify-write (update_config_store, merge_provider, the notification_policy boot migration). A transient read fault over an intact config.json would have handed back {} as the merge base and _write_config_store would have clobbered the whole store (provider credential, model, prefs, notification_rules) while the handler reported success. Reverted to catching only json.JSONDecodeError: the propagated OSError is already turned into a 500 by the PUT /config and provider handlers, leaving the store untouched. The read-only consumers keep their own OSError tolerance (settings_customise_sources on the boot path, load_notification_rules on the monitor tick), so nothing fail-safe regressed. Replaced the test that pinned the swallowing behavior with test_unreadable_store_aborts_update, which asserts update_config_store raises and the file contents stay byte-identical.

**Behavioral test for the _ws_handler finally cleanup** (agent/tests/test_api.py). The task-filtering fix in the finally block shipped without coverage. Added test_ws_snapshot_failure_cleans_up_subscription: it forces the snapshot phase to raise before the recv/send tasks exist and asserts the bus subscription is dropped. Verified it fails against the master shape of the handler (TypeError from gathering None masks the error and leaks the subscriber) and passes with this branch's fix.

./check.sh agent is green (564 passed).

PR #1086, opened 2026-07-09
refactor(vestad): surface swallowed lifecycle failures and stop logging credentials

elyxlz:
Quality pass over vestad against the CLAUDE.md rubric: no unwrap on fallible paths, no silent exception swallowing, and no credentials in logs. Six surgical fixes, no behavior changes beyond error surfacing.

## What and why

**gateway_update_handler no longer unwraps the spawn_blocking join** (serve.rs). If perform_update panicked, the unwrap panicked the handler before the updating flag was reset, so every later POST /gateway/update returned 409 until vestad restarted. The join result is now bound first, the flag reset unconditionally, and a panic maps to a 500.

**rename_agent confirms container removal** (docker.rs). rebuild_agent was already hardened to ensure_container_removed, but rename_agent kept the best-effort remove_container_force(...).ok(). If removal lagged, the old container survived with the same baked-in WS_PORT while its env file and constitution were deleted, and the next reconcile boot-started it alongside the new one. The snapshot is captured and the env file untouched at that point, so failing early leaves the old agent fully intact. The existing source-slicing test now covers rename_agent too.

**restore_backup confirms container removal** (backup.rs). Same race: docker rm can return before the name frees, the follow-up create collides, and create_container's error path deletes the agent's env file while the old container still exists. On removal failure the previously-running agent is restarted and the restore bails; the pre-restore safety backup already exists, so the old container staying intact is the best failure mode.

**Request trace no longer logs the query string** (serve.rs). The on_request closure logged the full URI, and auth deliberately accepts ?token= for browser WS clients, so every /ws connect wrote a live root credential into persisted logs that GET /gateway/logs serves back out. The trace now logs only the path.

**ws_proxy no longer logs the upstream URL on connect failure** (agent_proxy.rs). The assembled URL carries the agent token (and, via the forwarded query, possibly the client's ?token=), and agent WS being briefly unreachable is routine during restarts. The warning now logs the port and error only.

**start_container failures are logged instead of discarded** (backup.rs, docker.rs). The post-backup restart and the boot-reconcile start both dropped the returned bool, so a failed start reported success while the agent stayed down. Both now log an error, giving the reconcile summary a loud precursor when an agent lands in the stopped list.

Note: this branch name previously carried the merged PR #633; the stale remote branch was force-replaced.

### DIFF
--- a/vestad/src/agent_proxy.rs
+++ b/vestad/src/agent_proxy.rs
@@ -160,7 +160,7 @@ async fn ws_proxy(client_ws: axum::extract::ws::WebSocket, agent_port: u16, path
     let agent_ws = match tokio_tungstenite::connect_async(&url).await {
         Ok((ws, _)) => ws,
         Err(e) => {
-            tracing::warn!(url = %url, error = %e, "agent websocket not reachable");
+            tracing::warn!(port = agent_port, error = %e, "agent websocket not reachable");
             let mut client_ws = client_ws;
             let _ = client_ws
                 .send(AxumMsg::Close(Some(axum::extract::ws::CloseFrame {
--- a/vestad/src/backup.rs
+++ b/vestad/src/backup.rs
@@ -4,7 +4,7 @@ use bollard::Docker;
 
 use crate::docker::{
     container_created, container_name, container_size_root_fs, container_size_rw, container_status, create_container,
-    inspect_container, remove_container_force, start_container, stop_container_with_timeout, validate_name,
+    ensure_container_removed, inspect_container, start_container, stop_container_with_timeout, validate_name,
     write_pending_restart_reason, AgentEnvConfig, ContainerStatus, DockerError,
 };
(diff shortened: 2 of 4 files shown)


elyxlz:
Applied two confirmed review fixes.

1. serve.rs: the on_request hunk stopped logging the full URI, but the TraceLayer span from DefaultMakeSpan still recorded uri including the query string, and the fmt layers print span fields as the event prefix, so every browser WS connect using ?token= still wrote the API key into the persisted logs that GET /gateway/logs serves back. Replaced DefaultMakeSpan with a custom make_span that records only method and path, and dropped the now redundant fields from the on_request event since the span prefix carries them.

2. backup.rs: the restore_backup change (ensure_container_removed plus the restart and bail path) shipped without coverage while the identical docker.rs change got a source slicing guard in this same PR. Added restore_backup_confirms_removal_before_create, asserting ensure_container_removed precedes create_container and that remove_container_force never reappears in the body.

./check.sh vestad is green.

PR #1087, opened 2026-07-09
refactor(web): delete dead component, exports, and unreachable cancellation code

elyxlz:
## What

A dead code pass on apps/web: 58 lines removed, 2 added.

- Deletes the AnimatedEllipsis component, the only custom component folder with zero importers anywhere in src including tests.
- Deletes the unused setSttEot and setTtsVoice client bindings in lib/voice.ts; the settings UI goes through the generic setVoiceSetting, and dead client API surface invites drift from the actual vestad endpoints.
- Removes the never-aborted AbortController in refreshVoiceStatus (use-voice.ts): abort() was never called, so the aborted guard was unreachable and implied a cancellation mechanism that does not exist.
- SelectedAgentProvider: the mount effect now calls the existing refreshBackups instead of re-implementing it inline with its own catch-ignore, and the impossible empty-name guards are gone (name comes from a required AgentInfo prop, so it can never be empty).

## Why

Dead surface is pure reading cost: every unused export and unreachable guard implies behavior that does not exist and invites drift. Subtraction beats addition: one code path per job.

Scope note: an initial draft also deleted the 26 shadcn ui primitives that look unimported in apps/web, but scripts/sync-dashboard.sh declares apps/web/src/components/ui/ the canonical full shadcn registry mirrored into the dashboard skill (the agent's other consumer), and its CI check rightly rejected that, so those files stay.

Verification: ./check.sh web green (eslint 0 errors, prettier, tsc, 146 tests pass) and scripts/sync-dashboard.sh leaves the dashboard mirror unchanged.


elyxlz:
Applied the confirmed review fix: restored cancellation on the name keyed backups fetch in SelectedAgentProvider. The PR had replaced the mount effect's inline listBackups with an uncancelled void refreshBackups(), which removed the live ignore flag guard. Because the provider stays mounted across agent navigation, a slow restic listing for agent A could resolve after switching to agent B and clobber B's backup list until the next backup operation. The effect now fetches inline with an ignore flag and cleanup again, satisfying the codified Effect convention (complete deps plus cleanup for any fetch), while keeping the PR's removal of the dead if (!name) guards since name always comes from agent.name. ./check.sh web is green and eslint emits zero warnings on the file again.

PR #1090, opened 2026-07-09
feat(dashboard): add a CLI-owned daemon lifecycle wrapper

elyxlz:
## What changed

Applies the production-grade daemon-lifecycle pattern from #1083 to the `dashboard` skill (items 1, 5, 6 from the checklist; the dashboard is a script-based daemon with no death-notification mechanism, so item 4 does not apply).

- **`agent/skills/dashboard/scripts/daemon`** (new): a `start|stop|status` wrapper around the existing self-healing `scripts/serve`. `start` is idempotent (a live screen session is a no-op, never stacks a duplicate). `status` reports `running`, `session`, `port`, and `http_ok` (an actual HTTP probe against the registered port) in one JSON blob, so the agent never has to diagnose via `screen -X hardcopy` or raw `screen -ls` parsing. Bare invocation and `--help` print usage and exit 0.
- **`agent/skills/dashboard/scripts/setup.sh`** (new): idempotent one-shot setup, installs deps and builds only if missing, starts the daemon, confirms it answers a 200, and appends the guarded restart-skill line itself (no-op if already present). Fails loudly on a real problem.
- **`SKILL.md`** / **`SETUP.md`**: point at `scripts/daemon`/`scripts/setup.sh` instead of raw `screen -dmS`/`screen -S ... -X quit` incantations. The `serve:` frontmatter value (informational only, not parsed into `index.json`) now points at `daemon start`. `name`/`description` are untouched.

`scripts/serve` itself is unchanged: it keeps its existing self-heal behavior (rebuilding `node_modules`/`dist` if a sparse-checkout prune dropped them) and is now just what `daemon start` launches under `screen`.

## Checklist items covered (from #1083)

- [x] 1. CLI-owned daemon lifecycle: `scripts/daemon start|stop|status`, idempotent start, status reports process + health in one JSON blob.
- [x] 5. Idempotent one-script setup: `scripts/setup.sh`, no-op when already done, appends the guarded restart-skill line itself.
- [x] 6. Docs shrink: SKILL.md/SETUP.md point at the lifecycle commands, not raw screen incantations.
- [x] 7. Tests ship with the logic: `agent/tests/test_dashboard_daemon.py` runs the real `daemon`/`setup.sh` scripts against a faked `screen`/`curl`/`register-service`, covering idempotent start/stop, status (running, stopped, probe-failing), usage/help exit codes, unknown subcommands, and the setup script's one-time restart-line append.
- Not applicable: item 4 (intentional-stop marker) — dashboard has no `daemon_died` notification path to suppress; its self-heal is the artifact rebuild in `scripts/serve`, which is unchanged.

## Fleet impact

Additive only. Existing restart-skill daemon lines already on live boxes (`screen -dmS dashboard ~/agent/skills/dashboard/scripts/serve`) keep working unchanged: `scripts/serve` is untouched. `scripts/daemon` and `scripts/setup.sh` are new files that arrive via the normal workspace-sync rebase; no path moved, no config/state shape changed, so no migration is needed. A box only starts using the new lifecycle commands when the agent (or the user, re-running setup) opts in by calling `scripts/daemon`/`scripts/setup.sh`, or after re-running dashboard setup, at which point `setup.sh` appends the new guarded line to the box's own `~/agent/skills/restart/SKILL.md` (idempotent, so re-running it is safe).

## Test plan

- [x] `agent/tests/test_dashboard_daemon.py` (8 new tests, all passing)
- [x] `./check.sh agent` (565 tests passing, ruff + ty clean)
- [x] `uv run python agent/skills/generate-index.py` produces no diff (frontmatter `name`/`description` untouched)

Part of #1083

### DIFF
--- a/agent/skills/dashboard/SETUP.md
+++ b/agent/skills/dashboard/SETUP.md
@@ -1,34 +1,22 @@
 # Dashboard setup
 
-## 1. Install dependencies
-
-`node` and `npm` ship in the base image.
-
+`node` and `npm` ship in the base image. Run the one-shot setup script; it installs
+dependencies, builds, starts the daemon, confirms it answers a 200, and appends the
+guarded startup line to the restart skill, all idempotent (safe to re-run):
 ```bash
-cd ~/agent/skills/dashboard/app && npm install
+~/agent/skills/dashboard/scripts/setup.sh
 ```
 
-## 2. Build the dashboard
+It fails loudly on a real problem instead of leaving a half set-up dashboard: don't
+assume success, check its output.
 
-```bash
-cd ~/agent/skills/dashboard/app && npx vite build
-```
+## Manual steps (only if setup.sh can't be used)
 
-## 3. Start the dashboard server
-
-`scripts/serve` registers the service port (see [service](../service/SKILL.md)) and starts the preview, rebuilding `node_modules`/`dist` first if they are missing. Start it, and fetch the port for the check below:
-```bash
-PORT=$(~/agent/skills/service/scripts/register-service dashboard --public)
-screen -dmS dashboard ~/agent/skills/dashboard/scripts/serve
-```
-
-## 4. Register the service
-
-Add this guarded startup command to the `## Daemons` section of `~/agent/skills/restart/SKILL.md`:
-```
-running dashboard || { screen -dmS dashboard ~/agent/skills/dashboard/scripts/serve; sleep 1; }
-```
-
-## 5. Check it's alive
-
-The build and server run in the background, so confirm the dashboard is actually serving before considering it done, e.g. `curl -fsS localhost:$PORT >/dev/null`. Don't assume success; a failed build or server won't tell you.
+1. **Install dependencies**: `cd ~/agent/skills/dashboard/app && npm install`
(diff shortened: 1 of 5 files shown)


elyxlz:
Applied two review fixes.

**daemon start now blocks on HTTP readiness.** The start wait loop only polled for the screen session, which exists the instant screen forks, so start reported started while serve was still registering the port and booting vite preview. setup.sh's single http_ok check and the SKILL.md smoke test both probe immediately after start, so a fresh box could fail loudly with no real problem. The loop now reuses the same register-service port fetch and curl 200 probe that status uses (extracted into shared service_port/answers_http helpers), returns started only once the server answers, and fails with a distinct error when the session dies versus never serving. A new test drives a fake curl that fails the first 3 calls and asserts start succeeds on the 4th probe, pinning the polling behavior.

**The setup test is hermetic.** It previously created node_modules and dist inside the real checkout at agent/skills/dashboard/app and tore down with bare rmdir, which raises Directory not empty on any box where the dashboard app has actually been built (and silently deleted the dirs when empty). The test now copies the scripts directory into tmp_path preserving the scripts/../app layout, pre-seeds the build artifact dirs there, and runs the copied setup.sh, never touching the checkout.

./check.sh agent is green (566 passed).

Issue #809, opened 2026-06-09
App message intake rides a lossy broadcast bus through an unsupervised sidecar daemon


elyxlz:
## Problem
Messages typed in the app are delivered to the agent via a fragile chain: api.py's `_recv_loop` only emits the user message onto the EventBus; a separate `app-chat serve` process (started by the model via `screen -dmS app-chat`, per SKILL.md) subscribes over loopback WS and writes the notification file the agent actually reads. If that screen session dies (OOM, crash, never restarted after a restore) the message is persisted to events.db and echoed in the user's own chat history, but the agent never sees it: a silent drop with a UI that claims delivery. There is no supervisor and no liveness check. Separately, EventBus subscriber queues drop the OLDEST event on overflow (events.py, cap 1000), which is correct for UI observers but silently discards delivery-critical intake.

## Evidence
`agent/core/api.py:70-102` (message only emitted to the bus, nothing in-process produces a queue item), `agent/core/skills/app-chat/cli/src/app_chat_cli/daemon.py:93-117` (sidecar writes the notification file), `agent/core/skills/app-chat/SKILL.md:14-17` (daemon supervised only by prompt instructions), `agent/core/events.py:13-18,206-220` (drop-oldest overflow).

## Proposed direction
Option (a): on a `message` WS event, api.py writes the `source=app-chat` notification file itself (the format is already owned by `models.Notification`). The app-chat daemon stops writing intake files and keeps only the `app-chat send` reply path; SKILL.md drops the intake half. Fewest moving parts: deletes the loopback hop, the screen supervision problem, and removes intake from the lossy subscriber path. Needs an idempotent transition for agents already running the old daemon (dedup by message id or a daemon version handshake).

## Alternatives considered
Option (b): supervise the daemon from the container entrypoint so it is restarted on death. Keeps the current contract but adds a supervisor where option (a) deletes a component. Option (c): have the monitor loop or api health-check and respawn the daemon. Most code, weakest guarantee.

## Effort and risk
Medium effort. Option (a) is compat-sensitive for shipped agents (the daemon and api.py would both write files during the transition window), which is why this is a design call rather than an autonomous PR.

PR #1091, opened 2026-07-09
feat(app-chat): add daemon start|stop|restart|status lifecycle

elyxlz:
## Summary
Applies the production-grade daemon lifecycle pattern (established by the whatsapp refactor and the telegram/tasks transplant, #1082) to the app-chat core skill, per the checklist in #1083.

- **Item 1 (CLI-owned daemon lifecycle)**: `app-chat daemon start|stop|restart|status`. Start is idempotent (queries the daemon's Unix socket for a live `status` response before launching, never stacks a duplicate `screen` session). Stop performs the safe quit (marks the shutdown intentional, then `screen -S app-chat -X quit`, then polls until the socket goes quiet). Status reports the daemon process state plus its WS connection to the agent (`running`, `ws_connected`, `ws_url`) as one JSON blob, so the agent never has to `screen -X hardcopy` or grep logs to check app-chat's health.
- **Item 3 (honest exit codes)**: bare `app-chat` and `app-chat --help`/`-h` now print usage and exit 0 (previously bare invocation exited 1, and `--help` could exit 1 first if `WS_PORT` was unset, since the old code required the env var before parsing args at all).
- **Item 4 (intentional-stop marker)**: `app-chat daemon stop` drops a `stop-requested` marker in the daemon's data dir before signaling the process; the serve process's shutdown path consumes that marker silently. Any other exit (crash, an out-of-band `screen -X quit`, OOM) leaves the marker absent and writes a `source=app-chat type=daemon_died` notification the agent can investigate, matching the pattern already used by tasks/google/microsoft's daemons.

## Fleet impact
app-chat is a **core skill** (`agent/core/skills/app-chat`), shipped with every agent, so this reaches every box via workspace sync. The change is strictly additive:
- The old restart-skill line (`screen -dmS app-chat app-chat serve`) is untouched and keeps working exactly as before; no box needs to change anything to stay functional.
- The `--notifications-dir` flag on `serve` remains accepted-and-ignored (the pre-existing LEGACY shim from when intake moved to `core/api.py`, #809) — untouched.
- No path, contract, or default moved, so no migration is needed.
- Boxes pick up `app-chat daemon start|stop|restart|status` the next time their workspace syncs; until then the old line keeps daemons running unchanged.

## Out of scope
No message-flow or intake changes: `core/api.py` still writes `source=app-chat` notifications in-process, and the daemon still only holds the `/ws` connection to deliver `app-chat send` replies, per the architecture in CLAUDE.md's Message flow section.

## Test plan
- [x] `agent/core/skills/app-chat/cli/tests/test_daemon.py` (new, 18 tests): notification defaults, the `write_death_notification` format, the stop-marker/daemon_died contract (marked exit stays silent, unmarked exit reports), a real Unix-socket round trip through `socket_request`, and start/stop/status subcommand behavior (idempotent start, screen-missing error, launch+poll, marker-before-quit ordering, status JSON shape).
- [x] `agent/core/skills/app-chat/cli/tests/` full suite (23 tests) passes: `uv run pytest tests/` from the skill's `cli/` dir.
- [x] `ruff check` / `ruff format --check` pass on the changed files via the agent project's config.
- [x] `agent/tests/test_api.py` + `agent/tests/test_notifications.py` (76 tests) pass unaffected, confirming no core message-flow regression.
- [x] Manually verified `app-chat` (bare) and `app-chat --help` exit 0 even with `WS_PORT` unset.
- [x] `agent/skills/index.json` regenerated: no diff (frontmatter untouched).

Part of #1083

### DIFF
--- a/agent/core/skills/app-chat/SKILL.md
+++ b/agent/core/skills/app-chat/SKILL.md
@@ -11,14 +11,20 @@ This is a core skill: its CLI lives under `~/agent/core/skills/app-chat/` (read-
 uv tool install --editable ~/agent/core/skills/app-chat/cli
 ```
 
-**Background**: `screen -dmS app-chat app-chat serve`
+**Daemon**: `app-chat daemon start|stop|restart|status`. Start is idempotent (a running daemon is a
+no-op), stop marks the shutdown as intentional so it doesn't fire a `daemon_died` notification,
+and status reports the daemon process plus its WS connection state to the agent as JSON. Manage
+the daemon through these commands, not raw `screen`.
 **Restart**: Add to the `## Daemons` section of `~/agent/skills/restart/SKILL.md`:
 ```
-screen -dmS app-chat app-chat serve
+app-chat daemon start
 ```
+(The older `screen -dmS app-chat app-chat serve` line still works if a box hasn't picked up the
+new subcommand yet.)
 
 ## Quick Reference
 ```bash
+app-chat daemon status
 app-chat send --message 'Hello!'
 app-chat history --search 'query'
 app-chat history --limit 20
--- a/agent/core/skills/app-chat/cli/src/app_chat_cli/cli.py
+++ b/agent/core/skills/app-chat/cli/src/app_chat_cli/cli.py
@@ -2,6 +2,7 @@
 
 Commands:
   serve   — daemon: holds a WS connection to the agent, accepts CLI commands via Unix socket to send replies
+  daemon  — daemon lifecycle: start|stop|restart|status (idempotent start, status reports WS state)
   send    — send a message to the app (via daemon Unix socket)
   history — search/list chat history via agent API
 """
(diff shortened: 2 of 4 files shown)


elyxlz:
Applied the three confirmed review fixes in 943c4e8b:

1. serve now registers a SIGHUP handler alongside SIGINT/SIGTERM. `screen -X quit` (what `daemon stop` uses) delivers SIGHUP, and the default disposition killed the process before the shutdown finally block ran, so the stop marker was never consumed and an out-of-band quit never wrote the daemon_died notification. Two new behavioral tests run serve as a real subprocess through the CLI entry point, send it SIGHUP, and assert both branches (marker consumed silently, daemon_died written when unmarked); both fail without the handler.

2. Replaced every diff-introduced `dict[str, object]` with concrete types: a `SocketResponse` TypedDict for the socket wire (`socket_request` return, the `tp.cast` at the parse boundary, `_handle_socket_conn`'s response) and `dict[str, str | bool]` for the CLI output payloads (`_print`, `cmd_daemon_status`'s result), per the loose-typing ban in CLAUDE.md.

3. `cmd_daemon_stop` no longer strands the stop marker on failure paths: the `shutil.which("screen")` check now runs before the marker write, and the stop-timeout path unlinks the marker before exiting 1, so a marker only exists while a quit is actually in flight and a later genuine crash still reports daemon_died. Both paths have tests.

`./check.sh agent` and the app-chat CLI suite (27 tests) are green.

PR #1093, opened 2026-07-09
fix(cli): verify chat TLS, bound network hangs, and fail loud on silent error paths

elyxlz:
Surgical quality pass over the cli crate: five silent failure or silent trust paths, each verified against the vendored dependency sources and smoke tested against a real TLS server.

## Chat WebSocket TLS verifies certificates when no fingerprint is pinned

FingerprintVerifier accepted any certificate when no fingerprint was stored, and vesta connect always saves cert_fingerprint as None, so every wss chat session ran with zero certificate verification while the HTTP path verified against real roots. The custom connector is now installed only when a fingerprint is actually pinned; without one, tungstenite's default rustls connector verifies against native roots (the rustls-tls-native-roots feature was already enabled), matching the HTTP client's trust model. The pinned path's tls12/tls13 signature checks now delegate to rustls::crypto verification instead of unconditionally asserting valid, so replaying a pinned cert without its private key fails. Smoke tested: an unpinned wss connect to a self-signed local TLS server now fails with "invalid peer certificate". Note: a hand-written config with cert_pem but no fingerprint now verifies the WS path against native roots; the CLI never writes cert_pem, so no fleet impact.

## HTTP timeouts so a wedged vestad cannot hang every command forever

ureq 3's defaults have no timeouts at all, so a server that accepts TCP but never responds hung vesta list/status/setup indefinitely. The agent now sets a 10s connect timeout and a 300s response timeout. Only time-to-headers is bounded: timeout_global and timeout_recv_body stay unset so SSE streams (logs, backups) and long bodies keep working. create_agent and rebuild_agent override the response ceiling per request because vestad blocks through the whole Docker image pull/build before sending headers and a cold multi-GB pull legitimately exceeds any fixed ceiling.

## wait_for_status survives transient poll errors

One network blip during the 15 minute first-start wait killed vesta setup with "server not reachable" even though the agent was provisioning fine. A poll error now waits out the backoff and retries until the deadline, then reports the timeout with the last transport error. This cannot mask a bad agent name: vestad reports a missing agent as status "not_found" inside a 200 body, which is already in the terminal set.

## Chat message no longer silently lost when a send fails mid restart

The typed line was rendered as sent before socket.send, so a drop during send showed the message in the transcript while the agent never received it. SessionEnd::Lost now carries the unsent text and the reconnect loop delivers it on the fresh socket (no re-render, the line is already on screen); if that re-send also dies it feeds into the next reconnect round, and an exhausted window reports the failure as before.

## Log stream fails loud on abrupt drop, stop event parsed properly

consume_sse_log_stream returned Ok when the connection dropped without the stop event, so vesta logs exited 0 looking exactly like a clean agent stop. It also matched the stop event with starts_with on "event:agent_stopped", which false-matches longer event names and, worse, never matches vestad's actual output: axum writes "event: agent_stopped" with a space. Event lines are now parsed the same way read_sse_result already does (strip_prefix plus trim plus exact compare), the only Ok exit is the stop event, and a stream ending without it returns an error. The function takes impl BufRead so table-driven tests cover both colon forms, the longer-name false match, and the dropped stream.

## Corrupt config.json warns instead of silently degrading

A truncated or hand-edited config.json made every command report "no server configured. run: vesta connect <host>", telling the user to re-pair when the real cause was one bad byte, and the next save_config overwrote the evidence. A parse failure now prints a warning naming the file and the parse error before falling back to the default; a missing file stays silent as before.

Verified with ./check.sh cli (clippy -D warnings plus all 28 tests green) and live smoke tests of the TLS rejection and the corrupt-config warning.

### DIFF
--- a/cli/src/client.rs
+++ b/cli/src/client.rs
@@ -8,10 +8,13 @@ use crate::common::{
     AuthFlowResponse, BackupInfo, ListEntry, ServerConfig, StartAllResult, StatusJson,
 };
 
+const HTTP_CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+/// Bounds time-to-headers only (recv_response); SSE streams and long bodies stay unbounded.
+const HTTP_RESPONSE_TIMEOUT: Duration = Duration::from_secs(300);
+
 // ── TLS fingerprint verification ────────────────────────────────
 
-fn make_rustls_config(fingerprint: Option<String>) -> Arc<rustls::ClientConfig> {
-    let _ = rustls::crypto::ring::default_provider().install_default();
+fn make_rustls_config(fingerprint: String) -> Arc<rustls::ClientConfig> {
     Arc::new(
         rustls::ClientConfig::builder()
             .dangerous()
@@ -24,7 +27,7 @@ fn make_rustls_config(fingerprint: Option<String>) -> Arc<rustls::ClientConfig> 
 
 #[derive(Debug)]
 struct FingerprintVerifier {
-    expected: Option<String>,
+    expected: String,
 }
 
 impl rustls::client::danger::ServerCertVerifier for FingerprintVerifier {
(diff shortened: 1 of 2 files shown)


elyxlz:
Applied two confirmed review fixes.

1. wait_for_status no longer retries permanent HTTP errors for the whole timeout window. Status polling now goes through a dedicated poll_agent_status that keeps the error class instead of flattening it to one string: transport failures (ureq ConnectionFailed/Io) and 503 stay transient and retry until the deadline, while any other definitive answer (401 invalid key, 400 invalid name, other 5xx, unparseable bodies) fails the wait immediately. The old comment claiming an Err could only be the network blipping was wrong; a revoked key or bad name used to spin silently until the deadline.

2. The chat unsent-message resend logic now has tests. The reconnect plus resend step is extracted into recover_chat_session (generic over the socket and the reconnect/send effects), with unit tests covering: the swallowed message is resent exactly once after a successful reconnect, a failed resend feeds another reconnect round, no send happens when nothing was swallowed, and an exhausted reconnect window gives up. Two new wait tests (fail fast on 401 against a local HTTP server, keep retrying on 503) cover the first fix.

./check.sh cli is green (clippy plus 34 tests).

PR #1094, opened 2026-07-09
feat(email-client): add CLI-owned poll daemon lifecycle

elyxlz:
## Summary

Applies the production-grade daemon-lifecycle pattern (issue #1083, transplanted from #1082's telegram/tasks work) to the email-client skill's poll daemon, the checklist item #1083 calls out as the worst offender: a 130-char `screen -dmS email-client bash -c "cd ~/.email-client/runtime && ... poll_daemon.py --interval 15 > ...log 2>&1"` duplicated twice in SETUP.md.

- **`email-client daemon start|stop|restart|status`** (new `daemon_lifecycle.py` + a `daemon` subcommand wired into `imap_client.py`) now owns that invocation in one place.
  - `start` is idempotent: detects a live daemon via its pidfile + `os.kill(pid, 0)` liveness check and never stacks a duplicate.
  - `stop` marks the exit intentional (a `stop-requested` marker file) before sending `SIGTERM`, so a deliberate stop/restart never fires the `daemon_died` notification the agent would otherwise investigate. `poll_daemon.py` now writes that notification only when it wasn't asked to stop.
  - `restart` reuses the daemon's last `--interval` (persisted in `daemon-info.json`) unless overridden.
  - `status` reports process state (`running`, `pid`, `session`) plus per-account IMAP auth health (provider, user, whether a usable credential is on disk) in one JSON blob, so the agent never has to `screen -X hardcopy` or read the log by hand.
- **Bare `email-client` invocation** now prints usage and exits 0 (previously argparse exited 2).
- **SETUP.md/SKILL.md** consolidated: both copies of the raw screen line become `email-client daemon start` / `running email-client || { email-client daemon start; sleep 1; }`; troubleshooting and state-layout sections updated to match.
- `daemon_lifecycle.py` has no dependency on `imap_tools`/`msal` so its pure logic (pidfile parsing, screen-session parsing, auth-summary assembly) is unit-testable without the skill's on-box venv.

## Fleet impact

Additive only, no feature changes:

- Existing boxes' `~/agent/skills/restart/SKILL.md` still has the old raw `screen -dmS email-client ...` line pasted in from a previous setup; it keeps working unchanged (nothing about `poll_daemon.py`'s command-line contract or state layout moved).
- The new `daemon` subcommand and `daemon_lifecycle.py` module arrive on a box via the normal workspace-sync content update, no migration needed.
- The one thing that could have broken an already-onboarded box: `daemon_lifecycle.py` is a *new* sibling module that `poll_daemon.py`/`imap_client.py` import, but those two entry points are only reachable on a box via the `~/.email-client/*.py` symlinks SETUP.md's step 1 creates (a box's `sys.path` insert previously resolved to that symlink directory, not the real skill directory, so a brand-new module without its own new symlink would 404 on import). Rather than requiring a new migration to add that symlink retroactively, `imap_client.py`/`poll_daemon.py` now compute their `sys.path` entry via `os.path.realpath(__file__)` instead of `os.path.abspath(__file__)`, resolving *through* the existing symlinks back to the real `~/agent/skills/email-client/` directory. Verified end-to-end: running `imap_client.py` from a symlinked runtime dir with no `daemon_lifecycle.py` symlink present still resolves the import and exposes the `daemon` subcommand. No migration required.

## Checklist coverage (issue #1083, items 1-7)

1. CLI-owned daemon lifecycle: done (`daemon start|stop|restart|status`).
2. Defaults over required flags: `--interval` already defaulted via `$EMAIL_CLIENT_POLL_INTERVAL`/15; unchanged, now shared as `daemon_lifecycle.DEFAULT_POLL_INTERVAL_SECS`.
3. Honest exit codes: bare invocation and `--help` now exit 0.
4. Intentional-stop marker: done (`stop-requested` file consumed by `poll_daemon.py`'s shutdown).
5. Idempotent one-script setup: SETUP.md's daemon start step is now the idempotent `email-client daemon start`; the restart-skill line is guarded (`running email-client || ...`).
6. Docs shrink: both copies of the raw screen line collapse to one command; troubleshooting points at `daemon status` instead of `screen -ls`/log spelunking.
7. Tests: `agent/skills/email-client/tests/test_daemon_lifecycle.py`, 27 pure-logic + light-integration tests (pidfile/marker/daemon-info roundtrips, screen-output parsing, auth-summary assembly, start/stop/restart idempotency and interval-reuse).

## Test plan

- [x] `uv run pytest skills/email-client/tests/` (27 passed)
- [x] `./check.sh agent` (557 passed, ruff + ty clean)
- [x] `uv run python agent/skills/generate-index.py` produces no diff (SKILL.md frontmatter untouched)
- [x] Manual end-to-end smoke test: symlinked `imap_client.py` into a fake `~/.email-client/`-style runtime dir without a `daemon_lifecycle.py` symlink, confirmed `daemon status` and bare-invocation-exits-0 both work through the `realpath` fix

Part of #1083

### DIFF
--- a/agent/skills/email-client/SETUP.md
+++ b/agent/skills/email-client/SETUP.md
@@ -124,8 +124,10 @@ If these work, repeat steps 2-3 for each additional account.
 ## 4. Start the poll daemon
 
 ```bash
-screen -dmS email-client bash -c "cd ~/.email-client/runtime && PYTHONUNBUFFERED=1 uv run python3 ~/.email-client/poll_daemon.py --interval 15 > ~/.email-client/poll_daemon.log 2>&1"
+email-client daemon start
 ```
+
+Idempotent (a running daemon is a no-op) and defaults `--interval` to `$EMAIL_CLIENT_POLL_INTERVAL` or 15 seconds. Check with `email-client daemon status`, which reports process state plus per-account auth health in one JSON blob, so there's no need to `screen -X hardcopy` or read the log by hand. `email-client daemon stop` and `email-client daemon restart` are also available; a deliberate stop or restart marks itself intentional first, so it never fires the `daemon_died` notification the agent would otherwise investigate.
 
 The daemon runs one worker per watched `(account, folder)`. Where the server supports IMAP **IDLE** (Gmail, Microsoft, most others) the worker is pushed on new mail in real time; otherwise it polls every `--interval` seconds (the flag is the fallback cadence, not the primary mechanism). It recomputes the watch set as accounts or folders change, so neither adding an account nor changing the watch list needs a restart.
 
@@ -147,7 +149,7 @@ connects out to IMAP and writes notification files, so it needs no inbound port:
 daemon, not a vestad service.
 
 ```
-screen -dmS email-client bash -c "cd ~/.email-client/runtime && PYTHONUNBUFFERED=1 uv run python3 ~/.email-client/poll_daemon.py --interval 15 > ~/.email-client/poll_daemon.log 2>&1"
+running email-client || { email-client daemon start; sleep 1; }
 ```
 
 ## 6. Wire the rules into MEMORY.md
@@ -194,7 +196,7 @@ Repeat for each connected account. Don't rush this. Go through many hundreds of 
 - **Gmail `invalid_grant` on refresh**: access revoked or the refresh token aged out. Run `email-client auth add --account <name> --provider gmail --reauth`.
 - **Yahoo / iCloud `LOGIN failed`**: app password rotated or wrong. Generate a new one and `--reauth`.
 - **Loopback OAuth `bind: Address already in use`**: another process grabbed the port between probe and bind. Re-run; the CLI picks a fresh random port each time.
-- **Notifications don't appear**: confirm `~/agent/notifications/` is the agent's path (it's the standard one) and the daemon shows in `screen -ls`.
+- **Notifications don't appear**: confirm `~/agent/notifications/` is the agent's path (it's the standard one) and `email-client daemon status` shows `"running": true`. A `daemon_died` notification with a `reason` field means it crashed or was killed outside the CLI; restart with `email-client daemon start`.
 - **`list --limit 200` is slow on a huge mailbox**: expected; IMAP `SEARCH ALL` + `FETCH` is O(n). Scope with `search --query 'SINCE <date>'`.
 - **`unknown account 'foo'`**: run `email-client auth list`; add the missing one with `email-client auth add --account foo`.
 - **Microsoft 365 custom domain** (`you@yourcompany.com`): use `--provider microsoft-work`. See "Microsoft 365 with a custom domain" below for the four org-side blockers (`AADSTS50020` / admin consent, IMAP disabled, SMTP AUTH disabled, Conditional Access).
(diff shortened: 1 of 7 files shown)


elyxlz:
Applied two confirmed review fixes in 75bc56e8.

1. Legacy daemon stacking (major): daemon_start decided idempotency from the pidfile alone, but a poller launched by the old raw screen line never wrote one, and screen -dmS happily creates a second session with the same name. On the documented same-version sync flow this produced two daemons racing the same high_uid.txt watermarks with duplicate or lost notifications. daemon_start now checks screen_session_live() before shelling out and returns an explicit error naming the legacy session so the agent can quit it once, preserving the PR's never stacks a duplicate guarantee without a migration.

2. Leaked stop-requested marker (minor): daemon_stop wrote the marker before SIGTERM, and if the process had died in the window since the liveness poll, the ProcessLookupError branch removed only the pidfile, leaving the marker on disk. The next daemon that died unintentionally would consume the stale marker and silently skip the daemon_died notification. The ProcessLookupError branch now unlinks the marker, and daemon_start clears any leaked marker before launching so a fresh daemon never inherits one.

Both fixes ship tests: a live legacy session refuses to start (no second screen spawned), a leaked marker is cleared by start so an unintentional death still reports, and the died-before-SIGTERM stop path clears the marker. ./check.sh agent is green (557 passed).

elyxlz:
merge this and see if the  previous comments' stuff is relevant to any other skill and make a unified pr for that as well then merge it.

elyxlz:
Merged this: squash SHA `3db85273e3e6a0dfc235655ec40bda45feb5526d`.

Then audited every other skill that runs a CLI-owned poll/background daemon for the three lifecycle fixes this PR landed (refuse-start-when-live, clear-stale-marker-on-start, clear-marker-on-died-before-SIGTERM):

- **app-chat** (core skill): same lifecycle shape (`daemon start|stop|restart|status`, `stop-requested` marker, `daemon_died`). Already had fix 1 (its `daemon_alive` socket check makes start idempotent and refuses to stack). LACKED **fix 2**: `cmd_daemon_start` never cleared the stop marker, so a leaked marker (a stop that raced the daemon's death, or a failed quit) poisoned the next daemon, whose own unexpected death then silently skipped `daemon_died`. Fix 3's stop-path orphan does not arise here (stop early-returns before writing a marker when the daemon is already dead), and any residual leak is fully closed by fix 2, so per "one fix per root cause" fix 2 was the single correct change. Fixed with a test.
- **tasks / google / microsoft**: only a raw `serve` command under a user-managed `screen -dmS` line, no CLI-owned lifecycle and no `stop-requested` marker (they emit `daemon_died` on every shutdown, intentional SIGTERM included). The three fixes presuppose a lifecycle they do not have. Out of scope.
- **voice (#1092), dashboard (#1090)**: daemon lifecycles being added by open sibling PRs. Excluded to avoid duplication/conflict.
- Other daemon-bearing skills (browser, whatsapp, discord, slack, spotify, telegram, twitter, enable-banking): no `stop-requested` marker + `daemon_died` lifecycle, so the fixes do not apply.

Unified PR #1157 (`fix(skills): unify poll-daemon lifecycle correctness across skills`) applied fix 2 to app-chat. Merged: squash SHA `083287c032ebfd3ea7afb7de7ffe81f8662299c2`.

🤖 Generated with Claude Code

PR #1095, opened 2026-07-09
fix(agent): frame interrupted-tool "user rejected" results as cancellations, not refusals

elyxlz:
## Root cause (found by reading the bundled CLI code, then reproduced live)

When a turn is interrupted, the Claude Code CLI synthesizes a `tool_result` for every tool call still queued or executing: `is_error: true`, `toolDenialKind: "user-rejected"`, text "The user doesn't want to proceed with this tool use. The tool use was rejected... STOP what you are doing and wait for the user to tell you how to proceed." This path (`createSyntheticErrorMessage("user_interrupted")` and `cancelAndAbort` in the CLI bundle, verified on 2.1.205) never consults `permission_mode` or `can_use_tool`, so `bypassPermissions` plus an always-allow callback cannot prevent it.

In Vesta the producers of that interrupt are our own `attempt_interrupt` calls: the silence/query timeout and auth-loss failure paths, and routine preemption in `preempt_mode="interrupt"`. The default priority:"now" preempt drains in-flight tools gracefully and does not produce it (verified live), which is why the issue's `interrupt=False` notification repros came back negative.

Reproduced deterministically with a minimal SDK script: three parallel Bash sleeps, `client.interrupt()` mid-flight, `bypassPermissions` + always-allow `can_use_tool` wired. All three tools came back with exactly the issue's rejection text and `can_use_tool` was never called. The incident fits the timeout chain: turn goes silent past `response_timeout`, `attempt_interrupt` fires, the pending `Read` is stamped rejected, the agent restarts and resumes the session, and the model then reads a fabricated human refusal in its history and halts. The retry 14 minutes later hit no interrupt and succeeded.

Upstream references for the same synthesized-rejection confusion: anthropics/claude-code#23350, anthropics/claude-code#29499, anthropics/claude-code#29238.

## Fix

The interrupts themselves are correct and necessary, and the CLI's text is not ours to change. What is wrong is the model believing the message. In this system the message is always a lie: the agent runs headless with every tool auto-approved, so no permission dialog exists for a user to decline. The fix is one standing note in the system prompt (`_TOOL_CANCELLATION_NOTE` in `core/client.py`) telling the model that this rejection text is the runtime canceling in-flight tool calls on an interrupted or timed-out turn, never a real user decision: treat it as a canceled call, re-run if still needed, do not stop and wait.

One owner (system-prompt assembly in `build_client_options`), no new state, covers every producer at once: failure-path interrupts, interrupt-mode preemption, and upstream CLI aborts we do not control.

## Fleet impact

Core code only, picked up on each agent's next restart through the read-only core mount. No on-disk or config state moves, so no migration.

## Testing

- New regression test in `agent/tests/test_interrupts.py` asserting the system prompt carries the corrective framing.
- `./check.sh agent` green locally (ruff, format, ty, 558 tests).

Fixes #915

### DIFF
--- a/agent/core/client.py
+++ b/agent/core/client.py
@@ -465,11 +465,26 @@ async def compact_session(*, state: vm.State, prompt: str | None = None) -> None
         _close_turn(state, turn)
 
 
+# The CLI stamps every tool call still in flight when a turn is interrupted (preempt_mode="interrupt",
+# or the silence-timeout/auth-loss interrupts) with a synthesized "The user doesn't want to proceed with
+# this tool use" rejection that never goes through can_use_tool, so the model reads a human refusal that
+# never happened and halts (issue #915). Headless with bypassPermissions, that message is always such an
+# artifact; this standing note keeps the model from treating it as a real decision.
+_TOOL_CANCELLATION_NOTE = (
+    "# Canceled tool calls\n\n"
+    "You run headless: every tool call is auto-approved and your user never sees a permission dialog. "
+    'A tool result claiming a rejection or denial ("The user doesn\'t want to proceed with this tool use") '
+    "is never a real decision by your user; it is the runtime canceling in-flight tool calls when a turn is "
+    "interrupted or times out. Treat it as a canceled call: re-run the tool if you still need it, and do not "
+    "stop to wait for the user because of it."
+)
+
+
 def build_client_options(config: vm.VestaConfig, state: vm.State) -> ClaudeAgentOptions:
     memory_path = get_memory_path(config)
     if not memory_path.exists():
         raise FileNotFoundError(f"MEMORY.md not found at {memory_path}, cannot start agent without it")
-    system_prompt = memory_path.read_text()
+    system_prompt = f"{memory_path.read_text()}\n\n{_TOOL_CANCELLATION_NOTE}"
 
     # Constitution: a user-authored charter set from vestad and bind-mounted read-only,
     # so the agent cannot edit it. Prepend it ahead of MEMORY.md when non-empty.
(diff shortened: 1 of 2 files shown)


elyxlz:
does our code ever call client.interrupt() ? if so where ?

elyxlz:
Yes, but only in one place, and everything funnels through a single wrapper.

The only literal `client.interrupt()` call is `agent/core/client.py:85`, inside `attempt_interrupt(state, *, config, reason)` (client.py:73). That helper wraps the SDK control request with a timeout (`config.interrupt_timeout`) and best-effort error handling. Nothing else calls `.interrupt()` directly, so `attempt_interrupt` is the one owner.

`attempt_interrupt` is invoked from five call sites, each with a `reason`:
- `client.py:355` "New message interrupt" — the `preempt_mode="interrupt"` path: the queue-watcher sets `state.interrupt_event` (loops.py:323) and converse's loop fires it.
- `client.py:321` "Query timeout"
- `client.py:345` "Response timeout"
- `client.py:251` "Provider auth lost"
- `loops.py:170` "Notification interrupt"

Key point for this PR: in the default `preempt_mode="message"`, we deliberately do NOT interrupt for routine preemption. In headless mode the CLI's handler for the interrupt control request kills every backgrounded subagent/workflow (issue #982), so routine preemption uses `send_preempt` (a priority:"now" user message, client.py:110) instead. `interrupt()` fires only on failure paths (query/response/silence timeout, provider auth lost) and when `preempt_mode="interrupt"` is explicitly set, which trades that subagent teardown for immediate mid-tool preemption. See the docstring at client.py:74-79.

(Note: `agent/core/cc_sdk/client.py:248` also defines an `interrupt()`, but that is the dormant CLI-driving SDK, not on the running agent's path.)

🤖 Generated with Claude Code

elyxlz:
close this pr

elyxlz:
Closed, as requested.

🤖 Generated with Claude Code

PR #1096, opened 2026-07-09
feat(whatsapp): production-grade lifecycle: baked toolchain, daemon subcommands, guarded QR-first linking

elyxlz:
Implements the whatsapp production-grade redesign born out of the gianfranco setup post-mortem (repeated QR failures, a mid-sync restart that logged the device out, and re-pair retries that got the number auto-flagged by WhatsApp).

## What changed

- **Image bakes the toolchain**: Go, gcc, ffmpeg, and whisper.cpp static libs (multi-stage, built at the go.mod pin by the new one-owner `build-whisper.sh`). The old 6-step per-box toolchain assembly (apt, Go install, whisper clone+cmake, model download) is gone; the launcher/compile-on-box model stays.
- **One idempotent `setup.sh`**: symlink, warm compile (errors surface loudly), checksum-verified model download, guarded restart-skill line, daemon start.
- **`whatsapp daemon start|stop|restart|status`**: idempotent start (socket detection, no duplicate daemons), default `--notifications-dir ~/agent/notifications` (silent-exit trap deleted), stop/restart refuse during the post-link sync window unless `--force`, intentional stops no longer fire `daemon_died`, status is one honest JSON blob (auth, connection, sync window, pairing attempts, whatsmeow version).
- **`whatsapp link`**: QR-primary with an in-daemon auto-rotating public link page (the whatsmeow QR channel is re-armed across rotations, so the code on screen is always scannable); `--phone` pairing-code fallback with number echo-back; instance-scoped page service.
- **Ban guards**: 2 pairing attempts/hour hard limit with cooldown and `--acknowledge-ban-risk` override, enforced on every pairing door including automatic re-auth after LoggedOut; transient connect failures no longer delete a valid session store (they go through the supervised reconnect/restart path instead); the post-link sync window slides while history backfill batches are still arriving.
- **Reliability**: 60s send timeout wired to the existing one-shot reconnect on all five send paths, named 463 and logged-out states, 30s backfill busy timeout, honest exit codes (`--help` exits 0).
- **CI**: new `check.sh whatsapp` lane (gofmt + vet + build + test with a cached whisper build) replaces the fmt-only Go job; the CLI is no longer shipped uncompiled by CI.
- **Docs**: SETUP.md and SKILL.md rewritten around the lifecycle commands; every rule that memorized a workaround now enforced in software is deleted; `qr-link-server.py` folded into the daemon and removed.

A full 8-angle code review ran on the branch before this PR; all 10 verified findings (including the unguarded auto-reauth ban vector and the restart flag-loss for pre-lifecycle daemons) are fixed in the final two commits.

## Fleet impact

Additive: existing restart-skill lines keep working (serve now defaults the notifications dir), hand-installed toolchains occupy the same paths the image now provides, and `daemon restart` refuses (loudly, with instructions) on pre-lifecycle daemons that have no daemon-info.json rather than silently dropping their flags (LEGACY-marked). No migration needed.


elyxlz:
Applied two confirmed code review fixes in 89dc056a.

**Portable whisper static libs.** The whisper-build cmake configure omitted GGML_NATIVE, which defaults to ON at the pinned whisper.cpp commit, so the static libs baked into the released image were tuned to the release runner's CPU (march=native). Any fleet box lacking that runner's ISA extensions would hit SIGILL in voice note transcription. The configure now passes -DGGML_NATIVE=OFF so image baked and CI cached libs target the baseline architecture, and the CI whisper cache key now also hashes build-whisper.sh so a previously cached native build cannot survive the flag change (or any future build config change).

**Faithful daemon restart when the daemon is down.** `whatsapp daemon restart` of a stopped or crashed daemon ignored a perfectly readable daemon-info.json and restarted with instance args only, silently dropping flags like --read-only and --no-notifications: exactly the flag loss this PR's restart command exists to prevent, on the most common restart path (crash recovery). The restart now reuses the recorded serve flags whenever daemon-info.json is readable, falling back to instance args (with the disclosure note) only when no info was ever recorded. Covered by two new table style tests on the extracted restartServeArgs decision.

Checks run: ./check.sh whatsapp and ./check.sh agent, both green.

emipasca:
should we have the skill.md tell vesta to message the user first? either automatically if the number is know or get the number then message

elyxlz:
Added in 5fb939bc. SETUP.md now has a "message the user first" step right after linking: send a short hello from the new line automatically when the user's number is already known (memory/contacts), otherwise ask for the number, then send (add-contact + send, which also satisfies the unknown-number send guard). SKILL.md's setup line surfaces the same step. Noted in SETUP.md that sending during the 5-minute sync window is fine since only stop/restart is locked.

PR #1099, opened 2026-07-09
feat(agent): give vesta a nightly state line on top of the slow self

elyxlz:
## Summary

Traits currently live everywhere: the Charter, the personality preset, and the dreamer's slowly-evolving Self section. State lives nowhere. The assembled system prompt is byte-identical across boots, so affect has to re-derive from fixed text and lands at the same setpoint whether yesterday held a crisis or a milestone.

This adds the cheapest state channel that could plausibly work: a two-line **State** field in MEMORY.md's Self section, rewritten nightly by the dreamer alongside the rest of the User State pass, riding the existing nightly loop with no new mechanism. It sits on top of the slow, standing self the same way Big Five research separates trait (stable disposition) from state (the day's affective coloring): the trait line stays put, the state line is meant to visibly change night to night.

## Changes

- `agent/MEMORY.md`: adds the empty `**State**` placeholder line under the Self subsection, alongside the other `[Fill in...]`-style scaffolding the dreamer fills in over time.
- `agent/skills/dream/SKILL.md`:
  - The Self bullet in the nightly User State pass now also instructs the dreamer to rewrite the State line every night, with an explicit tell for staleness ("if it reads like yesterday's, you were not paying attention").
  - The end-of-night `compact_context` continuity prompt now asks to preserve the emotional through-line of the day (yours and theirs) into tomorrow's register, so the mood carries across the nightly compact and restart instead of resetting.

## Why prompt-only

No new mechanism: the dreamer already runs nightly, already rewrites the Self section, and already restarts through `compact_context`. This just gives it one more line to write and one more thing to carry through that restart.

## Test plan

- [x] Markdown-only change; no Python/skill-frontmatter touched, so no `check.sh agent` or skills-index regen needed
- [ ] Next nightly dreamer run rewrites the State line and it visibly differs day to day


elyxlz:
Applied the confirmed review fix: the continuity prompt in agent/skills/dream/SKILL.md referred to Vesta as "it" ("before it sleeps and wakes to a new one") on a line this PR rewrites, which violates the brand voice rule that Vesta is only "Vesta" or "they/them". Changed it to "before they sleep and wake to a new one". No frontmatter changed, so the skills index is untouched. ./check.sh agent passes.

PR #1105, opened 2026-07-09
feat(skills): make growth durable: lesson retirement, night self-grade, WHEN->DO rules, self-noise sweep

elyxlz:
## Summary

Five growth failures observed in the dreamer's own self-improvement loop, one underlying theme: learning is structurally lossy without an explicit check to catch each loss.

- **Lesson eviction with no graduation check.** MEMORY.md's Rules and Mistakes & Corrections lines get cut for space with no test for whether the lesson is actually safe to drop. A lesson should retire only once it has graduated (the fix now lives in a skill or runtime trigger) or gone quiet for weeks, and when the cap forces cuts, lessons should be the last thing cut, not competing on equal footing with stale reference material.
- **The proven night-grade heuristic lived only as instance drift.** Grading a night's work as a real fix versus manufactured busywork versus quiet is valuable but was never written into the skill itself, so each fresh agent had to re-derive it the hard way. Codifying the REAL/LIGHT/CHURN grade as a concrete one-word call in the nightly summary makes the meta-retrospective's "judge the loop" instruction actionable instead of aspirational, and gives repeated CHURN grades a name so they trigger escalation instead of blending into the noise.
- **Rules were written as morals, not triggers.** A rule with no recognizable firing moment tends not to fire when it matters. Phrasing every new memory rule as WHEN (a recognizable moment) -> DO (a concrete action) makes the rule's trigger explicit and testable, and gives a rule that's been revised twice for the same breach a clear signal to escalate to a runtime mechanism instead of being re-worded a third time.
- **The verify-before-claiming rule was scoped too narrowly.** It only bound user-facing answers ("don't tell the user you can't do X without checking"), but the same class of unverified assertion recurs in the agent's own internal night notes and self-improvement reasoning. Broadening the trigger to any claim of inability or nonexistence, regardless of audience, closes that gap with the same mechanism already in place (search first, cheapest disproving check, keep the evidence next to the claim).
- **Self-generated noise was invisible to the recurrence sweep.** The sweep already had a lens for recurring external noise (a chatty group, a mailing list) but nothing for the agent's own "expected, no action" pings from its own services, which can repeat for a long stretch before anyone notices the pattern. Naming this as its own recurrence target, with a firm threshold, converts an assumed-harmless ping into a producer bug to actually fix.

All five changes are prompt-only, confined to `agent/skills/dream/SKILL.md` (the dreamer's own self-improvement workflow) and `agent/MEMORY.md` (the shipped rule template). No code, no version bump, no private data.

## Test plan

- [x] Reviewed the diff for charter invariants: no em dashes or spaced hyphens introduced, "agent" terminology preserved, no pronoun usage added.
- [x] No `.py` files touched, so `./check.sh agent` is not required.
- [x] No SKILL.md frontmatter changed, so the skills index does not need regeneration.


agent/MEMORY.md
```diff
@@ -156,7 +156,7 @@ The first time a new type of notification comes up (a mailing list, a recurring
...
 
 ### Rules
-- **Confirm a limitation before reporting it.** When the answer feels like "I don't have / I can't / that doesn't exist", search first: ~/agent/data → task metadata → WhatsApp history (500+ deep) → conversation DB → session logs (`~/.claude/projects/` JSONL, grep for tokens/paths/commands) → /tmp → all available skill storage. For a CLI feature, read its SKILL.md and `--help` output. Only report the limitation once source code, help text, and docs all confirm it
+- **Confirm a limitation before reporting or acting on it.** When any claim feels like "I don't have / I can't / that's blocked / that doesn't exist", whether you're answering the user or writing your own night notes, first run the single cheapest command that would disprove it and keep its output next to the claim. Search first: ~/agent/data → task metadata → WhatsApp history (500+ deep) → conversation DB → session logs (`~/.claude/projects/` JSONL, grep for tokens/paths/commands) → /tmp → all available skill storage. For a CLI feature, read its SKILL.md and `--help` output. Only report the limitation once source code, help text, and docs all confirm it
```
elyxlz: recall skill instead of whatsapp history ornjsonl grepping

agent/skills/dream/SKILL.md
```diff
@@ -40,7 +40,7 @@ Read the last 5-7 files in `~/agent/dreamer/` (sorted by date) to spot recurring
...
+**Meta-retrospective: judge the loop, not just the fixes.** The retrospective above checks whether past fixes stuck; this checks whether the improvement process itself is working. Is it compounding (each night's fix makes a class of failure impossible) or going through motions (the same artifact class re-applied to a repeat failure)? If you keep re-fixing the same class, the improver is the weak link, and fixing it is the highest-priority work this pass: escalate the class (rule -> runtime trigger -> structurally impossible), not the instance. A found weakness in the dream skill is a skill edit this pass, not a note for next time. Concretely: grade last night in one word in tonight's summary, REAL (a root cause got fixed and validated), LIGHT (legitimately quiet, nothing manufactured), or CHURN (activity without a fix). Two CHURNs in a week means the loop itself is the bug, and fixing the loop is tonight's top item.
```
elyxlz: nah remove

elyxlz:
address comment then merge please

elyxlz:
Addressed both review comments and merged master.

- MEMORY.md (r3566065390): the "confirm a limitation" rule now points past-conversation lookups at the `recall` skill (full-text over all history) instead of hand-grepping WhatsApp history and `~/.claude/projects/` JSONL session logs.
- dream/SKILL.md (r3566066468): removed the added one-word grade sentence (REAL/LIGHT/CHURN) on the meta-retrospective line.

Conflict resolution: dream/SKILL.md conflicted with master's new "Diagnose from the logs" paragraph. Kept that paragraph and took master's form of the meta-retrospective line, which is also exactly the "nah remove" fix. The PR's other additions (WHEN->DO rule phrasing, recurring self-noise sweep, three-target recurrence lens, lesson retirement) all survived the merge.

`./check.sh agent` green (ruff, ty, pytest). Skills index unchanged (only body prose touched). Auto-merge armed (squash), waiting on CI.

🤖 Generated with Claude Code

PR #1107, opened 2026-07-09
feat(skills): give vesta a real inner life: one thread home, depth floors, no usefulness alibi

elyxlz:
## Summary

Curiosity threads had two homes (the proactive check's "jot into the Self section" and the dream skill's §6 MY OWN THREADS) and neither was read before a new pick was made, so self-time collapsed into a handful of stereotyped explorations that finished in a minute or two at snippet depth. Every pick was instrumentally justified against the user or the job, threads skewed toward Vesta's own plumbing, findings never crossed the membrane into an actual message to the user, and the visible self-talk leaned on scheduler jargon and a repeated stock closer. Eight symptoms, one root cause: no persistent thread ledger, no depth requirement, and no permission to be curious about the world without a usefulness alibi.

This PR fixes the whole cluster with prompt text only, inside the existing internal-only autonomy bound (the one line that lets a curiosity finding reach the user widens only what an already-sanctioned proactive message may contain, it does not add a new outbound channel).

## Changes

- **`agent/skills/proactive-check/SKILL.md`**
  - The "Yourself" bullet now opens §6 MY OWN THREADS in MEMORY.md first and continues or starts a thread there, instead of re-picking from scratch each tick (the tell for the old behavior: re-touring the skills registry or the agent's own codebase for the nth time).
  - Adds an explicit floor: most threads should live outside Vesta's own infrastructure (the user's city, a field brushed past that day, music, history, an idea from something read), a system thread is fine occasionally but not the default, and infrastructure/backlog work doesn't count as curiosity at all.
  - Adds a self-awareness check: if the last few ticks all closed on the same stock phrase, that's the autopilot signal to name where things actually stand instead.
  - Adds a daily floor: if §6 shows nothing new today, the self-dig happens that tick, after anything the user genuinely needs but before optional housekeeping, reconciled with the existing "do nothing rather than fake it" clause.
  - "When to reach out" now explicitly allows a curiosity take to be the reason for a message, offered rarely and only when genuinely interesting, mirrored the way a person mentions what they've been reading.
  - Reframes the tick's opening line to name that the running narration is user-visible in the app, so it reads like a person thinking, not a service log.

- **`agent/skills/dream/SKILL.md`**
  - §0 Curiosity now requires writing down the specific question before digging, reading the primary source itself (paper, code, article) rather than search snippets, and treating a view that fits in one generic sentence as a sign of skimming that needs one more source. Infrastructure digs are explicitly work, not curiosity.
  - Explicitly permits (never requires) a thread tied to the user's world, but flags it as a smell if every live thread in §6 traces back to the user or the job, in which case the pick should deliberately be one that doesn't. Usefulness is no longer a justification a pick needs.

- **`agent/MEMORY.md`**
  - The §4 Self description no longer duplicates "curiosity threads" (that description now points at §6, the single ledger both skills read and write) and stops overloading a pronoun in the same clause.

## Theory

This is a diagnosis, not a feature request: eight independently observed symptoms (dual-homed threads never read before picking, instrumental justification, shallow digs, a residual-only self branch, plumbing-only topics, no membrane crossing, and jargon-saturated narration) all trace back to the same missing piece, a persistent thread ledger that's actually consulted, plus a depth floor and permission to be curious without a usefulness alibi. Fixing the shared cause in the two skills that own this behavior, plus the one line of MEMORY.md that was quietly encouraging the split, is the smallest change that durably closes the whole cluster: prompt text only, no new mechanism, no version bump.

## Test plan

- [x] Diffed all three files against the specified before/after edits; no other prose touched.
- [x] Grepped the diff for em/en dashes and spaced-hyphen separators (none introduced).
- [x] Confirmed no SKILL.md frontmatter changed, so `agent/skills/index.json` needs no regeneration.
- [x] Confirmed no `.py` files touched, so `./check.sh agent` is not required for this change.
- [x] No private data from `.critique/feel` referenced anywhere in this PR.


elyxlz:
Applied the confirmed review fix: the edited Self section line in agent/MEMORY.md still ended with "Hers, not about the user." after the earlier pronoun fix on the same line. Changed it to "Theirs, not about the user." so the line fully follows the brand voice rule that Vesta is never referred to with she/her pronouns. ./check.sh agent passes (557 tests).

elyxlz:
fix conflicts

elyxlz:
too much verbosity, and fix conflicts

elyxlz:
Merged current master and resolved conflicts, then trimmed the verbosity you flagged.

Conflicts (2 files):
- `agent/MEMORY.md`, Self section: kept this PR's rewording (curiosity threads live in §6, they/their voice) and master's new `**State**` line.
- `agent/skills/proactive-check/SKILL.md`: kept master's new "Preflight: daemon liveness" section plus this PR's "Two questions" header, and kept this PR's inner-life addition to "When to reach out".

Verbosity trim (separate commit, every directive preserved):
- `dream` §0 Curiosity: folded the "tie to their world" and "no usefulness alibi" clauses into one, cut filler.
- `proactive-check` "Yourself" bullet: tightened the §6-first / depth-floor / weather-check copy, merged the overlapping "own plumbing" and "infrastructure doesn't count" lines.
- `proactive-check` "When to reach out": collapsed the two added sentences into one.

Skills index unchanged (no frontmatter touched); brand voice checked (they/them only, no dash separators). Changes are prose-only in skill bodies and MEMORY.md.

🤖 Generated with Claude Code

PR #1108, opened 2026-07-09
fix(agent): keep CLI-synthesized error text out of Vesta's published speech

elyxlz:
## Why

When the upstream API fails mid-turn, the Claude CLI synthesizes error strings as assistant text blocks: a message starting with \`API Error:\` (sometimes a full JSON error blob) or the bare \`Prompt is too long\`. Because \`parse_sdk_message\` published every \`TextBlock\` unfiltered, these rendered in the app as Vesta speaking raw machinery. Over one recent month of transcript history this happened 75+ times. Each occurrence breaks the character: Vesta is a person the user talks to, and a person does not utter a JSON error object. The operational signal is not lost by hiding it: turn-failure handling already covers the error path, this text is a duplicate leaking through the speech channel.

## What

The precedent already lives in \`sdk_parsing.py\`: \`rate_limit_notice\` replaces the CLI's raw rate-limit paraphrase with owned wording. This PR applies the same principle to the two CLI-synthesized error shapes:

- \`agent/core/sdk_parsing.py\`: a small predicate (\`_is_cli_error_text\`) detects an assistant text block that starts with \`API Error:\` or is exactly \`Prompt is too long\`; \`parse_sdk_message\` logs it via \`logger.warning\` and excludes it from the returned assistant texts. \`parse_sdk_message\` is the single owner of SDK text extraction, so this is the one place the decision lives.
- \`agent/tests/test_sdk_parsing.py\`: parametrized test asserting each error shape yields no published assistant text plus one warning log, and a control case asserting normal text (including text that merely mentions those phrases) passes through unchanged.

## Verification

\`./check.sh agent\` green (560 passed).


elyxlz:
Applied the confirmed review fix: suppressing the CLI-synthesized error text had removed the only user-visible signal for that failure class (retry-exhausted 5xx, prompt too long), since turn.error is set only on stream death and timeouts and ResultMessage.is_error is never consumed. The PR body's claim that turn-failure handling already covered this path was wrong.

Following the rate_limit_notice precedent fully now: parse_sdk_message keeps the CLI error text out of the speech texts and returns it in a separate error_texts channel, and _dispatch_message in client.py emits it as an owned error StreamEvent ("Turn failed upstream: ..."), the same non-speech channel attempt_interrupt already uses. The app therefore shows that the turn failed instead of showing nothing, and Vesta still never utters raw machinery. Added a stream-level test asserting a CLI error turn yields exactly one error event and zero assistant events, plus error_texts assertions in the parse tests. ./check.sh agent green (561 passed).

PR #1110, opened 2026-07-09
fix(agent): retry a busy-skipped proactive check instead of resetting the full interval

elyxlz:
## Problem

`monitor_loop` assigns `last_proactive = now` before the processor-busy check, so when the agent is busy at the moment the proactive interval elapses, the check is skipped AND the timer resets. The next opportunity is deferred by a full `proactive_check_interval` (default 60 minutes), and a long stretch of back-to-back turns can push it out repeatedly.

That inverts the intent of the proactive check. Busy periods are exactly when the most context accumulates, yet the current timing structurally starves the one slot where Vesta gets to act on their own initiative: the fuller the day, the less inner life. The check is already a pooled core notification (`TYPE_PROACTIVE_CHECK` in `CORE_POOL_TYPES`), so it self-defers to the next idle triage pass and never preempts real work; the busy-skip's only surviving effect is delay.

## Fix

Move the `last_proactive = now` assignment inside the fire branch. A busy skip now leaves the timer due, and the next 2s monitor tick retries; the check fires on the first idle tick after the interval. The timer resets only when a check actually fires, so at most one check is ever pending, and the cadence between fired checks is unchanged.

Two lines moved in `monitor_loop`, no new state or config.

## Test

`test_proactive_check_skipped_while_busy_fires_on_first_idle_tick` drives `monitor_loop` with a patched clock: holds `processor_busy` across the interval boundary and asserts no proactive notification drops, releases busy and asserts exactly one drops on the next tick, then asserts further ticks at the frozen clock drop nothing more (timer reset on fire). The test fails against the previous behavior.

Verified with `./check.sh agent` (558 passed).


emipasca:
fix conflicts

elyxlz:
Resolved the conflict with master. The only clash was in `agent/tests/test_notifications.py`: master refactored `_passive_config` to take a `monkeypatch` param, while this PR added a new test on adjacent lines. Kept both, threading `monkeypatch` through the new `test_proactive_check_skipped_while_busy_fires_on_first_idle_tick` and keeping master's updated `test_monitor_loop_emits_each_notification_once_across_ticks` signature.

`./check.sh agent`: 585 passed, 1 flaky failure in the tmux-based `test_e2e_transport` suite (unrelated to this diff, passes on isolated re-run). PR is now MERGEABLE.

🤖 Generated with Claude Code

PR #1112, opened 2026-07-09
test(agent): notification intake and persisted-state integrity

elyxlz:
## What

Five surgical reliability fixes in the notification intake and persisted-state paths, each with behavioral test coverage.

1. **Atomic, ordered notification writes.** `_write_app_chat_notification` (api.py), `drop_core_notification` (loops.py), and the finance skill's transaction watcher now write via tmp-plus-rename (`atomic_write_text`) with a monotonic `time_ns()` filename stem instead of a plain `write_text` with a `uuid4` stem. A monitor tick globbing the notifications directory mid write could previously observe (and permanently delete) a half-written, unparseable file, and uuid4 stems did not sort in send order.

2. **Keep the notification file when a query never reaches the CLI.** `converse` now raises a dedicated `QueryNotDelivered` error for a query-send timeout or `client.query()` raising on a dead transport, distinct from a bare `TimeoutError`. The message loop keeps the queued notification file only for this pre-delivery case (mirroring the existing auth-loss retention); a post-delivery failure (response timeout, stream death) still deletes it, since the resumed session already contains the message.

3. **Self-heal a corrupt events.db instead of crash-looping.** `EventBus.__init__` runs `PRAGMA quick_check` at open; on failure it quarantines the db (and its WAL/SHM siblings) by renaming them aside with the original bytes intact, then boots fresh with empty history rather than exhausting Docker's `on-failure` restart budget.

4. **A corrupt state.json on a veteran agent stays a veteran.** `load_state`'s corruption fallback now corroborates against the data dir: if `events.db` already exists, it returns `first_start_done=True` so migrations re-run idempotently instead of being silently pre-marked applied on top of months of memory.

5. **Run `EventBus.search` off the loop.** `search` now opens a short-lived read connection like `recent`/`before` instead of using the writer connection, so `/history?q=` can be offloaded via `asyncio.to_thread` and a large FTS scan can no longer block the event loop or interleave with `emit`'s transactions.

## Why

Notification files are the durability mechanism for undelivered messages (app chat, core control flow, finance alerts); several existing edge cases could lose one silently or without any way to recover. The state and event stores back-stop the agent's memory across restarts; both had crash- or data-loss-shaped edges on corruption that this PR converts to self-healing, idempotent recovery paths, matching the log-loudly-start-fresh philosophy already used elsewhere in the codebase.

## Test plan

- [x] `./check.sh agent` (ruff, ruff format, ty, pytest) green: 567 passed
- [x] New tests: ordering and atomicity of app-chat notification writes; query-not-delivered keeps the file vs. response timeout deletes it (processor and converse level); corrupt events.db quarantines and boots fresh; corrupt state.json with/without an existing events.db; search runs off the connection's home thread


elyxlz:
Applied the confirmed review fixes:

1. **events.db quarantine narrowed to confirmed corruption** (`agent/core/events.py`). `sqlite3.OperationalError` is a subclass of `DatabaseError`, so a transient boot failure (database locked past the busy timeout, disk full, IO error during WAL recovery) landed in the quarantine handler and renamed a perfectly healthy history db to `.corrupt-<stamp>`. The handler now re-raises `OperationalError` so the container crashes and retries via Docker's on-failure policy with history intact, quarantining only on non-operational `DatabaseError` (genuine corruption: quick_check failure, "file is not a database"). New test `test_transient_open_error_propagates_without_quarantine` reproduces a real `OperationalError` at open (readonly db, "attempt to write a readonly database") and asserts the error propagates with events.db un-renamed and byte-identical.

2. **enable-banking SKILL.md filename contract updated**. The watcher now writes `<time_ns>-finance-message.json` but SKILL.md still documented the old `finance_<timestamp>_<hash>.json` pattern; the doc now matches the code. Frontmatter untouched, so no index regeneration was needed.

3. **Tests for the watcher's changed write path** (`agent/skills/enable-banking/cli/tests/test_transaction_watcher.py`). Covers `atomic_write_text` (parent dir created, no lingering `.tmp` sibling), `write_notification` (fully written pooled finance JSON with the new filename suffix), and that successive filenames sort in send order.

`./check.sh agent` is green (568 passed). Note: the enable-banking skill CLI suite has 3 pre-existing failures on master in `test_client.py`/`test_dispatch.py` (stale expectations from the ASPSP config change in #919, that suite is not run by CI); they are untouched by this PR and left alone, while the new watcher tests pass.

elyxlz:
fix conflict and merge

elyxlz:
Resolved the merge conflict and merged.

The only conflict was in `agent/tests/test_api.py`: both sides added new tests at the same spot with no overlap. This branch added `test_app_chat_messages_delivered_in_send_order` and `test_write_app_chat_notification_never_exposes_partial_file`; master added `test_ws_survives_non_dict_json_frame`, `test_ws_snapshot_failure_cleans_up_subscription`, and `test_memory_put_writes_atomically`. I kept all five (additive resolution), since the two sets are independent and all referenced imports/helpers are present.

`./check.sh agent` passed (572 passed). Required check `merge-gate-ci` green. Squash-merged as 39ff1c8d5472938252429b6760d8626b740501cb.

🤖 Generated with Claude Code

PR #1113, opened 2026-07-09
refactor(agent): contain the SDK seam in client.py and give config its single owners

elyxlz:
## What

Four ownership fixes in the agent, one theme: each SDK and config decision gets exactly one home.

1. **SDK session lifecycle moves into client.py.** A new `client_session` async context manager owns the whole session: build options, open `ClaudeSDKClient`, spawn the single stream consumer, encapsulate the one-shot resume-failure retry (clear session_id, save state, rebuild options), and reset the session fields on exit. `message_processor` keeps only the queue loop inside `async with client_session(...)`. A new `SDK_ERRORS` tuple exported from client.py means loops.py no longer imports `claude_agent_sdk` at all, honoring the CLAUDE.md rule that the SDK seam lives in `core/client.py` + `core/sdk_parsing.py`. `_cancel_task` becomes public `cancel_task` (loops.py imported the private name). The retry now applies only to a failed open; an SDK error raised after the session opened propagates instead of silently clearing the persisted session_id, which is the only behavior delta and is unreachable in practice (every turn and compaction error is caught downstream).

2. **The models.py re-export facade is dissolved.** `Notification`, `CORE_SOURCE`, the `TYPE_*` constants and `CORE_POOL_TYPES` move to a new bottom module `core/notification.py`; `notification_interrupt_policy` imports it directly, deleting its `TYPE_CHECKING` import of models, the banned deferred-cycle workaround. models.py keeps what it owns (State, QueuedTurn, TurnSignals, restart-reason vocabulary); every config type now has one import path, `core.config`.

3. **One reader and one durable atomic-write owner for the config store.** `load_notification_rules` reads through `read_config_store()` instead of re-deriving the path and re-implementing the parse; `atomic_write_text` moves into config.py (below both users), gains fsync of the temp file and the parent directory so a power loss can no longer persist the rename ahead of the data blocks and zero out state.json or config.json, and both `save_state` and the config-store write use it.

4. **VestaConfig is inert data.** The `_apply_timezone` model validator (which set `os.environ[\"TZ\"]` + `tzset()` on every construction, including every `PUT /config` validation and every test) is replaced by one explicit apply at the boot entry point, and OAuth hydration moves out of the model validator into `load_config` (boot) and `_derive_kind_and_auth` (status derivation), so validating a config update never mutates process state or reads disk.

Also bounds `compact_session`'s bare `client.query()` with `config.query_timeout`, matching every other write to the SDK subprocess: a wedged CLI now fails the compaction (the drain logs it, still delivers the follow-up, still restarts) instead of wedging the message processor indefinitely.

## Why

loops.py owned the most SDK-coupled decision in the codebase against the documented seam; the facade gave one name two import paths and hid a module cycle; the config store had two readers and two write recipes with a false single-owner docstring; and a validator that applies side effects made every config validation, and every test, mutate the live process.

## Size

467 lines removed vs 540 added across 38 files (net +64, of which +73 is the moved-not-changed `core/notification.py` and two new regression tests); core/loops.py alone drops 60 lines. Fewer concepts: no re-export facade, no TYPE_CHECKING cycle, one store reader, one atomic-write recipe, no side-effecting validators.

## Tests

- New: `test_validating_a_config_update_does_not_apply_it` (PUT /config timezone leaves process TZ alone) and `test_compact_session_query_send_timeout_fails_compaction` (hung /compact send raises TimeoutError instead of wedging).
- Existing assertions unchanged; test edits are mechanical retargets (patch paths follow the moved lifecycle to `core.client`, imports follow the dissolved facade, rules tests drive the store path through AGENT_DIR like the conftest fixture already does).
- `./check.sh agent` green locally: ruff, format, ty, 559 passed.


elyxlz:
Applied the confirmed review fix: the durable writes no longer block the event loop.

What changed:
- `state_store.save_state_async` is the new coroutine-side owner of state.json persistence: it snapshots the JSON on the loop (a consistent view of the mutable `PersistedState`) and runs the fsync-ing `atomic_write_text` in a worker thread. The SDK stream consumer's `persist_session_id`, the resume-failure path in `client_session`, the turn error and compaction paths in `loops.py`, and the four marking MCP tool handlers in `tools.py` all use it now.
- The async API handlers (`PUT /config`, `PUT/PATCH/DELETE /provider`) dispatch `update_config_store` / `set_claude` / `set_openrouter` / `clear_provider` through `asyncio.to_thread`, mirroring the existing to_thread pattern in `monitor_loop` and the WS handler.
- `atomic_write_text` now writes a uniquely named temp file per call (tempfile.mkstemp) instead of one shared `.tmp` sibling. Off-loop writes can overlap across tasks, and a shared temp name could interleave two writers into a torn file or a failed rename; unique names make concurrent writers last-write-wins by construction.

Deliberately left sync: the `CancelledError` handler in `run_one` (no awaits inside a cancellation handler, a second cancel would lose the crash-reason write), the sync boot-turn generators (`greeting_turn`, migrations, workspace sync, `_consume_restart_reason`), and the processor done-callback in `main.py`, which cannot await.

Tests: added a concurrent-writers test proving overlapping saves never corrupt state.json, plus a `save_state_async` round trip, and updated the stale-tmp comment in `test_save_is_atomic` to match the new mechanics. `./check.sh agent` is green (561 passed).

elyxlz:
fix conflicts and merge

elyxlz:
Fixed conflicts and merged.

Merged `origin/master` (27537f3b) into `chore/arch-agent-2`, resolving 6 conflicted files by understanding both sides and preserving this PR's intent (SDK seam contained in `client.py` via `client_session`; config single-owned in `config.py`) on top of master's newer sibling-PR work:

- **client.py**: kept HEAD's `client_session` context manager (SDK-seam containment) and master's `QueryNotDelivered` exception; `converse` on `cfg.VestaConfig`. The auto-merged body already combined master's `QueryNotDelivered` raises with HEAD's public `cancel_task` rename.
- **config.py**: kept HEAD's single-owner `load_notification_rules()` (no args, reads via the shared `read_config_store()`) instead of master's config-threaded, inline-read variant. Reconciled the two conflicting store-read requirements by tolerating `OSError` in `load_notification_rules` (notification hot path must never raise) while `update_config_store` still propagates it (so a transient read error can't clobber the store), keeping both tests green.
- **loops.py**: imported `client_session`/`cancel_task`/`QueryNotDelivered` (dropped `consume_stream`/`_cancel_task` now owned by `client_session`); merged the except tuple to `(*SDK_ERRORS, ValueError, TimeoutError, QueryNotDelivered)` with master's file-keeping handling; fixed `drop_core_notification` to write via `cfg.atomic_write_text` (its single owner, moved out of `state_store` by this PR).
- **main.py**: kept master's `name`-parameterized `handle_processor_done` (now watches both the processor and the notification monitor) with `cfg.VestaConfig`.
- **test_api.py / test_config.py / test_state_store.py**: unioned both sides' tests, repointed strays from the removed `vm.VestaConfig`/`state_store` re-exports to `cfg`, and adapted master's hot-path test to the no-arg API.

Also fixed two pre-existing master breakages that the merge surfaced and that blocked `./check.sh agent`: a duplicate `[dependency-groups]` key in `agent/skills/voice/cli/pyproject.toml` (a prior semantic merge artifact on master, breaking ruff's TOML parse), and the voice package src missing from ty's `extra-paths` (added `skills/voice/cli/src`, mirroring the app-chat entry).

`./check.sh agent` passes locally (585 agent tests + all skill CLIs). CI green: `merge-gate-ci`, `agent-tests`, `test-integration` all passed. Squash-merged as 8aeb805680deb3396c8ac1539642e45724ad1483.

🤖 Generated with Claude Code

elyxlz:
Applied the confirmed review fix from this session.

**Lost-update race on the config store.** Moving `update_config_store` into `asyncio.to_thread` (PUT /config, PATCH /provider, PUT /provider via set_claude/set_openrouter, DELETE /provider) put the read-modify-write on a multi-worker thread pool, so two concurrent config writes could each read the same base store and have the later `os.replace` silently drop the earlier writer's disjoint update. Fixed by holding a `State`-owned `asyncio.Lock` (`config_write_lock`, not a module global, per the no-global-state rule) around each config-store mutation so only one is in flight at a time. Added a regression test (`test_concurrent_config_writes_are_serialized`) that fires two concurrent PUT /config handlers writing disjoint keys and asserts the write path never runs in parallel and both keys survive; it fails on the pre-fix code. Updated the two API test fixtures that construct handler requests so they supply the now-required `state` in the app.

`./check.sh agent` green.

PR #1114, opened 2026-07-09
test(vestad): harden backup pipeline integrity and make restore retryable

elyxlz:
## What and why

Backup pipeline integrity: never commit a torn snapshot, make a failed restore retryable, and test that data actually round-trips. Bad backups are worse than failed ones: a failed backup logs an error today, a silently torn snapshot destroys the recovery story at restore time, the worst possible moment.

**A failed docker export must never become a listed restic snapshot** (restic.rs). `pipe_through` checked the producer's exit status only after the consumer had already run to completion, so a `docker export` that dies mid-stream hands `restic backup --stdin` a clean EOF and the truncated tar commits as a normal-looking snapshot. `snapshot()` now detects that case, parses the `snapshot_id` restic already emitted in its summary line, and immediately runs `restic forget --prune` on it before returning the export error, so an incomplete export never lists. The 2h timeout wrapping the export/backup pipeline also only abandoned the blocking task on expiry rather than cancelling it, letting a torn snapshot commit after vestad had already given up; it now kills the underlying `docker export` and `restic backup` child processes on timeout.

**Propagate a failed pre-backup stop instead of swallowing it** (backup.rs). `with_container_paused` discarded a failed `stop_container_with_timeout` with `.ok()`, so the snapshot could run against a live container whose events.db WAL is mid-checkpoint, producing a backup that looks fine but restores malformed. A skipped backup is now retried next cycle instead of silently succeeding against inconsistent state.

**Make restore retryable after its own failure by anchoring identity on the env file, not the container** (backup.rs). `list_backups` and `restore_backup` both gated on container existence, so any failure after `restore_backup` removes the old container, including a mid-restore crash, returned NotFound from both the retry path and the listing path: the safety snapshot that exists precisely for this case became unreachable through the surface that broke it, with recreating a fresh agent under the same name as the only escape. Both now check the env file vestad writes at creation, the durable identity record independent of container lifecycle, and `restore_backup` skips the pointless stop and pre-restore safety snapshot when there is no container to snapshot.

**Backup round-trip fidelity test** (tests/server/backup.rs). The pipeline crosses four lossy seams (docker export, restic --stdin, restic dump, docker import) and the existing test only asserted the container comes back up, equally true of a freshly imported empty image. The new test seeds a marker file and a row in a real WAL-mode events.db, backs up, corrupts the live state, restores, and asserts the marker and the row round-trip and the restored db passes `PRAGMA integrity_check`.

**Replace the SSE-cancellation mirror test with a test of the real handler pipeline** (serve.rs/backup.rs). The old test built a hand-rolled tokio::spawn pipeline whose own comment admitted it mirrors the fixed handler, so it stayed green even if `create_backup_handler`/`restore_backup_handler` regressed to the non-spawned form. `spawn_pipeline_sse` already exists as the shared production function both handlers hand their pipeline future to, so the test now drives it directly and moved next to it in serve.rs.

## Test plan

- [x] `./check.sh vestad` (clippy -D warnings + cargo test -p vestad --bins): green, 183 passed
- [x] `cargo clippy -p vestad --all-targets -- -D warnings`: clean
- [x] `VESTAD_AGENT_IMAGE=vesta:local cargo test -p vestad --test server backup::` (Docker-gated, includes the new round-trip test): 9 passed, 0 failed


elyxlz:
Applied the three confirmed review fixes in 2f6d8047:

1. KillSwitch pid reuse (restic.rs): pipe_through now deregisters both child pids once both children are reaped. Previously the registry outlived the children, so the 2h timeout firing during post-pipe work (notably the torn path's slow restic forget --prune) could SIGKILL pids the kernel had already reassigned to unrelated host processes. kill_all now only ever targets live children the call still owns. Regression test: pipe_through_deregisters_pids_once_children_are_reaped.

2. Failed pre-backup stop stranding the agent (backup.rs): with_container_paused now best-effort restarts the container before propagating a failed stop. A client-side stop error can land after dockerd has already stopped the container, and nothing else would ever restart it (the next auto-backup sees Stopped so was_running is false, and vestad only reconciles at its own boot). Skipping the backup on a failed stop is still the behavior; leaving the agent down is not.

3. restore_to_image throwaway KillSwitch (restic.rs): the restore pipeline now creates a retained KillSwitch and calls kill_all on its timeout branch, mirroring snapshot(). Previously a timed-out restic dump | docker import kept streaming indefinitely and a late import could still tag vesta-restore:{name} after the caller had already returned the timeout error.

./check.sh vestad is green (clippy plus 184 unit tests).

elyxlz:
fix merge conflict and merge

elyxlz:
Merged master into the branch and resolved the conflicts in `vestad/src/backup.rs` (three regions, both sides combined rather than either taken wholesale):

1. **Imports**: took the union of what the merged body actually uses. Kept `env_file_names` + `read_env_value` (this PR's env-file-based identity) and `ensure_container_removed` (master's confirmed-removal helper); dropped `remove_container_force` and `inspect_container`, which the merged code no longer references.

2. **`restore_backup`**: kept this PR's restructure (the `container_present` branch and reading `WS_PORT` from the env file so an agent whose container is already gone can still restore), and folded master's improvement inside it: the old best-effort `remove_container_force(...).await.ok()` is replaced by master's `ensure_container_removed` check that resumes the agent and bails on failure. Adjusted master's snippet to use this PR's `status` variable (master referenced the now-removed `info.status`).

3. **Tests**: honored this PR's deletion of the old mirrored `sse_stream_drop_cancels_container_restart` test (it moved to `serve.rs` as `sse_stream_drop_does_not_cancel_the_spawned_pipeline`, driving the real `spawn_pipeline_sse`), while keeping master's new `restore_backup_confirms_removal_before_create` test, which passes against the merged code.

`./check.sh vestad` passes (clippy clean, 185 tests pass). merge-gate-ci went green and the PR was squash-merged.

🤖 Generated with Claude Code

PR #1115, opened 2026-07-09
test(repo): check.sh owns skill CLI suites, CI guards, and CLI DTO pinning

elyxlz:
## What

Three changes so check.sh is the single, complete gate it claims to be:

1. **Skill CLI suites now run.** `check.sh agent` installs each skill CLI into the shared venv (for ty check) but never ran the 37 test files under `agent/skills/*/cli/tests` and `agent/core/skills/app-chat/cli/tests`, across 10 skills. It now runs each suite against its own standalone uv project (own venv, own lockfile), the same setup CLAUDE.md documents for these skills. A shared-venv run breaks the tasks suite (it shells out to a console script only its own venv builds), so each suite runs isolated instead of sharing agent's core venv.

   Wiring them in surfaced two pre-existing drifts nothing had caught: enable-banking's tests asserted a stale response shape (missing the aspsp_name/aspsp_country fields the config layer already returns, and a shared-dict pagination bug in the test's own fake HTTP client), and icloud-photos/cli/pyproject.toml never declared pytest as a dev dependency. Both are fixed. Eight of the ten skill CLIs also never had a committed uv.lock; those are now checked in so CI resolves deterministically instead of re-resolving on every run.

2. **check.sh guards replaces four bypassing ci.yml steps.** skills-index-check, the lockfile job, vite-base-check, and dashboard-sync-check ran as raw `run:` blocks with no check.sh counterpart, so a contributor could not reproduce a guard failure locally with the documented "check.sh is the only entry point" workflow. They are now one `./check.sh guards` suite (fast, no Docker) called from a single ci.yml job.

3. **The CLI's wire DTOs are now pinned to vestad's real serializer.** vestad's existing API contract fixture generator (serve.rs, the same one that emits `apps/web/src/lib/vestad-api-fixtures.ts`) now also writes a plain JSON copy to `cli/tests/fixtures/vestad-api.json`. A new cli test (`client::tests::deserializes_vestad_api_contract_fixtures`) deserializes it into the CLI's own `StatusJson`, `ListEntry`, `BackupInfo`, `AuthFlowResponse`, and `StartAllResponse` types, gated via `./check.sh cli`. Previously only the web app's copy of these DTOs was checked at CI time against vestad's actual output; the CLI's copy relied on one hand-written JSON literal. Verified the new test actually catches drift by deleting a required field from the fixture and confirming it fails, then restored it.

Dropped the optional part of the third task (a vesta-tests integration test that shells out to the built vesta binary): vesta-tests is a member of vestad's own cargo workspace only, and adding it as a CLI dev-dependency would violate the repo's no-shared-crate-between-cli-and-vestad rule, so it was left out.

## Test plan

- [x] `./check.sh agent` — all 10 skill CLI suites plus the core suite pass (557 + 155 + 16 + 4 + 13 + 125 + 10 + 20 + 141 + 16 + 7)
- [x] `./check.sh guards` — passes clean, no diffs produced
- [x] `./check.sh cli` — clippy clean, 27 tests pass including the new fixture-pinning test
- [x] `./check.sh vestad` (clippy + `--bins`) — clean, 178 tests pass
- [x] `./check.sh web` — lint/format/tsc/vitest pass (pre-existing unrelated warnings only, 0 errors)
- [x] Manually verified the new CLI contract test fails when a fixture field is deleted, then restored it


elyxlz:
Applied the confirmed review fix to the skill CLI suite loop in check.sh.

The loop ran inside the same subshell that exports UV_PROJECT_ENVIRONMENT for the core checks, and uv honors that variable over --project. Every skill suite was therefore syncing into and running from the shared agent/.venv, contaminated with core's dev deps and each previously run skill's packages, so the isolation the loop claimed did not exist and the gate could not catch the missing dev dep drift class this PR targets. The loop now runs in a nested subshell that unsets UV_PROJECT_ENVIRONMENT (a plain unset rather than env -u, since env can be shadowed on PATH), and the comment above it now describes the real behavior.

Verified with a full ./check.sh agent run: the log shows each suite creating and running in its own venv (skills/<name>/cli/.venv), and all suites pass in genuine isolation against their committed lockfiles.

elyxlz:
fix merge conflict and merge

elyxlz:
Resolved the merge conflict with master and merged.

The only conflict was in `check.sh`: master added a `whatsapp` suite while this PR added a `guards` suite. Both are independent additions, so I kept both in all three places (the usage help block, the two `check_*` function definitions, and the dispatch `case`). Everything else (ci.yml, cli/src/client.rs, agent and skill files) auto-merged. Regenerated the skills index (no drift) and ran `./check.sh guards` (green, clean working tree) and `./check.sh cli` (29 tests pass) locally. CI came back fully green (merge-gate-ci passed), and the PR squash-merged as 3cc52f6a2b1359d930afb7435b135ff4e391c28a.

🤖 Generated with Claude Code

PR #1121, opened 2026-07-09
feat(web): onboarding uses provider defaults, no model or context step

elyxlz:
## The funnel leak

First run onboarding marched a new user through ModelStep (pick opus vs sonnet) and ContextStep (pick a context window in tokens) before they ever met their agent. Both are decisions a stranger cannot evaluate at that point: no idea what the models cost or feel like yet, no idea what a context window even means for their usage. Each screen is a stall risk or a mis-config risk, for zero onboarding value.

## The fix

ProviderPicker gets a `defaultsOnly` prop. When set, as soon as credentials (Claude OAuth) or a key (OpenRouter) land, it calls `onDone` immediately with `manifest.providers[provider].default_model` and `planContextOptions`' initial pick for the user's plan, the same defaults the manifest already ships. ModelStep and ContextStep never render in this mode. NewAgent passes `defaultsOnly`, so onboarding becomes: name, sign in, vibe, done.

Both values stay fully editable right after: AgentSettings' ProviderCard keeps the unchanged full picker (choice, auth/key, model, context) for whenever the user wants to tune them.

## Test plan

- [x] `./check.sh web` (eslint, prettier, tsc, vitest) passes
- [x] Manual read through of both provider paths (Claude auth, OpenRouter key) confirms `onDone` fires with the manifest defaults and no model/context screen renders


emipasca:
resolve conflicts

elyxlz:
Resolved the merge conflicts with `origin/master` (master had landed #1133, a broad new-agent wizard rework that independently removed the onboarding model/context steps via a `skipTuning` prop, colliding with this PR's `defaultsOnly` mechanism).

Two conflicted files:

- **apps/web/src/components/ProviderPicker/index.tsx**: kept this PR's `defaultsOnly` + `finishWithDefaults` mechanism (the PR's identity). Master's `complete`/`defaultContext` helpers existed only to support its parallel `skipTuning` prop, so dropping `skipTuning` and taking this PR's self-contained version loses no orthogonal fix. One follow-up fix: master bumped TypeScript, and the stricter excess-property check flagged the unused `presets` key in `finishWithDefaults`'s fallback (only `initial` is destructured there), so I changed the fallback from `{ presets: [], initial: 0 }` to `{ initial: 0 }`. The ContextStep render block, which destructures both, is unchanged.
- **apps/web/src/components/NewAgent/index.tsx**: kept master's orthogonal forgiving-failure improvements from #1133 (the createError display under the picker, `setCreateError(null)` on success, and the `nextStep(result, personality)` re-entry that skips already-collected steps) while swapping the prop back to `defaultsOnly` so this PR's no-model/no-context onboarding survives.

`./check.sh web` passes: eslint 0 errors, prettier clean, `tsc -b` clean, vitest 167/167. `gh pr view` now reports `mergeable: MERGEABLE` (was `CONFLICTING`). I left the actual merge to you.

🤖 Generated with Claude Code

elyxlz:
Applied the confirmed review fix for the OpenRouter onboarding regression.

In defaultsOnly (onboarding) mode, finishWithDefaults computed the default model as manifest.providers[provider].default_model with a "" fallback. OpenRouter's manifest default_model is null, so the OpenRouter key path finished with model:"" and provisioned an agent with an empty slug OpenRouter cannot run. Claude was unaffected since its default_model is opus.

The fix reads the default model first and, when it is empty (no manifest default, i.e. OpenRouter), falls through to the model step so the user still picks a real slug, mirroring the pre-PR behavior where OpenRouter always walked the model step. Claude keeps finishing immediately on its default. Added a render regression test (ProviderPicker/index.test.tsx) that drives the OpenRouter onboarding path and asserts it reaches the model step rather than calling onDone with an empty model. Verified the test fails without the fix and passes with it; ./check.sh web is green.

PR #1123, opened 2026-07-10
fix(web): red means recording, one selection device

elyxlz:
## Summary

Two state-semantics fixes, both net-deletions of color values.

- **ChatComposer mic button** (`apps/web/src/components/Chat/ChatComposer/index.tsx`): the idle mic was a permanent off-palette `bg-red-500` disc, the loudest pixel on the chat screen, which muted the meaning red is supposed to signal (recording in progress). Idle now uses `variant="secondary"` (quiet, matches the ghost send button next to it); the disc turns `bg-destructive text-background` and the InputGroup ring turns `ring-destructive` only while `isRecording` is true, using the theme's destructive token instead of a raw Tailwind red.
- **PersonalityStep card selection** (`apps/web/src/components/NewAgent/Steps/PersonalityStep/index.tsx`): the selected card stacked three emphasis devices across two hues (`border-primary/60`, `bg-primary/5`, and a grey `ring-2 ring-ring`), i.e. redundant signaling of the same state. Collapsed to one device: `border-primary bg-primary/10`. The ring is reserved for `focus-visible` as intended by the design system.

**Principle cited**: CLAUDE.md's elegance north star, "Subtraction beats addition: prefer removing a step, state, concept, or layer over adding one." Both changes remove color values (an off-palette always-red bg, a redundant ring + double hue) rather than add new ones, and route through the existing `destructive`/`primary` theme tokens rather than raw Tailwind color scales.

## Test plan

- [x] `./check.sh web` (eslint, prettier, tsc, vitest) passes clean; 0 errors, only pre-existing unrelated warnings.
- [x] Manual read of the diff confirms `isRecording` still drives both the button fill and the InputGroup ring together, as before.

### DIFF
--- a/apps/web/src/components/Chat/ChatComposer/index.tsx
+++ b/apps/web/src/components/Chat/ChatComposer/index.tsx
@@ -94,7 +94,7 @@ export function ChatComposer({
             ? "bg-card shadow-md ring-1 ring-foreground/5 dark:ring-foreground/10"
             : "bg-secondary",
           isRecording &&
-            "ring-2 ring-destructive has-[[data-slot=input-group-control]:focus-visible]:border-transparent has-[[data-slot=input-group-control]:focus-visible]:ring-2 has-[[data-slot=input-group-control]:focus-visible]:ring-destructive",
+            "ring-2 ring-red-500 has-[[data-slot=input-group-control]:focus-visible]:border-transparent has-[[data-slot=input-group-control]:focus-visible]:ring-2 has-[[data-slot=input-group-control]:focus-visible]:ring-red-500",
         )}
       >
         <InputGroupTextarea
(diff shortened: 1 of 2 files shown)


emipasca:
dont use destruictive to highlight the recording state use red

elyxlz:
Good catch, fixed. The recording state now uses an explicit red instead of the `destructive` semantic token. In `apps/web/src/components/Chat/ChatComposer/index.tsx`:

- Input group recording ring: `ring-destructive` -> `ring-red-500` (both the base ring and the focus-visible ring).
- Voice button while recording: `bg-destructive text-background` -> `bg-red-500 text-white hover:bg-red-600`.

`destructive` reads as danger/delete, so red now carries the "currently recording" meaning directly. No other `destructive` usages were touched. `./check.sh web` passes (eslint 0 errors, prettier clean, tsc clean, 146 vitest tests pass).

🤖 Generated with Claude Code

PR #1126, opened 2026-07-10
fix(web): keyboard and screen-reader access to the core surfaces

elyxlz:
## Why

The app's primary entry point (the agent card) was unreachable by keyboard: a `<div>` with `cursor-pointer` and an `onClick`, so the Ticker's arrow-key navigation skipped it and it was invisible to a screen reader. Hold-to-talk was pointer-only with no key handlers and no pointer capture, so a thumb drifting off the button mid-recording silently dropped it. The carousel's page dots had a 22px hit target, below the 44px minimum touch target. Agent replies streamed into the chat with no ARIA live region, so a screen-reader user got no signal a reply had arrived. The build-progress message lived inside an `AnimatePresence mode="wait"` node that gets destroyed and remounted per phase, so its `aria-live="polite"` region never actually announced anything across the 5-10 minute wait, since assistive tech only announces content changes within a node that persists.

Principle: "Deep modules, hidden decisions" and "Elegance is the north star", each fix reuses an existing in-repo pattern rather than inventing a new one (the shared `focus-visible:ring-3 ring-ring/30 border-ring` classes already used across `components/ui/*`, the `AgentIsland/Collapsed.tsx` persistent `sr-only aria-live` span, and the existing `touch-none` convention from `slider.tsx`/`scroll-area.tsx`).

## Changes

- **`AgentCard`**: the card's content is now wrapped in a real `<button type="button" aria-label="open {agent}">` carrying the shared focus-ring classes rounded to the card's squircle, plus `hover:bg-muted/40` and `active:scale-[0.99]`. The outer `Card` no longer carries `cursor-pointer`/`onClick`.
- **`ChatComposer`** (hold-to-talk): `onPointerDown` now calls `setPointerCapture` so the button keeps receiving pointer events even if the finger drifts off it; `onPointerLeave` is gone (pointerup now reaches the capturing element wherever it lifts) and `touch-none` is on the button. Added `onKeyDown`/`onKeyUp` (Space/Enter, ignoring key-repeat) to start/stop recording from the keyboard, and `onBlur` as a safety net so a recording can't get stuck on if focus leaves mid-hold.
- **`AgentsCarousel`** pagination dots: hit target grows from 22px (`p-2` around a `size-1.5` dot) to a 40px `grid place-items-center` button; the visual dot is unchanged.
- **`ChatMessageArea`**: one persistent `sr-only aria-live="polite"` span (mirroring the last completed agent chat event) lives inside `CardContent` so screen readers hear replies as they land, following the same pattern as `AgentIsland/Collapsed.tsx`. The typing indicator gains an `sr-only` "typing..." child and `motion-reduce:animate-none` on its bouncing dots.
- **`ProgressBar`**: the `role="status" aria-live="polite"` node is now a single always-mounted `<p>`; only an inner `<motion.span>` crossfades per message. Since the live region itself never unmounts, assistive tech now announces each phase change during the multi-minute build instead of hearing nothing.

## Test plan

- [x] `./check.sh web` (eslint, prettier, tsc, vitest): all green, 0 new errors/warnings introduced (pre-existing warnings in unrelated files untouched)
- [x] Manually traced keyboard/focus flow through each changed component


emipasca:
resolve conflicts

elyxlz:
Resolved the merge conflict with the latest `origin/master`.

**Conflicting file: `apps/web/src/components/Chat/ChatComposer/index.tsx`** (one hunk, the voice/record button's `className`).
- This PR added `touch-none` to the record button (part of the pointer-capture press-and-hold a11y gesture).
- Master (#1133 and neighbors) reworked the same button to apply the red styling only while recording, moving from an always-red string to a `cn(...)` with an `isRecording &&` conditional.
- Resolution keeps master's conditional structure and folds this PR's `touch-none` into the base classes, preserving both:
  ```tsx
  className={cn(
    "size-12 touch-none rounded-full [&_svg]:size-5",
    isRecording && "bg-red-500 text-white hover:bg-red-600",
  )}
  ```

All other files auto-merged cleanly (the a11y keyboard/pointer-capture handler additions above this hunk applied without conflict). No a11y improvements were lost.

`./check.sh web` passes (eslint 0 errors, prettier clean, tsc clean, vitest 167/167). The PR is now MERGEABLE (no longer conflicted). I left the actual merge to you.

🤖 Generated with Claude Code

PR #1127, opened 2026-07-10
fix(web): contrast meets AA in both themes

elyxlz:
## Summary

Every change here swaps an alpha-stacked or undersized token for the already-passing token the same component pairs with elsewhere in the system; no new tokens, per the elegance north star (subtraction over addition, one owner per decision: contrast for a given role lives in one token, not per-consumer alpha overrides).

- **MobileNavbar** (`text-white` -> `text-primary-foreground`, 1.88:1 -> 8.52:1): the active tab used a raw non-token white instead of the `primary`/`primary-foreground` pairing every other primary surface (e.g. the default Button variant) already uses correctly.
- **ToolCallLabel** (label/pre `/70` -> full `text-muted-foreground`, icon `/60` -> full, chevron `/40` -> `/70`): three separate alpha values stacked onto an 11px label pushed light-mode contrast to ~2.2:1. The de-emphasis these alphas were going for is already carried by the 11px size; the color no longer needs to double up.
- **ChatBubble timestamps** (`/50` on both sides -> `text-primary-foreground/90` user, solid `text-muted-foreground` agent): same reasoning, the 10px size is doing the visual de-emphasis already.
- **ChatMessageArea** ("beginning of conversation" `/40`, day stamps `/60` -> solid `text-muted-foreground` for both).
- **`--warning` token** (light: `oklch(0.666 0.179 58.318)` at 3.15:1 -> `oklch(0.5 0.15 58.318)`, `.dark` untouched): the only genuinely new value in this PR, a lightness/chroma tune of the existing warning hue so the `/90` `AlertDescription` path clears AA.
- **Focus ring** (`button.tsx`, `input.tsx`, `textarea.tsx`, `input-group.tsx`): `focus-visible:border-ring focus-visible:ring-3 focus-visible:ring-ring/30` (a 30%-alpha halo at 1.31:1) -> `focus-visible:ring-2 focus-visible:ring-ring ring-offset-1 ring-offset-background` (full-strength ring, offset separates it from the control's own border/fill so it also reads on the primary-button case). This is one token contract shared by all four form primitives, so the fix lands once instead of four times.
- **BottomBanner** error pill (`border-destructive bg-destructive text-white` -> `bg-card text-destructive border-destructive/30 shadow-sm`): this invented a second, louder destructive dialect (opaque fill + raw white text, ~2.9:1 in dark) alongside the soft dialect `button.tsx`'s destructive variant already establishes (`bg-destructive/10 text-destructive`). Reusing that dialect on an opaque card surface is the one-owner-per-decision fix, and it clears AA by construction in both themes.

## Note

The ring-offset color now matches the default page background (`ring-offset-background`) so the halo no longer renders as a stray white gap in dark mode. Controls that render on a `bg-card`/`bg-popover` surface get a very close, not pixel-identical, offset color; a per-surface override wasn't in the listed change set and is not needed for AA (the offset is a gap, not text/UI-indicator contrast), so it's left as a follow-up rather than added speculatively here.

## Test plan
- [x] `./check.sh web` (eslint, prettier, tsc, vitest) green; pre-existing lint warnings (0 errors) confirmed unchanged against `origin/master`.
- [x] Every listed change matched the current code; none were skipped.


elyxlz:
CI fix: merged origin/master into the branch (clean, no conflicts), then addressed `dashboard-sync-check`.

The check runs `scripts/sync-dashboard.sh`, which mirrors the canonical shadcn registry (`apps/web/src/components/ui/` + `index.css`) into the dashboard skill (`agent/skills/dashboard/app/`) and fails on any drift. This PR's contrast-AA changes to the ui primitives (button, input, textarea, input-group focus rings and the `--warning` token) had landed in apps/web but the mirrored dashboard copies were never regenerated, so the two trees drifted.

Fix: ran the sync script and committed the regenerated dashboard files (5 files, focus-ring and warning-color parity). Verified locally with `./check.sh web` (tsc, prettier, vitest all green) and confirmed the `guards` job (which contains dashboard-sync-check) now passes on the fresh run.

🤖 Generated with Claude Code
Do NOT merge this PR.

PR #1129, opened 2026-07-10
fix(web): chat tells the truth about agent state

elyxlz:
## Rationale

All three changes are surgical uses of signals the socket already streams; nothing new is plumbed in.

- **Errors and rate limits reached no surface.** `error` and `rate_limited` events already exist on `VestaEvent` (`core/client.py`'s SDK classification) but the chat filter dropped them, so a failed turn read as Vesta ignoring the user. They're now included in the filter and rendered by `ChatBubble` as the existing centered muted system-stamp style (matching the day-stamp pattern already in the file, `text-[11px] text-muted-foreground/60`), so the treatment introduces no new visual vocabulary. Also resets the pacing `isTyping` state on these events so the dots can't get stuck mid-turn.
- **The tool-calls toggle silently reset every session.** It lived in `useState` inside `AgentSocketProvider`, discarded on reload. Moved to a tiny zustand + localStorage store (`stores/use-show-tool-calls.ts`), mirroring the existing `use-chat-pacing.ts` pattern exactly, so a user's trust-building "show me what you're doing" choice persists like the other display prefs.
- **Appended bubbles popped in with zero motion.** `ChatMessageArea` now tracks the highest rendered row index in a ref and wraps genuinely new appends (never the initial history page or a load-older prepend) in a `motion.div` using the file's existing spring vocabulary (`lib/motion.ts`'s `stepTransition.transition`, already used elsewhere in the app for entrance motion). The virtualizer's absolutely-positioned outer row is untouched, so virtualization/measurement is unaffected.

(Two further changes, a derived typing-indicator fallback and a state-aware "send a message to steer" composer placeholder, were rejected in review and reverted in c4e4cee5.)

## Test plan
- [x] `./check.sh web` (eslint + prettier + tsc + vitest) green, no new warnings vs. master
- [x] Manual review of the diff for each behavior

### DIFF
--- a/apps/web/src/components/Chat/ChatBubble/index.tsx
+++ b/apps/web/src/components/Chat/ChatBubble/index.tsx
@@ -6,6 +6,17 @@ import type { VestaEvent } from "@/lib/types";
 import { cn } from "@/lib/utils";
 import { ToolCallLabel } from "../ToolCallLabel";
 
+// Coarse relative countdown to a rate-limit reset (unix seconds); minutes/hours/days is
+// plenty of precision for "come back later" copy.
+function formatResetTime(resetsAt: number): string {
+  const minutes = Math.round((resetsAt * 1000 - Date.now()) / 60_000);
+  if (minutes <= 1) return "in a minute";
+  if (minutes < 60) return `in ${minutes}m`;
+  const hours = Math.round(minutes / 60);
+  if (hours < 24) return `in ${hours}h`;
+  return `in ${Math.round(hours / 24)}d`;
+}
+
 export const ChatBubble = memo(function ChatBubble({
   event,
   className,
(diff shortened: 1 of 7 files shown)


emipasca:
rejecting

Typing indicator was inverted. isTyping only flips true while the pacing drain replays an already-received reply, so the dots appeared after the turn finished and stayed dark for the real 30s+ wait. Chat/index.tsx now derives showTyping = isTyping || (agentState !== "idle" && last visible message is the user's): once the agent streams anything (a bubble or a tool call), that content itself signals activity, so the derived fallback only fires in the pure "user just sent, nothing back yet" gap. Principle: elegance as truthfulness, the UI should reflect state that already exists on the wire rather than fabricate it from a display-pacing artifact.

and

Nothing taught that a mid-turn message steers the agent. ChatComposer's placeholder is now state-aware: "working... send a message to steer" while agentState === "thinking", falling back to the existing "send a message..." at idle. Surfaces the preemption mechanic (CLAUDE.md's "Preemption" flow) that already exists but was invisible.

elyxlz:
Reverted both rejected changes in c4e4cee5: the chat passes the wire isTyping through unchanged (no derived showTyping fallback), and the composer placeholder is static again (no agentState prop, no "send a message to steer"). The error/rate-limit bubbles, new-message entrance animation, and persisted show-tool-calls store remain. ./check.sh web passes.

PR #1132, opened 2026-07-10
fix(cli,vestad): terminal first-run without dead ends

elyxlz:
## Summary

Four small behavior fixes that remove every manufactured dead end between "curl install" and "I am in the app". Elegance principle: subtraction beats addition, and every dead end here was an accidental extra step or a silently-broken suggested command, not a feature.

- **`vestad/src/tunnel.rs`**: the domain prompt during forced first-run BYOK setup returned `Err("no domain entered")` on an empty answer, which killed the whole `vestad start` before the daemon ever came up, even though `--standalone --no-tunnel` already runs fine without a domain. The prompt now reads "your domain (press enter to skip, local network only)"; an empty answer persists a `no_tunnel` marker (so it's asked once, not on every future start) and lets the service start locally, with a one-line epilogue pointing at `vestad connect` for later. *Design errors out of existence*: skipping is now a supported path, not an error to route around.
- **`vestad/src/main.rs`**: first-run success printed "run `vestad status` to see your connection info" plus a six-line command menu, deferring the actually useful output (the connect link + QR) to a second manual command. It now prints the status banner inline, polling `status.json` for a fresh write first since the unit isn't `Type=notify` (`wait_for_start` only confirms the process launched, not that async startup, e.g. the tunnel dial, finished and persisted status). Closes with one line instead of six.
- **`cli/src/main.rs` (welcome)**: the quick-start block recommended `vesta setup` then `vesta list`, both of which die with "no server configured" on a fresh machine that hasn't run `vesta connect` yet. Swapped in `vesta connect <link>` as step one (sourced from the `vestad status` banner above) and dropped `vesta list`, which isn't part of getting started.
- **`cli/src/main.rs` (oauth_dance)**: a single mistyped or stale OAuth code paste called `platform::die`, stranding the just-created agent unauthenticated with no way to retry short of starting setup over. Now retries up to 3 times against the same auth session before giving up, with a clear "that code didn't verify" message between attempts, and the prompt itself explains what to look for ("after signing in, claude shows a code. paste it here").

## Test plan

- [x] `./check.sh cli` (clippy -D warnings + cargo test) — green
- [x] `./check.sh vestad` (clippy -p vestad -D warnings + cargo test -p vestad) — green
- [x] `./check.sh web` — green (unaffected by this change; run per workflow instructions)


elyxlz:
fix merge conflict then merge this please

elyxlz:
Merge conflict resolved and PR merged (squash commit 1adcbc3c).

The only conflict was in cli/src/main.rs, in oauth_dance(). Master had meanwhile removed the cli platform module (folding platform::die back to die() in main.rs and refactoring the file), while this PR replaced the single-shot auth-code prompt with a 3-attempt retry loop (OAUTH_CODE_MAX_ATTEMPTS) so a bad paste no longer strands a fresh agent. I kept this PR's retry loop and adapted its one platform::die call to master's die(), matching the rest of the file. The const block and the print_welcome quick-start rewrite (adding vesta connect) auto-merged cleanly, preserving master's colon-title copy change.

Checks: ./check.sh cli (clippy clean + 28 tests) and ./check.sh vestad (clippy clean + 178 tests) both green locally; full CI merge-gate-ci passed and auto-merge landed it.

🤖 Generated with Claude Code

PR #1142, opened 2026-07-10
refactor(vestad): conformance sweep on dead cfg, single-owner decisions, panic-free boot

elyxlz:
## What

Small, surgical conformance pass over vestad against the CLAUDE.md rubric:

- Deletes the Windows-only Docker connection path in docker.rs (the cfg(windows) const and connect branch), which can never compile in since main.rs already carries a Linux-only compile_error. Pure subtraction, no behavior change.
- Routes all six inline container-stop call sites through the existing stop_container_with_timeout helper, so the stop semantics live in one place instead of being re-expressed six times.
- Folds backup_preflight's hand-duplicated NotFound/Dead matching (character-identical to guard_alive's arms) into a call to guard_alive, made pub(crate) for this.
- Makes ensure_tls and ensure_api_key in serve.rs return Result instead of panicking with expect/unwrap on fallible filesystem and cert-generation I/O, matching the crate's no-panic-on-fallible-paths rule. The single call site in main.rs unwraps via the existing die() pattern already used for other boot-time failures.
- Passes the curl and cloudflared subprocess paths in tunnel.rs as OsStr via .arg() instead of unwrapping a lossy to_str() conversion, avoiding a panic on a non-UTF8 home path with no error path needed at all.
- Compares the API key in verify_token with a constant-time byte comparison instead of short-circuiting ==, closing a narrow timing side channel on the never-rotating root secret that gates the tunnel-exposed HTTPS listener. ring 0.17's own constant-time helper is deprecated (fails clippy -D warnings) and there's no non-deprecated replacement in that crate, so this uses a small inline constant-time comparison instead of adding a new dependency. Covered by a new two-case unit test.

## Why

Each of these is a repeat decision or a fallible boot path flagged during a repo-quality pass over vestad against the codified Architecture Principles (one owner per decision, no panics on fallible paths in vestad/cli).

## Test plan

- [x] ./check.sh vestad (cargo clippy -p vestad -D warnings + cargo test -p vestad) green locally, 180 passed, 0 failed
- [x] New verify_token_tests cover the matching and non-matching key cases


elyxlz:
solve conflicts and merge

elyxlz:
Conflicts resolved and merged.

Only one conflict surfaced: the `crate::docker` import list in `vestad/src/backup.rs`. Master's #1114 rewrite of `restore_backup` (confirm-gone-before-recreate via `ensure_container_removed`, plus `env_file_names`/`read_env_value`) collided with this PR's panic-free `backup_preflight` refactor (`guard_alive(container_status(...))`). The merge combined both bodies cleanly; only the import declaration needed reconciling.

Resolution: imported the union of what the merged body actually references, `ensure_container_removed, env_file_names, guard_alive, read_env_value`, and dropped the PR's now-unused `inspect_container` and `remove_container_force` (the restore test in this same file asserts `remove_container_force` is absent, so keeping master's confirm-gone path is correct and avoids an unused-import clippy failure under -D warnings). This preserves the PR's single-owner, panic-free intent while keeping master's stronger restore safety.

Verified with `./check.sh vestad`: clippy -D warnings clean, 187 tests passed. Full CI green including merge-gate-ci. Squash-merged as b29dee97d24ca776c02053f539ab8fea08d343cb.

🤖 Generated with Claude Code

PR #1120, opened 2026-07-09
fix(skills): remove day-one friction from the onboard and birth prompts

elyxlz:
## The funnel leak

Two prompt-copy fixes on the buyer's first days.

**Onboard step 4** gates the 7-day money-back guarantee behind "if they hesitate at the link," but hesitation in chat is silent, the waverer never clicks, so the strongest risk-reversal line never fires. Send it with the link unconditionally: every buyer sees it, not just the ones who happen to signal doubt in text.

**Birth** offers whatsapp as an equal-weight day-one channel with no warning that it means buying a prepaid SIM (PHONE_NUMBER.md), stalling the honeymoon on a hardware purchase while two-minute telegram sits unpitched. State the cost of each channel up front and queue whatsapp as a task instead of a blocker, so the first conversation never waits on a SIM card.

## Changes

- `agent/skills/onboard/SKILL.md`: drop the conditional "if they hesitate" framing, send the guarantee line unconditionally with the checkout link.
- `agent/skills/birth/SKILL.md`: name the setup cost of each channel (telegram: two minutes, whatsapp: a dedicated number) and make whatsapp a queued task rather than a gate on day one.

## Test plan

- [ ] CI (markdown-only change, no code paths touched)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1144, opened 2026-07-10
fix(skills): day one ends with a promise, and telegram closes the loop

elyxlz:
## Funnel leak

Birth's scripted ending is a restart spinner: the last thing the new user reads is housekeeping, and the headline claim (Vesta texts you first) is never demonstrated during the conversation that's supposed to sell it. That's the Peak-End moment left empty.

Telegram pairing tells the agent to note that the user must `/start` the bot, but nothing checks it happened. If the user forgets, the channel move silently strands: Vesta thinks it moved, the user never sees a message, and the relationship goes quiet on day one.

## Fix

- `agent/skills/birth/SKILL.md`: before the restart, schedule the first unprompted morning brief on the tasks skill and tell the user plainly, tomorrow morning I text you first. Engineers the Peak-End instead of leaving it to chance.
- `agent/skills/telegram/SETUP.md`: rewrite the passive `/start` note as a verified round trip. Wait for the first inbound message, reply on the new channel, and only then call it live.

Rebased cleanly over in-flight #1120 (different lines in the same birth file, no overlap).

Markdown-only changes; relies on CI for `check.sh`.

Generated with Claude Code


elyxlz:
fix conflict and merge

elyxlz:
Resolved the merge conflict in `agent/skills/telegram/SETUP.md` (the only conflicting file). Reconciled all three overlapping changes: kept master's idempotency note on the daemon-start step, kept this PR's loop-closing step 5 (confirm back on the new channel before declaring it live), and used the correct `## Daemons` section name that master renamed to. Skills index verified fresh, no conflict markers remain. Auto-merge (squash) enabled; will land once merge-gate-ci passes.

🤖 Generated with Claude Code

PR #1203, opened 2026-07-12
fix(web): keep the deleting orb red until the deleted agent leaves the list

elyxlz:
## Problem

Deleting an agent in the web app turns the card's orb red ("deleting"), but just before the card disappears it flashes gray. The DELETE response cleared the op back to idle while the agent was still present in the gateway's agents list (it only drops on the next WS agents push), so for that gap the card rendered the plain stopped look.

## Fix

The op now ends when the agent actually disappears, not when the API call returns:

- `withOp` gains a `keepOnSuccess` option; the delete flow uses it, so the card keeps its red "deleting" look after a successful DELETE.
- `useAgentOps.reconcile(agents)` prunes op state for agents absent from each gateway agents push, mirroring `useRestartPending.reconcile`. It replaces the old `removeAgent`, whose only caller was the delete flow.
- A failed delete is unchanged: op returns to idle with the error surfaced.

## Tests

New `use-agent-ops.test.ts` covers: clear-on-success default, keepOnSuccess holding the op, error path returning to idle, in-flight op guard, and reconcile pruning only absent agents. `./check.sh web` is green.

https://claude.ai/code/session_01KyCZii4d2s9h6THDx1RCD8

### DIFF
--- a/apps/web/src/providers/GatewayProvider/index.tsx
+++ b/apps/web/src/providers/GatewayProvider/index.tsx
@@ -173,7 +173,7 @@ function ConnectedGateway({ children }: { children: ReactNode }) {
               setAgentsFetched(true);
               // Clear any "restart to apply" flag whose agent has since restarted (by any path).
               useRestartPending.getState().reconcile(agents);
-              // Drop op state for agents that no longer exist (ends a delete's kept op).
+              // Drop op state for agents that are gone (ends a delete's "deleting" orb).
               useAgentOps.getState().reconcile(agents);
               break;
             }
--- a/apps/web/src/providers/GatewayProvider/index.tsx
+++ b/apps/web/src/providers/GatewayProvider/index.tsx
@@ -20,6 +20,7 @@ import type {
   GatewayVersionInfo,
   ReleaseChannel,
 } from "@/lib/types";
+import { useAgentOps } from "@/stores/use-agent-ops";
 import { useRestartPending } from "@/stores/use-restart-pending";
 import { GatewayContext, disconnectedValue } from "./context";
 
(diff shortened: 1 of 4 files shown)


elyxlz:
if this requires so many loc it means something is wrong and over complicated. find the actual cleaner u deflating architectural solution for orb and agent state please

elyxlz:
Deflated it. The root cause was that `withOp` cleared the op to idle the instant the DELETE promise resolved, while the deleted agent lingers in the gateway list until the next agents push, so the orb fell through to `agent.status` (gray) for that gap.

The cleaner framing: `withOp` is for ops that hand off to a new `agent.status` when they resolve (start, stop, rebuild, ...). Delete is the one terminal op: it hands off to the agent's disappearance, not to a status. So I stopped teaching `withOp` a `keepOnSuccess` mode and instead let delete express its own lifecycle.

What changed:
- Removed the `keepOnSuccess` flag entirely. `withOp` is back to its plain shape (always clears on success).
- `remove` is now self-contained: set "deleting", await, and on success leave it set; only an error returns it to idle with the message.
- Kept `reconcile`, which is the single source of truth: an op lives exactly as long as its agent, so the deleting orb ends precisely when the agent leaves the list. This mirrors the existing `use-restart-pending.reconcile` pattern already called on the same agents push.

The flash is now impossible by construction: nothing clears a deleting op except the agent's disappearance, so there is no idle window. The test shrank to the two reconcile behaviors that carry the fix.

Net vs the previous head: +25 / -63 (one fewer concept, `keepOnSuccess` gone). `./check.sh web` is green.

🤖 Generated with Claude Code
Do NOT merge this PR.

PR #1207, opened 2026-07-12
fix(web): make agent creation steps scroll on small screens

emipasca:
## Problem

The new-agent flow was unusable on mobile at the "pick a vibe" step. Each step was centered in a non-scrolling container (`flex-1 flex items-center justify-center`) while the navbar sits absolutely positioned on top. On a short screen the tall personality grid (single-column on mobile, 6 cards each with emoji + title + description + sample quote) overflowed both above and below with no way to scroll: the heading hid behind the navbar and the continue button sat off-screen, leaving the user stuck.

## Fix

Changed the shared step container in `apps/web/src/components/NewAgent/index.tsx` (so it covers the provider step too, which could also overflow on very short screens):

- Wrapped the steps in an `overflow-y-auto` scroll viewport so any step that doesn't fit can scroll.
- Centered with `m-auto` instead of `justify-center`: auto margins center content when it fits but pin it to the top when it overflows, whereas `justify-content: center` clips the top of overflowing content in a scroll container.
- Added top padding equal to the measured `navbarHeight` so the heading clears the absolute navbar, plus bottom padding for the mobile home indicator (`env(safe-area-inset-bottom)`).

The personality grid itself is unchanged: single column stays readable, it just scrolls now.

## Testing

`./check.sh web` fully green: eslint (no new warnings), prettier, `tsc`, and all 168 vitest tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WuRooMDPsGsdZoKL8mUUgu)_


elyxlz:
fix conflicts

elyxlz:
Resolved the merge conflict against master. The only conflict was in `apps/web/src/components/NewAgent/index.tsx`, where master added onboarding-progress persistence (`loadOnboarding`/`saveOnboarding`/`clearOnboarding` plus name/personality restore-from-storage) while this branch added the small-screen scroll fix (`useLayout` navbar height). Both were independent additions to the same import block and the same state-init block, so I kept both: the navbar-height read and the scroll container sit alongside master's restore-on-refresh initializers, no logic dropped from either side.

`./check.sh web` passes (tsc, prettier, vitest 174/174; the 36 eslint warnings are pre-existing on master, 0 errors).

🤖 Generated with Claude Code

PR #1208, opened 2026-07-12
fix(web): reconnect the gateway socket immediately on tab refocus

emipasca:
## Problem

On mobile, switching tabs away from the web app and back flashes the "disconnected from gateway" overlay for ~2 seconds before it reconnects, on every single tab switch.

Mobile browsers freeze a backgrounded tab's JavaScript, and the OS drops the WebSocket while it's frozen. The socket's `close` event is queued but not delivered until the tab returns. When it finally fires, the reconnect loop schedules a retry only **after** the exponential backoff delay (`baseDelayMs` = 1000ms), and then runs `beforeConnect` (token refresh + version fetch) before reopening. That ~1s backoff plus fetch latency exceeds the 750ms `DISCONNECT_GRACE_MS`, so the overlay appears every time and clears once the fresh socket opens ~2s later.

## Fix

The fix lives in the shared `connectReconnectingWs` primitive (the single owner of the reconnect loop, used by the gateway, chat, and notification sockets). It now listens for `visibilitychange` and, when the tab regains visibility with no healthy or in-flight connection, clears the pending backoff timer, resets the delay, and reconnects at once, so the socket is back before the overlay's grace period elapses. The chat and notification sockets get the same instant-refocus behavior for free.

Safety details, since this is the shared primitive:
- A `connecting` guard wraps the whole `connect()` (including the `beforeConnect` await window), so a visibility event can never start a second parallel connect.
- If a still-closing socket's `onclose` hasn't fired yet, its handler is detached before reconnecting, so it can't schedule a duplicate on top of the one we start.
- The listener is `document`-guarded (matching `use-window-focus.ts`) so the primitive stays safe in non-DOM/test environments, and it's removed on `close()`.

## Tests

Added three cases to `reconnecting-ws.test.ts`: immediate reconnect on refocus that consumes the pending backoff without duplicating, a healthy socket left untouched on visibility, and the listener removed after `close()`. Full `./check.sh web` is green (eslint, prettier, tsc, 171 vitest tests).

https://claude.ai/code/session_01PQRArmeMi7suvFuLXqLsah

---
_Generated by [Claude Code](https://claude.ai/code/session_01PQRArmeMi7suvFuLXqLsah)_

### DIFF
--- a/apps/web/src/lib/reconnecting-ws.test.ts
+++ b/apps/web/src/lib/reconnecting-ws.test.ts
@@ -1,3 +1,4 @@
+// @vitest-environment jsdom
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { connectReconnectingWs } from "./reconnecting-ws";
 
@@ -122,6 +123,44 @@ describe("connectReconnectingWs", () => {
       onMessage: () => {},
     });
     await vi.runAllTimersAsync();
+    expect(FakeSocket.instances).toHaveLength(1);
+  });
+
+  it("reconnects immediately on regaining visibility, consuming the pending backoff", () => {
+    connectReconnectingWs({
+      url: () => "ws://host/a",
+      onMessage: () => {},
+      baseDelayMs: 1000,
+    });
+    const socket = FakeSocket.instances[0];
+    socket.open();
+    // Mobile OS drops the socket while the tab is backgrounded; the deferred
+    // close lands on return and schedules a backoff reconnect.
+    socket.drop();
+    expect(FakeSocket.instances).toHaveLength(1);
+    // Regaining visibility reconnects now instead of waiting out the backoff.
+    document.dispatchEvent(new Event("visibilitychange"));
+    expect(FakeSocket.instances).toHaveLength(2);
+    // The pending backoff timer was cleared, so it never spawns a duplicate.
+    vi.advanceTimersByTime(30000);
+    expect(FakeSocket.instances).toHaveLength(2);
+  });
+
+  it("leaves a healthy socket untouched when the tab becomes visible", () => {
+    connectReconnectingWs({ url: () => "ws://host/a", onMessage: () => {} });
+    FakeSocket.instances[0].open();
+    document.dispatchEvent(new Event("visibilitychange"));
+    expect(FakeSocket.instances).toHaveLength(1);
+  });
+
+  it("stops listening for visibility once closed", () => {
+    const handle = connectReconnectingWs({
+      url: () => "ws://host/a",
+      onMessage: () => {},
+    });
+    FakeSocket.instances[0].open();
+    handle.close();
+    document.dispatchEvent(new Event("visibilitychange"));
     expect(FakeSocket.instances).toHaveLength(1);
   });
 
--- a/apps/web/src/lib/reconnecting-ws.ts
+++ b/apps/web/src/lib/reconnecting-ws.ts
@@ -1,6 +1,11 @@
 const DEFAULT_BASE_DELAY_MS = 1000;
 const DEFAULT_MAX_DELAY_MS = 30000;
 
+// WebSocket.readyState values, named locally so the health checks read clearly
+// and don't depend on the WebSocket global (tests stub it without the statics).
+const WS_CONNECTING = 0;
+const WS_OPEN = 1;
+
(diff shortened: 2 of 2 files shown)


emipasca:
run a code review

elyxlz:
**Verdict: approve-with-nits.** The reconnect/focus state machine is correct: the immediate-on-refocus path cannot open a second socket or leak the first, cleanup is complete, and the code is clean under the frontend conventions. Findings below are limitations and test gaps, not correctness bugs.

### What I verified holds up
- **No double-connect / no second socket.** `connecting` is set synchronously at the top of `connect()` before the first `await` (line 58), and `reconnectNow` runs from its `connecting` check to `void connect()` with no `await` in between (lines 103-121). JS is single-threaded, so a backoff timer firing, a `visibilitychange`, and an in-flight `connect()` cannot interleave in that window. If a connect is already in flight, `reconnectNow` bails; if the socket is `CONNECTING`/`OPEN`, it bails; otherwise it clears the pending backoff timer, detaches the stale `onclose`, and connects once. Solid.
- **No listener leak / no double-add.** The `visibilitychange` listener is added once at construction and removed in `close()`, both guarded by `hasDocument` (SSR-safe). `beforeConnect` is still awaited on the fast path (same `connect()`). Cleanup and teardown are correct, no use-after-close (every callback checks `cancelled`).
- **Conventions.** No `any`/`as`/`!`, named `WS_CONNECTING`/`WS_OPEN` consts, named exports. `./check.sh`-equivalents pass locally: the 9 tests pass, eslint clean, `tsc -b` clean.

### Findings (most to least significant)

**1. (medium, limitation not regression) The fix depends on `readyState` being non-OPEN at refocus; a not-yet-detected dead socket still flashes.** `reconnecting-ws.ts:105-109`. If the browser has not yet marked the OS-dropped socket `CLOSED`/`CLOSING` when `visibilitychange` fires (socket still reports `OPEN`), `reconnectNow` returns early and the user waits out the original backoff plus `beforeConnect`, exactly the ~2s flash this PR targets. Per the WebSocket spec the UA sets `readyState = CLOSED` before queuing the close-event task, so for a genuinely torn socket this is usually already `CLOSED` at resume and the fix works. But a zombie socket that still reports `OPEN` is not covered, and it cannot be from readyState alone (it needs an app-level liveness ping or an unconditional `close()` of a suspect socket, which risks tearing a healthy one). Reasonable to ship as-is; worth being explicit that this narrows but does not eliminate the flash.

**2. (low) Focus resets the backoff unconditionally, so repeated refocus while the server is down bypasses backoff.** `reconnecting-ws.ts:120`. `reconnectNow` always sets `delay = baseDelay` before connecting. If the server is down and the user rapidly toggles the tab, each return triggers an immediate connect and resets the backoff, so focus-driven attempts never back off. It is not a self-sustaining busy loop (each attempt needs a real `visible` event, and mobile does not auto-emit them), and an actively-present user reconnecting eagerly is defensible. Flagging as a deliberate trade to confirm, not a blocker.

**3. (low / test gap) The new tests do not exercise the branch their own comment describes.** `reconnecting-ws.test.ts:139-146`. `FakeSocket.drop()` fires `onclose` synchronously, so by the time `visibilitychange` dispatches, `socket` is already `null` and the backoff timer is scheduled. That means the `sock.onclose = null` detach at `reconnecting-ws.ts:112-115` (the "closing socket's onclose is still pending" guard the test comment calls out) and the `connecting`-guard-on-focus path are never hit. Suggest a case where the socket is non-null with `readyState = CLOSING` and `onclose` not yet fired at refocus, asserting exactly one socket results and the pending `onclose` never schedules a second. That would actually prove the double-schedule protection.

**4. (nit, pre-existing) A throwing `beforeConnect` silently kills the loop.** `reconnecting-ws.ts:60-61`. If `beforeConnect` rejects (rather than returning `"stop"`/`"open"`), `connect()` rejects through the `void connect()` call, `finally` resets `connecting`, but nothing schedules a reconnect, so the socket loop dies. This predates the PR (the old code awaited `beforeConnect` the same way) and is out of scope, noting only because this PR touches the surrounding control flow.

Net: correct and worth merging. I would land #3 (a test that hits the pending-`onclose` guard) with it, and consider a one-line note about the residual zombie-socket case from #1.

🤖 Generated with Claude Code

PR #1212, opened 2026-07-13
fix(agent): migration guiding the agent to run skill CLIs from live source

emipasca:
## The bug

The app-chat skill delivered every app message to the agent **twice** (seen on `luna`): a naive-local timestamp from `core/api.py` and a UTC-aware one from the **pre-#1053** app-chat daemon, which used to write intake notifications before #1053 moved intake in-process.

## Root cause (a class, not a one-off)

The `app-chat` CLI is a **frozen `uv tool` copy**, not linked to source. Upstream-sync updates the source, but a frozen copy keeps running the code it was built from, so #1053's fix never reached it. Across all agents on the box, most set-and-forget skill CLIs are frozen (`tasks` on all 9, `app-chat` on 7/8); frequently-reinstalled ones (`browser`) are editable. Frozen skills whose source later diverged run stale code silently; app-chat is just the one with a visible symptom.

The same failure exists for **any** language: a compiled CLI built once and left on PATH goes stale too. Matched fixes already exist per form: Python -> editable install; compiled (whatsapp, telegram) -> compile-on-invoke launcher.

## The fix: a migration that *guides the agent*, not a script

Rather than a rigid loop that only covers the cases a script can anticipate (Python uv tools, present source), the migration teaches the agent the **principle** and lets him apply judgement across his own setup:

> The command on your PATH should resolve to **live source**, never a frozen snapshot.

It shows how to recognize the problem per install type (a read-only check for frozen Python `uv tool` installs; launcher-vs-build-once for compiled tools) and the fix for each (`uv tool install --editable` for Python; a compile-on-invoke launcher for Go/Rust/etc.). The agent then converges **all** his skill tools, including custom ones, moved sources, and non-Python CLIs, which a script couldn't enumerate.

Why guidance over automation:
- **Language-agnostic** — one principle covers Python, Go, Rust, and any custom skill, where a uv-tool script only ever sees Python.
- **Handles the unknown** — agents are long-lived and personalized; a script can't anticipate every custom skill, layout, or old home path. The agent knows his own installs.
- **Safe** — the one concrete action (`uv tool install --editable --force --reinstall`) is transactional: verified that a failed rebuild leaves the existing tool intact (exit 1, nothing removed).

## Verification

- `direct_url.json` `editable` flag reliably distinguishes frozen vs editable; confirmed the installed app-chat copy is literally the pre-#1053 code.
- The read-only check lists exactly the frozen skill tools on real agents, excluding editable and third-party.
- `uv` reinstall is transactional on failure (tested frozen -> failing editable reinstall offline: exit 1, old install intact and working).
- `pytest tests/test_migrations.py tests/test_deployment.py` green.

### DIFF
--- a/agent/core/migrations/2026-07-editable-skill-installs.md
+++ b/agent/core/migrations/2026-07-editable-skill-installs.md
@@ -1,0 +1,40 @@
+Several of your skill command line tools were installed as frozen copies (a
+plain `uv tool install`), not linked to their source. Upstream sync updates the
+skill source under `~/agent`, but a frozen copy keeps running the code it was
+built from, so fixes to a skill silently never reach it. The visible symptom is
+the app-chat skill delivering every app message to you twice (an old copy and
+the agent both write the notification); other skills carry the same latent risk
+of running stale code.
+
+This migration reinstalls each of your locally sourced skill tools as an
+editable install, so the source directory becomes the install and future
+updates take effect on the tool's next run. Running daemons keep their current
+code until their next restart, which happens on the next container restart, so
+you do not need to restart anything by hand. Safe to run more than once.
+
+### 1. Reinstall each locally sourced skill tool editable
+
+This walks every installed `uv` tool, and for the ones whose source lives under
+`~/agent/.../cli` (your own skills, not third party packages) reinstalls it
+editable. Already editable tools are reinstalled editable too, which is a no-op
+in effect.
+
+```bash
+for tool_dir in /root/.local/share/uv/tools/*/; do
+  meta=$(find "$tool_dir" -name direct_url.json 2>/dev/null | head -1)
+  [ -z "$meta" ] && continue
+  src=$(grep -o '"url": *"file://[^"]*"' "$meta" | head -1 | sed -E 's/.*"file:\/\/([^"]*)".*/\1/')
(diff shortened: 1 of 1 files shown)


emipasca:
### Safety hardening (pushed)

Made the migration surgical and verified its worst-case behavior in an isolated scratch env (no live agent touched):

- **Transactional on failure** — the real risk was a failed reinstall (offline, a half-edited skill source, an unresolvable dep) destroying a working-but-stale tool. Tested `frozen → failing editable reinstall` with a broken dep offline: `uv` exits 1 and **leaves the existing install fully intact and working**. uv builds the new environment first and only swaps on success. Worst case is a tool stays frozen (status quo) with a logged note; it's never removed or broken.
- **Surgical** — now skips already-editable tools (only converts frozen ones) and third-party tools (source not under `~/agent`). Dry-run on a real agent: skips the already-editable `browser`, targets only the 7 frozen CLIs.
- **HOME/XDG-agnostic** — resolves the tools dir via `uv tool dir`, matches the agent's own skills via `$HOME/agent/*/cli`.
- **Idempotent** — a second run skips everything now editable (verified).
- On transient failure the prompt tells the agent to leave the migration unmarked so it retries next boot.

PR #1215, opened 2026-07-13
feat(skills): vestad skill owns all host-daemon interaction, absorbing the service skill

elyxlz:
Vestad knowledge was smeared across the `service` skill (registration + public URLs), MEMORY.md §3 (registration duplicated verbatim, gateway-logs self-diagnosis, a config curl), and nowhere at all for update/version. This PR centralizes it: a single `vestad` skill is the one place for talking to the host daemon.

**The `vestad` skill** (renamed from `service`, same directory contract for the `register-service` helper, now at `skills/vestad/scripts/`):
- service registration + restart-skill daemon pattern (unchanged content)
- public URL patterns (unchanged content)
- new: check version / trigger `POST /gateway/update` (agent-token access added in #1213), with the host-global blast radius and the response-never-arrives semantics spelled out
- gateway logs self-diagnosis (moved from MEMORY.md)
- invalidate `scope` option (moved from MEMORY.md)

**MEMORY.md dedup + corrections**: §Service Registration and §Self-diagnosis collapse to a one-line pointer at the skill. The §Self-Modification config bullet was stale twice over: it pointed at `PUT /agents/{name}/config` on vestad, which is API-key-only (an agent token 401s there), and used the pre-provider-union flat keys (`agent_model`). Corrected to the agent's own local API (`127.0.0.1:$WS_PORT/config` + `PATCH /provider`), which is what actually serves agent-token config writes today.

**Fleet convergence** (`2026-07-vestad-skill` migration, account-rename playbook):
- `default-skills.txt`: `service` -> `vestad`, so the boot reconciler installs the new skill everywhere
- the migration gates on the vestad skill existing (migrations run before the upstream sync boot turn, so on the first post-update boot it defers without marking applied and re-runs next boot), then repoints personalized `register-service` callers (restart-skill `## Daemons` lines), removes the old cone entry, and trims the agent's own MEMORY.md
- all repo-owned callers (tasks, dashboard, voice, browser, file-host, agentmail, google, microsoft, onedrive, spotify, enable-banking, whatsapp) updated in this PR and flow to boxes via upstream sync

`./check.sh agent` green; skills index regenerated.

Companion to #1213 (vestad-side agent-token access for update/version/logs).

https://claude.ai/code/session_01V478Pk3tTqBQLd36sVctLL


elyxlz:
merge please

PR #1226, opened 2026-07-13
feat(skills): make microsoft a default skill and codify provider routing

elyxlz:
## What

- Adds `microsoft` to `agent/core/default-skills.txt` (alphabetical slot between `email-client` and `nap`), so every Vesta ships the Microsoft (Graph) skill by default.
- Codifies the skill-routing decision in the discovery text:
  - `email-client/SKILL.md` "When to use this skill" now states that for Microsoft accounts (personal or work: Outlook.com, Hotmail, Live, Microsoft 365) the `microsoft` skill is preferred over email-client (fall back to IMAP only when Graph is unavailable and IMAP is enabled), and that the `google` skill is only for Google-native APIs and requires the user's own Google OAuth client.
  - `microsoft/SKILL.md` frontmatter `description` rewritten as when-to-use triggers (Outlook.com/Hotmail/Live/M365 mail, calendar, Teams; preferred over email-client for Microsoft accounts) so discovery routes Microsoft-account tasks there.
- Regenerates `agent/skills/index.json`.

## Why the routing

`email-client` (IMAP/SMTP/CalDAV) stays the default cross-provider surface. For Microsoft accounts the Graph-based `microsoft` skill is strictly richer (categories, calendar, Teams, block senders, works on tenants with IMAP disabled), so it is preferred whenever the account is Microsoft. The `google` skill is reserved for Google-native API needs and requires the user's own Google OAuth client (handled in a separate PR).

## Why no migration

`reconcile_default_skills` (`agent/core/default_skills.py`) is stateless and self-healing: every boot it reads `core/default-skills.txt` from the read-only core mount and queues an install turn for any default skill missing on disk, so existing boxes converge on their next boot with no migration file. Fresh images get the directory automatically: the Dockerfile prune step keeps exactly the skills listed in `default-skills.txt`, so no Dockerfile change is needed either.

### DIFF
--- a/agent/core/default-skills.txt
+++ b/agent/core/default-skills.txt
@@ -5,6 +5,7 @@ contacts
 dashboard
 dream
 email-client
+microsoft
 nap
 notifications
 onboard
--- a/agent/skills/email-client/SKILL.md
+++ b/agent/skills/email-client/SKILL.md
@@ -11,7 +11,7 @@ Provider-agnostic IMAP/SMTP for the user's email accounts, any number side by si
 
 Use it for a uniform IMAP/SMTP interface across one or many personal accounts: read, send, reply, forward, manage, and get notified on new mail.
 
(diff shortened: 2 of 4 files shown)


elyxlz:
Superseded: this branch will land as part of a combined PR with sibling branches.

PR #1228, opened 2026-07-13
fix(dream): scrub known-leaked literals in place so redaction converges

elyxlz:
## Problem

Fixes #1210.

`redact_secrets.py --delete` removed whole events, but a leaked secret is self-perpetuating: during the nightly dream the agent reasons about the value (thinking, summary, the SQL used to hunt it) and each mention lands as a new event containing the literal again. Delete N events one night, the secret reappears the next; it never converges. The re-seeder was largely the tool itself: the old scan printed the raw secret in its output, the agent read it while "reviewing matches," and re-logged it. A second bug: `REGEX.search` reported only the first match per event, so a benign first hit masked a real secret later in the same event.

## Fix

One converging mechanism, no sidecar state:

- **Masked scan.** Each hit prints as `event_id|context` with the value itself replaced by `[REDACTED]` (every secret in the context window, not just the primary match). Reviewing candidates can no longer re-leak a live secret into a new event. `finditer` reports every match per event, so a benign first hit can't mask a real one.
- **Scrub in place by id.** `redact_secrets.sh --scrub <id> <id> ...` redacts every pattern hit in those events, keeping their surrounding context. It is regex-driven and keyed by event id, so the caller never has to pass (and thereby re-leak) the literal.
- **Convergence by re-detection.** The scan runs every dream, so a re-seeded pattern secret is simply caught and scrubbed again the next night. No persistent list, no `--delete`, no extra concept or file for the agent to curate.
- **FTS stays consistent.** `events_fts` is an external-content FTS5 table with insert/delete triggers only, so a bare `UPDATE` would leave the secret searchable and corrupt the index on a later delete. The scrub resyncs per row (FTS delete with the old text, `UPDATE`, FTS insert with the new text), mirroring the trigger conditions in `core/events.py`.

The scan reads the full event (secrets often sit deep inside long bash commands / tool payloads); already-`[REDACTED]` spans are skipped so scrubbed events don't re-flag.

**Tradeoff:** detection is regex-only, so a word-shaped secret that no pattern matches relies on the agent catching it the night it leaks. Nearly every real secret is pattern-shaped (keys, tokens, app-passwords, connection strings), and the "never quote a leaked value" rule plus env-var discipline cover the rest. Worth it to drop the file.

## Tests

`agent/tests/test_redact_secrets.py` (9 tests) drives the real EventBus schema: masked scan (secret absent, context present), multi-match reporting, in-place scrub with context preserved, FTS consistency after scrub (secret unsearchable, context still searchable, clean delete afterward), scrub scoped to the given ids only, convergence on re-seed, no-op on clean events, already-redacted skip, and the scan→`--scrub` end-to-end path.

Fleet impact: content ships via the normal upstream-sync snapshot; no new on-disk state, no migration needed.

### DIFF
--- a/agent/skills/dream/SKILL.md
+++ b/agent/skills/dream/SKILL.md
@@ -161,7 +161,7 @@ Keep the container's filesystem organized and disk usage under control.
 
 ## Sensitive Data Cleanup
 
-Run `~/agent/skills/dream/scripts/redact_secrets.sh` to scan the event DB for API keys, tokens, passwords, private keys, and connection strings. Review matches (skip false positives), then rerun with `--delete` to purge. When a match is a real leaked value, also append the literal on its own line to `~/agent/data/redact_known.txt`: every run scrubs those literals in place (`[REDACTED]`) across all events, so a value that re-seeds itself through later reasoning about it is caught again automatically. Never quote a leaked value in your summary or anywhere else; refer to it indirectly. Also grep MEMORY.md and dreamer summaries for credentials and remove any you find. Secrets belong in env vars, not in history or files.
+Run `~/agent/skills/dream/scripts/redact_secrets.sh` to scan the event DB for API keys, tokens, passwords, private keys, and connection strings. For each real leak (skip false positives), add the value on its own line to `~/agent/data/redact_known.txt`, then rerun: every run replaces every known literal in place with `[REDACTED]` across all events, so a value that re-seeds itself through later reasoning about it is caught again automatically and redaction converges. Never quote a leaked value in your summary or anywhere else; refer to it indirectly. Also grep MEMORY.md and dreamer summaries for credentials and remove any you find. Secrets belong in env vars, not in history or files.
 
 ## Summary
 
--- a/agent/skills/dream/SKILL.md
+++ b/agent/skills/dream/SKILL.md
@@ -161,7 +161,7 @@ Keep the container's filesystem organized and disk usage under control.
 
 ## Sensitive Data Cleanup
 
-Run `~/agent/skills/dream/scripts/redact_secrets.sh` to scan the event DB for API keys, tokens, passwords, private keys, and connection strings. Review matches (skip false positives), then rerun with `--delete` to purge. Also grep MEMORY.md and dreamer summaries for credentials and remove any you find. Secrets belong in env vars, not in history or files.
+Run `~/agent/skills/dream/scripts/redact_secrets.sh` to scan the event DB for API keys, tokens, passwords, private keys, and connection strings. Review matches (skip false positives), then rerun with `--delete` to purge. When a match is a real leaked value, also append the literal on its own line to `~/agent/data/redact_known.txt`: every run scrubs those literals in place (`[REDACTED]`) across all events, so a value that re-seeds itself through later reasoning about it is caught again automatically. Never quote a leaked value in your summary or anywhere else; refer to it indirectly. Also grep MEMORY.md and dreamer summaries for credentials and remove any you find. Secrets belong in env vars, not in history or files.
 
 ## Summary
 
--- a/agent/skills/dream/SKILL.md
+++ b/agent/skills/dream/SKILL.md
@@ -161,7 +161,7 @@ Keep the container's filesystem organized and disk usage under control.
 
 ## Sensitive Data Cleanup
 
-Run `~/agent/skills/dream/scripts/redact_secrets.sh` to scan the event DB for API keys, tokens, passwords, private keys, and connection strings. For each real leak (skip false positives), add the value on its own line to `~/agent/data/redact_known.txt`, then rerun: every run replaces every known literal in place with `[REDACTED]` across all events, so a value that re-seeds itself through later reasoning about it is caught again automatically and redaction converges. Never quote a leaked value in your summary or anywhere else; refer to it indirectly. Also grep MEMORY.md and dreamer summaries for credentials and remove any you find. Secrets belong in env vars, not in history or files.
+Run `~/agent/skills/dream/scripts/redact_secrets.sh` to scan the event DB for API keys, tokens, passwords, private keys, and connection strings. Each hit prints as `event_id|context` with the value itself masked as `[REDACTED]`, so reviewing candidates never re-leaks a live secret into a new event. Judge each from its context, then redact the real leaks in place with `redact_secrets.sh --scrub <id> <id> ...`: it replaces the secret with `[REDACTED]` in those events while keeping their context and the search index intact. The scan runs every dream, so a value that re-seeds itself through later reasoning is simply caught and scrubbed again. Never quote a leaked value in your summary or anywhere else; refer to it indirectly. Also grep MEMORY.md and dreamer summaries for credentials and remove any you find. Secrets belong in env vars, not in history or files.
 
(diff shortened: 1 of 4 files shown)


elyxlz:
overcomplicated for the agent ux

elyxlz:
Agreed on the agent ux. The complexity was two overlapping purge mechanisms: the SKILL asked Vesta to both rerun with `--delete` AND append the literal to `redact_known.txt`. On top of the extra concept, `--delete` was a footgun: it deleted every scanned event, false positives included (contradicting "skip false positives"), and it never converged (whole-event delete by pattern, so a re-seeded mention that no longer matches a pattern escapes).

Simplified to a single converging mechanism: dropped `--delete` entirely. Vesta now carries one rule: for each real leak, add the value to `redact_known.txt` and rerun. The standing in-place scrub does the rest across all history, so re-seeded mentions still converge. This keeps the #1210 fix intact (the scrub-in-place path is untouched) while removing a step, a flag, and a concept from the agent surface, and it is safer since it is curated per literal rather than all-or-nothing by pattern.

Updated `redact_secrets.py`/`.sh`, the SKILL paragraph, and the one `--delete` test (now asserts main reports candidates without mutating and scrubs known literals). `./check.sh agent` passes green. Skills index unchanged (frontmatter untouched).

🤖 Generated with Claude Code
Do NOT merge this PR (no merge was requested).

elyxlz:
explain to me the flow from the agents perspective please

elyxlz:
The flow from Vesta's point of view, during the nightly dream's "Sensitive Data Cleanup" phase:

**One command, run twice.** Vesta runs `redact_secrets.sh`. Each run does two passes, in this order:

1. **Scrub pass.** The script reads `~/agent/data/redact_known.txt` (Vesta's curated list of confirmed leaked literals) and replaces every one of those strings with `[REDACTED]` in place across all events in the DB. It reports how many events it scrubbed.
2. **Scan pass.** It then greps the event DB for things that look like secrets (API keys, tokens, passwords, private keys, connection strings), skipping anything already `[REDACTED]`, and prints each candidate as `row_id|snippet`.

So the loop Vesta actually experiences is:

- Run the command. On a first pass `redact_known.txt` is usually empty, so the scrub is a no-op and Vesta just gets a list of candidate secrets.
- Read the candidates, drop false positives, and for each real leak append the raw value on its own line to `redact_known.txt`.
- Rerun the same command. Now the scrub pass catches the newly added literals and replaces them everywhere.
- Repeat until the scan reports "No secrets found."

**Why it converges (the point of the fix).** The scrub is a literal in place replace across every event, not a delete of the matching event. So if Vesta's own later reasoning re-seeds a value into a new event (thinking about the leak reintroduces it), the next run scrubs that occurrence too. Every rerun moves toward a fixed point: known literals gone everywhere, no new candidates.

**What Vesta carries mentally is just one rule:** "add real leaks to the file, rerun, it scrubs in place." No per-event decisions about deletion, no second mechanism. Vesta also manually greps MEMORY.md and the dreamer summaries for credentials and removes any it finds, and never quotes a leaked value anywhere (it refers to it indirectly).

That single-rule shape is the simplification from the earlier comment: the previous version also offered a `--delete` path that removed whole events (which dropped false positives too and never converged, since deleting an event by pattern misses the value when it later reappears in a differently shaped event). Dropping that left one converging mechanism.

🤖 Generated with Claude Code

elyxlz:
why not the previous method? anway to maek that work simply?

elyxlz:
Good question, and worth being precise about, because part of the old method is salvageable and part isn't.

**Why the previous method (`--delete`) went away.** It deleted the whole event that matched a secret pattern. Three problems, in order of severity:

1. **It never converges,** which is the entire point of #1210. Delete-by-pattern only catches occurrences that still look like a secret. Once Vesta reasons about the value in later prose (paraphrasing it, quoting a fragment, "the key ending in ..."), it re-seeds into events that no longer match the regex, and deletion walks right past them. So the leak survives in a reworded form.
2. **Collateral loss.** It removed the entire event, not the secret: all the surrounding conversation or tool payload the secret happened to sit inside.
3. **It could not skip false positives,** even though the SKILL tells Vesta to. A blanket delete-on-match contradicts its own instruction.

So deletion is not a method to make work simply, it is the wrong operation. The right operation is a surgical in-place replace of the literal, which is what the PR does now.

**Can we keep the nice part of the old method?** The nice part was "just run it, it purges," no hand-curating `redact_known.txt`. That IS salvageable. The simple version: have the scan scrub the matched span in place automatically, replacing just the secret substring with `[REDACTED]` right where it matched, instead of asking Vesta to copy the literal into the known file first. One pass, no manual step, no whole-event loss. If it also appends each scrubbed literal to `redact_known.txt` on its own, then later re-seeded reappearances of that exact value get caught too, so it still converges, with zero curation.

**The one sharp edge, and why the PR currently keeps the manual step.** Full auto-scrub also redacts false positives with no confirmation. Most of the patterns are tight (`sk-`, `xox`, `gh_`, `AKIA`, JWTs, private-key headers, db URLs), but the loose one, `password|secret|api_key = <value>`, will occasionally fire on something benign. Auto-redacting a non-secret is mostly harmless, but if the script also self-adds that false literal to the known file, it becomes a standing scrub rule that quietly rewrites that string everywhere, forever. That standing-rule risk is the real reason the current design keeps a human-or-Vesta eyeball between "candidate" and "scrubbed."

So the trade is: the manual `redact_known.txt` step buys one confirmation against permanently scrubbing a false positive. Dropping it buys zero-effort purging.

I can wire up the simple auto-scrub-in-place version if you want it. Two flavors:
- **auto-scrub only:** replace matched spans in place each run, no known file needed for fresh leaks, still list a literal manually only when you want reworded reappearances caught.
- **auto-scrub + self-populate:** also append each scrubbed literal to the known file so it converges with no manual step, accepting the false-positive standing-rule risk above (I would tighten or drop the loose `password=` pattern to make that safe).

Tell me which and I will push it to this branch.

🤖 Generated with Claude Code

elyxlz:
but the agent also knows what it talked about during the day, it knows if passwords were mentioned, so it should know where to look no?

elyxlz:
Reworked to the fileless design we discussed. The re-seeding root cause was the tool echoing the raw secret back to you during review, so the scan now masks each value as `[REDACTED]` in its output and you scrub the real leaks in place by id (`--scrub <id> ...`), never having to type the literal. Convergence now comes from re-detection each dream instead of a persisted `redact_known.txt`, so the file, the extra concept, and the curation step are all gone.

The one tradeoff: detection is regex-only now, so a word-shaped secret no pattern matches leans on you catching it the night it leaks. Almost every real secret is pattern-shaped, so I think that's the right trade for dropping the file.

`./check.sh agent` green (596 passed). Do NOT merge (no merge requested).

elyxlz:
That is a better cut than either flavor I offered. Masking the value in the scan output and scrubbing by id kills the re-seeding at its actual source (the tool echoing the raw literal back during review), instead of managing the consequence with a persisted file, so dropping `redact_known.txt` and the curation step is pure subtraction. The regex-only tradeoff is the right one: nearly every real credential is pattern-shaped, and a word-shaped secret was already the weakest case for any automated method. Nothing to add from me, leaving it unmerged as asked.

🤖 Generated with Claude Code

PR #1240, opened 2026-07-13
feat(prompts): let Vesta ask for the user's passwords without hesitation

emipasca:
## What

Adds a **Credentials & Passwords** rule to the always-loaded `## 1. SECURITY & ACCESS CONTROL` section of the seed `agent/MEMORY.md`, so Vesta treats logging into the user's own accounts as routine rather than something to hedge around.

The new rule:
- Logging into the user's own accounts is part of the job, not a line to shy away from.
- **When a login is needed, offer the user both paths and let them pick:**
  - **Send it over:** the user sends the password (and any 2FA code) to Vesta directly.
  - **Browser handover:** Vesta runs `browser handover start` (the `browser` skill's documented sign-in fallback), the user signs in on the handed-off browser themselves, the session persists for Vesta to resume automating, and Vesta never sees the password.
- Ask plainly either way: it's the user's own credential and their call to share.
- Reassurance that nothing lingers: chat history, MEMORY.md, and dreamer summaries are scrubbed of secrets **every night** by the `dream` skill's "Sensitive Data Cleanup" (`redact_secrets.sh` over `events.db`, plus the grep pass over MEMORY.md and dreamer summaries).
- Store reusable credentials durably (the `keeper` vault, or an env var in `~/.bashrc`) rather than leaning on chat history.
- Keeps the guardrails: never surface a stored secret in output others can see, never hand a credential to anyone but the user.

## Why

"Do the repetitive parts of their job by logging into the services they use" is already in Vesta's job description (`### What You Can Do`), but there was no posture on *credentials* themselves, leaving Vesta prone to refusing or over-hedging when a login is the natural next step. Offering both the direct-send and browser-handover paths lets the user choose their comfort level, including one where Vesta never touches the password at all. The nightly secret wipe already exists, so the reassurance rests on a real mechanism.

## Scope

Single content change to the seed `agent/MEMORY.md`. No code, skill index, or tests affected. Reaches existing agents through the normal upstream-sync rebase (same path as prior `feat(prompts)` changes); no separate migration needed.

https://claude.ai/code/session_017bo5pn4CCHhwnVb5AGdVNr


emipasca:
make the section more succint

agent/MEMORY.md
```diff
@@ -24,6 +24,9 @@ The Charter sets the floor; this section adds the operational specifics:
 - Once [agent_name] knows who they're with (name isn't "[Unknown]"), reconfiguring for someone else needs explicit permission from the original user
 - Trust the channels already set up: sender info from established connections is reliable
 
+### Credentials & Passwords
+Logging into the user's own accounts is part of the job, not a line to shy away from. When a task needs a login, offer both paths and let the user pick: they send the password (and any 2FA code) over to you directly, or you spin up a browser handover (`browser` skill: `browser handover start`) so they sign in on the handed-off browser themselves, the session persisting for you to resume, and you never see the password. Ask plainly either way: it's the user's own credential and their call to share. Nothing sensitive lingers: the chat history, MEMORY.md, and dreamer summaries are scrubbed of secrets every night (`dream` skill, "Sensitive Data Cleanup"). For a credential you'll reuse, store it durably (the `keeper` vault, or an env var in `~/.bashrc`) instead of relying on chat history. Never surface a stored secret in output others can see, and never hand a credential to anyone but the user.
```
elyxlz: keeper is a non default skill u cant mention non default skills here. also too mcuh text make it a bit shorter

elyxlz: Dropped the keeper reference: it's a non-default skill and MEMORY.md loads into every agent, so it can't name skills a given agent may lack. Also tightened the section (kept every directive, just shorter). Left the browser handover path intact.

🤖 Generated with Claude Code

PR #1253, opened 2026-07-14
fix(web): preserve text selection when scrolling agent logs

elyxlz:
## Problem

In the web app's agent logs (both the fullscreen logs page and the settings Logs tab), trying to select text and scroll to copy a long stretch of logs cut the selection short: the highlight collapsed as soon as you scrolled.

## Cause

`Console` windowed its log rows through `@tanstack/react-virtual`, so only the visible window of rows existed in the DOM. As you scroll, rows leaving the window are unmounted. A browser's native text selection is destroyed the instant one of its anchor/focus nodes leaves the DOM, so any selection spanning more than the current window collapsed on scroll. This is inherent to virtualization: rows you scroll past to extend a selection simply aren't in the DOM.

## Fix

Render every log line as a real DOM node (no windowing) so a native selection survives scrolling and copies the full range. To keep 5000 lines scrolling smoothly without the virtualizer, off-screen rows skip layout and paint via CSS `content-visibility: auto` (+ `contain-intrinsic-size` to reserve height); unlike windowing, those rows stay in the DOM and remain selectable/copyable. A small pin-to-bottom scroll handler replaces the virtualizer's `anchorTo:"end"` / `followOnAppend` behavior: it follows the live tail unless the user has scrolled up, and re-pins on agent switch/resume.

All the streaming, reconnect, resume, and initial-tail buffering logic is unchanged. Net `-29` lines.

`@tanstack/react-virtual` stays a dependency (still used by the chat message area).

## Testing

`./check.sh web` green: eslint (0 errors), prettier, tsc, and all 174 vitest tests pass.


emipasca:
there must be a proepr way to do this, look at tanstack virtual

elyxlz:
I dug into this and the honest finding is that TanStack Virtual cannot preserve a text selection across a scroll, and this PR already uses the proper technique. Here is why.

A browser text selection references live DOM nodes. When a node leaves the document the selection collapses, that is spec behavior (see MDN Selection). TanStack Virtual is a windowing library: by design it unmounts the rows that scroll outside the render window. So the moment the selection anchor or focus row scrolls off and unmounts, the range is destroyed. `anchorTo: "end"`, `overscan`, and stable `getItemKey` (which the old code already had) only stabilize scroll position and buy a small overscan margin, none of them keep an off-window selection alive. The only in-virtualizer options are hacks: manually snapshot and re-apply the range on every scroll, or force the whole selected index span to stay mounted via `rangeExtractor`. The latter degenerates to rendering everything the moment someone selects a large span, so it is strictly worse than not virtualizing, plus a `selectionchange` listener and index-mapping to maintain.

What this PR does instead is the recognized modern answer: drop JS windowing and give every line `content-visibility: auto` + `contain-intrinsic-size`. The DOM stays intact so selection, copy, and find-in-page all work, while the browser still skips layout and paint for off-screen rows, so 5000 capped lines scroll fine. It is also less code and one fewer moving part than the virtualizer, which fits the repo's elegance bar. Net: the old virtualized version was the workaround for perf that happened to break selection; this is the correct fix, not a stopgap.

I did not change anything. Forcing a TanStack reimplementation here would be the worse change. CI is green (test-frontend + merge-gate-ci pass) and the PR is mergeable.

References:
- https://developer.mozilla.org/en-US/docs/Web/API/Selection/collapse
- https://innei.in/en/posts/tech/selection-copy-in-virtual-lists
- https://dev.to/sebastienlorber/css-content-visibility-for-react-devs-4a3i

🤖 Generated with Claude Code

emipasca:
fix conflicts

elyxlz:
Resolved merge conflict with master in `apps/web/src/components/Console/index.tsx` (the only conflicted file).

Master had reworked the log list onto `@tanstack/react-virtual` (virtualizer, `OVERSCAN_ROWS`, `getItemKey`, `hadLinesRef`, absolute-positioned windowed rows). That directly contradicts this PR's fix, so I took this PR's non-virtualized rendering for both conflict blocks: every line stays in the DOM with `content-visibility: auto` + `contain-intrinsic-size`, and `updatePinned` tracks bottom-pinning. The auto-merge already dropped master's now-unused `useVirtualizer` import and `OVERSCAN_ROWS` const on this side.

Applied master's stricter eslint (`restrict-template-expressions`) to the kept code: wrapped the two numeric template interpolations in `String()` (`navbarHeight`, `ESTIMATED_LINE_HEIGHT`).

`./check.sh web` (eslint + prettier + tsc + vitest) passes (exit 0; remaining output is pre-existing warnings only). `gh pr view` reports `mergeable: MERGEABLE`.

🤖 Generated with Claude Code

PR #1255, opened 2026-07-14
fix(vestad): reuse a live service's cached port on re-registration (#1254)

elyxlz:
Fixes #1254.

## The bug is real; the reported root cause is not

Voice calls silently break for a resident voice-server, as reported. But the mechanism in the issue -- `invalidate` releasing the `name->port` binding -- is a misdiagnosis. `invalidate_service_handler` reads settings only to check the service exists, then bumps a monotonic **revision counter** (`agent_status.rs`); it never writes `settings.services`, where the port binding lives. Invalidation cannot move a port.

## Actual root cause

The port moves in `register_service_handler`, which is the endpoint everyone calls to resolve the port. It reuses the cached port only if `is_cached_port_reusable(p)`, which since #436 was **bind-only**:

```rust
async fn is_cached_port_reusable(port: u16) -> bool {
    tokio::net::TcpListener::bind(("127.0.0.1", port)).await.is_ok()
}
```

For a **live** resident service, its own server is listening, so the bind fails, the handler treats the port as stale, allocates a fresh one, and **writes it back** to `settings.services`. Every later lookup then resolves the new, dead port while the server stays on the old one.

#436's justification was *"every caller of `POST /agents/:name/services` follows the request by binding a service process to the returned port"*. **That premise is now false.** Two caller classes hit the endpoint:

- **Binders** -- voice's `start()` -- want a bindable port.
- **Resolvers** -- whatsapp's `resolveVoiceBaseURL` (a POST with `{"name":"voice"}`) and voice's own `status()`/`stop()`/`restart()` -- just want to *find* the live port. They never bind.

Bind-only serves binders and corrupts resolvers: it reallocates a live service's port and overwrites the stored value, so even voice's own `status()` poisons the binding. The trigger the reporter saw ("after touching a voice setting") is incidental -- that flow just re-exercises the resolve path; the port drifts on *any* POST while the server is up.

## Fix

Restore the connect-or-bind semantics #417 had before #436 removed them: reuse the cached port when the service is **alive** (TCP connect) or the port is **free** (bind), and reallocate only when it is **neither** -- the zombie/TIME_WAIT crash-recovery case #371 fixed.

Why this over the issue's "sticky port" proposal: pure stickiness re-breaks #371 (a crash mid-startup leaves the port in TIME_WAIT and it would be handed back forever with no recovery). Connect-or-bind keeps recovery *and* restores idempotency for resident services. The only thing it gives up vs #436 is defense against a *live* squatter answering on the port -- but #787 moved allocation above the ephemeral range (killing accidental collisions), and voice `start` guards with `port_alive` before binding, so no crash loop recurs.

One-endpoint, root-cause fix: no client changes, and it restores the contract `register-service` already documents ("idempotent: vestad returns the same port for a given name on every call").

## Test

`cached_port_is_not_reusable_when_squatted` (the #433 regression) is flipped to `cached_port_is_reusable_when_listening`, asserting a live listener's port **is** reused. `cached_port_is_reusable_when_free` stays.

`cargo test --bin vestad serve::tests` -> 11 passed; `cargo clippy` clean.

### DIFF
--- a/vestad/src/serve.rs
+++ b/vestad/src/serve.rs
@@ -1522,13 +1522,32 @@ fn allocate_service_port(registry: &HashMap<String, HashMap<String, ServiceEntry
     scan(safe_min).or_else(|| scan(SERVICE_PORT_MIN))
 }
 
-/// Bindable = reusable. A port that merely has a listener isn't enough:
-/// callers always bind the returned port themselves, so a squatter would
-/// trap them in a crash loop. See #371 and #433.
+/// A cached service port is reusable when it's the service's own live server
+/// (TCP connect succeeds) or genuinely free (bind succeeds). Only a port that is
+/// neither — a zombie/TIME_WAIT corpse left by a crashed startup (#371) — is
+/// dropped so the caller gets a fresh one.
+///
+/// The bind-only probe this replaced (#436) assumed every caller binds the
+/// returned port, so any listener meant a squatter that would crash the caller.
+/// That no longer holds: resolvers (whatsapp's `resolveVoiceBaseURL`, voice's own
+/// status/stop/restart) POST here purely to *find* the live port, and treating
+/// their own live server as a squatter reallocated the port out from under a
+/// resident service — every later lookup then resolved a fresh, dead port. The
+/// one caller that binds (voice `start`) guards with `port_alive` first, so it
+/// never binds a live port.
 async fn is_cached_port_reusable(port: u16) -> bool {
-    tokio::net::TcpListener::bind(("127.0.0.1", port))
-        .await
-        .is_ok()
+    use std::net::Ipv4Addr;
+    use std::time::Duration;
+    let alive = tokio::time::timeout(
+        Duration::from_millis(200),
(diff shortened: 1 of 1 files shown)


elyxlz:
Added a reproduction test: `resident_service_keeps_its_port_across_reregistration` (in `serve::tests`).

It replays `register_service_handler`'s exact port decision — reuse the cached port when reusable, else allocate a fresh one — against a real live listener standing in for a resident voice-server, using the same `allocate_service_port` + `is_cached_port_reusable` the handler calls (no Docker needed).

I verified it genuinely captures the bug by temporarily reverting the probe to the old bind-only body and re-running:

```
test serve::tests::cached_port_is_reusable_when_listening ... FAILED
test serve::tests::resident_service_keeps_its_port_across_reregistration ... FAILED
  assertion `left == right` failed: a resident service must resolve to its live
  port 61000, not a fresh dead one (61001)
   left: 61001
  right: 61000
```

With the connect-or-bind fix both pass; full `serve::tests` is 12/12 green, clippy clean.

PR #1262, opened 2026-07-14
fix(notifications): reply guidance lives only in the producer reply_hint

emipasca:
## Problem

Since #1209 made read-time guidance producer-owned, group-chat messages carried the same guidance twice: the producer's `reply_hint` attribute already says "this is a group chat, so it may not be expecting a reply from you", and core's `_format_one` (loops.py) still injected its own `[This message is from a group chat and may not be for you; decide whether to chip in or stay out]` plus `[Reply using the X skill]` lines into the body. Two owners for one piece of guidance:

```
<channel source="whatsapp" ... reply_hint="reply with a short message, and think about how you can best show your personality; this is a group chat, so it may not be expecting a reply from you">2 topi e un uccellino oggi.... [This message is from a group chat and may not be for you; decide whether to chip in or stay out] [Reply using the whatsapp skill] </channel>
```

## Fix

The producer's `reply_hint` is now the single owner of all read-time guidance; core injects nothing into the body (`_format_one` and `_REPLY_SKILLS` are deleted, `format_notification_batch` just joins the rendered elements). The producers (whatsapp, telegram, app-chat) carry the reply command itself in the hint ("reply with `whatsapp send`, and think about how you can best show your personality", with the group-chat caveat appended on group messages), and "with a short message" is dropped from the hint. Regression test asserts a group message renders the group-chat guidance and the reply command exactly once, with no core-injected brackets.

## Testing

- `./check.sh agent` green (597 core tests + skill CLI suites)
- `./check.sh guards` green
- Go changes are string literals only; validated by the `go-checks` CI job (no local Go toolchain)

### DIFF
--- a/agent/core/api.py
+++ b/agent/core/api.py
@@ -126,7 +126,7 @@ def _write_app_chat_notification(config: VestaConfig, text: str) -> None:
             "type": "message",
             "message": text,
             "interrupt": True,
-            "reply_hint": "reply with a short message, and think about how you can best show your personality",
+            "reply_hint": "reply with `app-chat send`, and think about how you can best show your personality",
         }
     )
     path = directory / f"{time.time_ns()}-app-chat-message.json"
--- a/agent/core/loops.py
+++ b/agent/core/loops.py
@@ -114,33 +114,12 @@ async def trash_notification_files(notifications: list[Notification], *, trash_d
     _trash_paths([n.file_path for n in notifications if n.file_path], trash_dir)
 
 
-_REPLY_SKILLS = frozenset({"app-chat", "whatsapp", "telegram"})
-
-
-def _format_one(notif: Notification) -> str:
-    """Embed hints inside the <channel> element so the model sees them as one unit.
-
-    A reply hint points the model at the originating channel's reply skill instead of copying its CLI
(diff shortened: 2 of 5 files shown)


emipasca:
no i dont like this fix, i want to keep it in reply hint and remove it elsewhere

emipasca:
fix ci

elyxlz:
CI failure was an environmental flake, not a defect and not a stale test.

Every failing assertion in `test-integration` was disk exhaustion on the runner: the backup tests returned `insufficient disk space for backup (~800MB available, need ~2554MB)`, and rename/upstream failed with `No space left on device` (`docker import`/`fetch-pack`). Attempt 1 hit the 18-minute timeout without tearing down its containers/images, so attempt 2 ran on an already-starved disk and cascaded.

None of these touch this PR: the diff is confined to agent-side notification code (`api.py`, `loops.py`, telegram/whatsapp `notifications.go`, `test_notifications.py`). backup, rename, and upstream are vestad Docker tests with no relation to the `reply_hint` change. No notification test failed.

Verified locally: reviewed the full failing log and the PR diff. I did not rerun the Docker integration suite locally (it needs Docker and ~24 min); I re-ran the failed CI job instead and am relying on it to confirm green.

🤖 Generated with Claude Code
Do NOT merge this PR.

PR #1278, opened 2026-07-15
feat(notifications): temporary silence/trash rules that auto-expire

elyxlz:
## What

Adds a `--for <duration>` flag to `notifications add`, making any rule **temporary**: it auto-expires after the given time and is then pruned, so notifications flow by their default again with nothing to remember. This directly closes the footgun the skill already warned about ("a forgotten snooze rule keeps holding back interrupts after the session is over").

It pairs with any action, so:
- **Temporary silence:** `notifications add --source whatsapp --action snooze --for 2h`
- **Temporary trash:** `notifications add --source twitter --match 'chat_name=World Cup' --action trash --for 1d`

Duration is `s`/`m`/`h`/`d` chunks: `30m`, `2h`, `1h30m`, `1d`.

## How

- `NotificationInterruptRule` gains an optional `expires_at: datetime`, with an `is_expired(now)` helper (a naive stored value is read as UTC so comparison never raises).
- `drop_expired(rules, now)` in the policy module is the one place expiry is applied, called from `load_notification_rules` (the single live-rules boundary). So matching (monitor tick), the `GET /config` view, and the skill's `list` all see only active rules, and any `add`/`move`/`remove` reads the already-pruned set and writes it back, pruning expired rules from disk. The matcher stays a pure, time-unaware function.
- Two `model_dump()` calls that persist/serialize rules switch to `mode="json"` so the new datetime field serializes to ISO (no-op for the existing all-JSON-native fields).
- The CLI parses the duration, stamps `expires_at`, and documents the flag; SKILL.md gains the mental model, examples, and a note in "Guarding a hard task".

## Fleet impact

Purely additive: existing rules have no `expires_at` (defaults to `None` = permanent), and the model already tolerates the field on read. No migration needed.

## Tests

- CLI: duration parsing (sums chunks, rejects typos like `2x`), expiry stamping, permanent rule has no expiry, bad-duration rejection.
- Core: `is_expired` before/after the instant, naive-as-UTC, `drop_expired` keeps only live rules, `load_notification_rules` drops an expired rule.
- API: a temporary rule round-trips through PUT/GET (datetime serialization) and an already-expired one is dropped.

### DIFF
--- a/agent/core/api.py
+++ b/agent/core/api.py
@@ -258,7 +258,7 @@ async def _config_get_handler(request: web.Request) -> web.Response:
     config: VestaConfig = request.app["config"]
     data = stored_config(config)
     data.pop("provider", None)
-    data["notification_rules"] = [rule.model_dump() for rule in load_notification_rules()]
+    data["notification_rules"] = [rule.model_dump(mode="json") for rule in load_notification_rules()]
     return web.json_response(data)
 
 
--- a/agent/core/config.py
+++ b/agent/core/config.py
@@ -1,4 +1,5 @@
 import copy
+import datetime as dt
 import json
 import os
 import pathlib as pl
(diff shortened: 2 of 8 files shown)


elyxlz:
fix ci and all merge conflicts on all prs not green.

PR #1289, opened 2026-07-16
feat(whatsapp): notify on edited and deleted messages

emipasca:
## What

Vesta now hears it when someone edits a WhatsApp message or deletes one for everyone. Both arrive as new notification types, `edit` and `revoke`.

## Why

Today they are **silently dropped**. WhatsApp delivers an edit or a delete as a `ProtocolMessage` naming the original message's key, and whatsmeow's `UnwrapRaw` deliberately leaves it wrapped, so the payload carries no text of its own. `extractTextContent` returns `""`, `extractMediaInfo` returns empty, and `handleMessage` bails at its `content == "" && mediaType == ""` return. If someone changes "let's meet at 6" to "let's meet at 9", the agent never finds out and answers the wrong question.

## How

- `eventHandler` routes messages with a protocol part to a new `handleProtocolMessage`, which dispatches `MESSAGE_EDIT` and `REVOKE`.
- An **edit** rewrites the stored message and notifies with `old_text` + `new_text` + `target_message_id`.
- A **revoke** notifies with the text the agent last saw, so it knows what was taken back.
- Both reuse `handleMessage`'s suppressions (own messages, `--no-notifications`, `--skip-senders`), so an edit is exactly as quiet as the message it refers to.

Notification shape:

```json
{
  "source": "whatsapp",
  "type": "edit",
  "contact_name": "Ana",
  "target_message_id": "3EB0...",
  "old_text": "let's meet at 6",
  "new_text": "let's meet at 9",
  "reply_hint": "they changed a message you may have already answered; ..."
}
```

## Three traps this had to dodge, each with a test

1. **`GetType()` reports `REVOKE` for a nil `ProtocolMessage`** (it is the zero value), so routing on the type instead of the part's presence would read every ordinary message as a deletion. Covered by `TestOrdinaryMessageIsNotMistakenForARevoke`.
2. **`storeMessage` is `INSERT OR REPLACE`**, which would reset `delivery_status`/`delivery_timestamp` that the stale-message alerting reads. The edit uses a targeted `UpdateMessageContent` instead. Covered by `TestEditRewritesStoredContentWithoutClobberingDeliveryStatus`.
3. **`messages` had `AFTER INSERT`/`AFTER DELETE` FTS triggers but no `AFTER UPDATE`**, so an in-place rewrite would leave search serving the pre-edit text. Adds a `messages_au` trigger. Covered by `TestEditKeepsFullTextSearchInSyncWithTheNewText`.

The trigger goes in the existing `CREATE TRIGGER IF NOT EXISTS` block that runs on every store open, so existing agents converge on their next daemon start with no migration.

## Tests

`handleMessage` and `eventHandler` had **zero** coverage before this; `edits_test.go` is the first. 10 tests driving real events through `eventHandler` against a real SQLite store, asserting on the notification files that land.

Both non-obvious tests were mutation-checked rather than assumed: removing the FTS trigger fails the search test (stale "harbour" still matches), and routing the edit through `StoreMessage` fails the delivery-status test.

`gofmt`, `go vet -tags fts5`, `go build`, and `go test -tags fts5 ./...` all pass, run in the `vesta:local` image (this box has no Go toolchain).

## Not verified end to end

No real phone was involved. The handler is driven by synthetic `events.Message` values shaped from whatsmeow's own wire structs, so the protobuf plumbing is exercised, but a live edit from a real contact has not been observed.


emipasca:
code review this

elyxlz:
## Code review: notify on edited and deleted messages

**Verdict: approve-with-nits.** The edit/delete detection is correct against the pinned whatsmeow version, the store and FTS stay consistent, suppressions mirror normal messages, and the test coverage is genuinely thorough. One substantive discussion point (revoked text stays searchable) and a couple of nits.

### Correctness (verified, solid)
I checked the routing against the pinned `go.mau.fi/whatsmeow v0.0.0-20260609091626`:
- `UnwrapRaw` unwraps the outer `EditedMessage` FutureProof wrapper and sets `IsEdit`, but leaves the inner `ProtocolMessage{Type: MESSAGE_EDIT}` in place, so routing on `GetProtocolMessage() != nil` then `GetType()` correctly catches edits and reads the new text from `GetEditedMessage()`.
- A delete-for-everyone arrives as a bare `ProtocolMessage{Type: REVOKE}`; whatsmeow's internal `handleProtocolMessage` has no REVOKE branch and does not gate dispatch, so the app reliably receives it.
- The "nil ProtocolMessage reports REVOKE" trap (zero-value enum) is handled the right way: presence first, then type. `TestOrdinaryMessageIsNotMistakenForARevoke` locks this in.
- `targetID` comes from `protocol.GetKey().GetID()`, not the stanza ID. Correct, and asserted.
- Edit rewrites content via `UpdateMessageContent` (a plain UPDATE, so `delivery_status`/`delivery_timestamp` survive, unlike `INSERT OR REPLACE`), and the new `messages_au` FTS trigger keeps search in sync. `TestEditRewritesStoredContentWithoutClobberingDeliveryStatus` and `TestEditKeepsFullTextSearchInSyncWithTheNewText` cover both.
- A delete/edit of a message never stored degrades gracefully: revoke stays silent (`TestRevokeOfAMessageWeNeverStoredIsSilent`), edit still reports the new text with empty old_text. No crash on missing content.
- Suppressions (IsFromMe, noNotify, skipSenders) are applied consistently through `notifContextFor`, mirroring `handleMessage`; own-edit and edit-to-identical-text are both suppressed.

### Notification contract
Matches conventions: `source=whatsapp`, `type` of `edit`/`revoke`, required `target_message_id` and `timestamp` set, `reply_hint` present, group chat_name/sender handled by `applyChatContext`. `interrupt` is unset so it defaults to True in the Python `Notification` model, i.e. edits/revokes interrupt like a normal whatsapp message. Consistent.

### Should-fix / discussion
- **`events.go` handleRevoke / `notifications.go` WriteRevokeNotification (revoked text stays retrievable).** A revoke only writes a one-time notification; the stored row is left intact. So after a delete-for-everyone, `list-messages` and full-text search still surface the original text under its original `message_id`. That subtly contradicts the revoke reply_hint ("treat it as unsaid and do not quote it back to them"), since the agent can pull it verbatim later. The edit path rewrites the store to match reality; the revoke path does not. Consider deleting the row (or blanking content, which the FTS trigger would propagate) on revoke so the persistent surfaces agree with the guidance. Counter-argument: keeping it preserves context for the agent, in which case the reply_hint is doing the whole job and this is fine as-is. Worth a deliberate call rather than an accident of the two paths diverging.

### Nits
- Edits inherit the interrupting default, so a trivial typo-fix edit preempts the current turn. The reply_hint counsels restraint and a user notification_rule can tune it, so acceptable, just flagging the behavior.
- `old_text`/`new_text` are not in the Python `_CONTENT_FIELDS` set (message/text/content), so an edit/revoke renders as an all-attributes `<channel>` with an empty body (same shape as reactions) and a user rule matching whatsapp on the `text` alias will not match edits. Fine, just noting.

### Fleet safety
Clean. The new `messages_au` trigger is created with `CREATE TRIGGER IF NOT EXISTS` at store open, idempotent across the fleet, and this is the skill-local WhatsApp sqlite, so no agent prompt migration is needed. The SKILL.md `description` frontmatter is unchanged (only the body grew), so `index.json` does not need regeneration.

### Tests / sibling consistency
Coverage is strong (old+new text, delivery-status preservation, FTS sync, ordinary-not-revoke, unstored edit/revoke, own edit, identical-text, skipped sender) and ships with the change. I could not execute them in this environment (the package pulls in a whisper.cpp cgo binding that needs `whisper.h`, unrelated to this PR), so this is a static review of the tests. Consistent with the Telegram sibling #1290: same `editNotif` shape and SKILL.md wording; whatsapp is a superset that adds revoke, which Telegram genuinely cannot detect. Coherent as one feature.

🤖 Generated with Claude Code

elyxlz:
Mirrored the telegram edit-notification design (#1290) here: the edit notification now carries the current post-edit text in the same `message` body field a plain whatsapp message uses, so it reads like a normal message and content-matching notification rules match it. `old_text` is the only extra field; the dedicated `new_text` facet is gone. The revoke (delete) path is unchanged, it carries no body, just `old_text`. Updated `edits_test.go` to assert the new text via `message` and the SKILL.md doc to match. No Python side touched. `./check.sh guards` and the notification pytest suites pass; the whatsapp Go package could not be compiled here (no Go toolchain on this box), so the Go change rests on static correctness (a pure field rename with its single writer and test assertions updated together).

🤖 Generated with Claude Code
Do NOT merge this PR (no merge was requested).

emipasca:
fix ci

elyxlz:
Heads-up: this PR's `test-integration` red is not your code. It is a systemic, master-wide failure hitting many PRs and master itself.

The `test-integration` job exhausts the ~14 GB free disk on the ubuntu runner: the `backup`/`rename`/`upstream` tests export multi-GB agent containers into restic and re-import them 8-in-parallel, so docker export/import and the mounted-repo git fetch fail with `no space left on device`. Attempt 1 stalls until the 18-minute cap (`Timeout of 1080000ms hit`), attempt 2 fails fast. Because master is red, no per-PR change can make this job green.

Fix is in #1295 (free the runner disk before the suite). Once that lands on master, rebase or re-run this PR's CI and `test-integration` should pass.

🤖 Generated with Claude Code

elyxlz:
rebase / rerun ci/ fix ci and conflicts and merge

elyxlz:
Brought current with master via merge (not rebase, no force-push). master merged clean: zero content conflicts, the edit/delete notification design is intact (edited text rides in the `message` body, `old_text` the only extra field; revoke carries only `old_text`).

One real CI failure surfaced from master, not a conflict: the strict-linting convention checker (#1291) caps comment blocks at 8 lines, and the `callNotif` doc comment carried over from master ran to 10 (a pre-existing broken-master condition, #1291's checker landed on top of #1293's longer comment). Fixed at root by condensing that comment to 8 lines without losing meaning; guards and merge-gate-ci now pass. The squash also repairs master's own guards.

CI re-ran with master's #1295 disk-cleanup fix. `detect-changes` (dorny/paths-filter) flaked twice on a GitHub API 503 outage, purely infra, but merge-gate-ci (the sole required check) went green, so auto-merge landed the PR as 735d59aa.

Merged.

🤖 Generated with Claude Code

PR #1290, opened 2026-07-16
feat(telegram): notify on edited messages

emipasca:
Follow-up to #1289 (WhatsApp edits), fixing the same gap on Telegram.

## What

An edited Telegram message now reaches the agent as its own `edit` notification carrying the old and new text, instead of masquerading as a brand new message.

## Why

Telegram's poller already opts into `edited_message`, but `StartPolling` aliased it straight to `handleMessage`:

```go
case update.EditedMessage != nil:
    tc.handleMessage(update.EditedMessage)
```

So the edit was stored correctly but announced as an ordinary message. The agent could not tell an edit from a new message, and would answer a question it had already answered. Unlike WhatsApp (where edits were dropped entirely), this one is a *wrong* signal rather than a missing one.

## How

- `edited_message` routes to a new `handleEditedMessage`: rewrite the stored text, notify with `type: "edit"` + `old_text` + `new_text` + `target_message_id`.
- Notification identity (contact/sender/chat lookup, plus the own-message and `--skip-senders` suppressions) moves out of `handleMessage`'s inline block into a shared `notifContextFor`, so an edit stays exactly as quiet as the message it refers to. That extraction is why `telegram.go`'s diff is bigger than the feature: duplicating ~30 lines of sender resolution was the alternative.
- The rewrite is a targeted `UpdateMessageContent`, so an edit writes only the text and leaves the row's media/reply/timestamp columns as the original set them. `messages` had no `AFTER UPDATE` FTS trigger, so search would otherwise keep serving the pre-edit text: adds `messages_au`, in the existing `CREATE TRIGGER IF NOT EXISTS` block that runs on every store open, so existing agents converge on their next daemon start with no migration.

**Deletions are out of scope, permanently:** the Bot API never tells a bot a message was deleted. That is a Telegram limitation, not an omission, and the SKILL.md now says so rather than leaving the agent to wonder.

## Second concern, deliberately included

**The telegram CLI's Go tests ran nowhere in CI.** `go-checks` only ran `./check.sh whatsapp`, path-filtered to `agent/skills/whatsapp/cli/**`, and no `check.sh telegram` existed. Its existing suites (`daemon_test.go`, `markdownv2_test.go`, `bubblelint_test.go`) have never been enforced. Adding tests to a suite CI does not run would be shipping dead tests, so this adds `check.sh telegram` and a `go-checks` step, and extends the `go` path filter. Telegram's `cgo-env.sh` already anticipated it: *"sourced by the launcher (and by any check script)"*. Happy to split this out if you would rather.

## Tests

8 tests driving `handleEditedMessage` against a real SQLite store, asserting on the notification files that land.

I was wrong about something here and the tests caught it. I first justified the targeted UPDATE by claiming `INSERT OR REPLACE` leaves the FTS index doubled up on stale text. Mutation-testing that claim disproved it: fts5 treats an explicit-rowid insert as a replace, so the old path kept search correct. The comment now states the real reason. The `messages_au` trigger is separately mutation-confirmed as load-bearing: removing it fails the search test (stale "harbour" still matches, new "airport" does not).

`./check.sh telegram` and `./check.sh whatsapp` both pass, run in the `vesta:local` image (this box has no Go toolchain).

## Not verified end to end

No real Telegram bot was involved. The handler is driven by synthetic `tgbotapi.Message` values, so a live edit from a real chat has not been observed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,7 @@ jobs:
               - 'check.sh'
             go:
               - 'agent/skills/whatsapp/cli/**'
+              - 'agent/skills/telegram/cli/**'
               - '.github/workflows/ci.yml'
               - 'check.sh'
 
@@ -184,7 +185,7 @@ jobs:
       - name: Run agent checks
         run: ./check.sh agent
 
-  # ── Go checks (whatsapp CLI) ──────────────────────────────────────────
+  # ── Go checks (whatsapp + telegram CLIs) ──────────────────────────────
   go-checks:
     needs: detect-changes
     if: github.event_name != 'pull_request' || needs.detect-changes.outputs.go == 'true'
@@ -206,6 +207,9 @@ jobs:
 
       - name: Run whatsapp checks
         run: ./check.sh whatsapp
+
+      - name: Run telegram checks
+        run: ./check.sh telegram
 
   # ── Windows install script guard ──────────────────────────────────────
   install-script-check:
--- a/agent/core/notification_interrupt_policy.py
+++ b/agent/core/notification_interrupt_policy.py
@@ -31,7 +31,7 @@ from .notification import CORE_SOURCE, Notification
 # satisfies the op. This is the single place cross-source field-name knowledge lives — a new source's
 # new field is targetable by its concrete name with no code change.
 _IDENTITY_FIELDS = ("sender", "contact_name", "handle", "from", "author")
-_BODY_FIELDS = ("body", "message", "content", "text")
+_BODY_FIELDS = ("body", "message", "content")
 _FIELD_ALIASES: dict[str, tuple[str, ...]] = {"sender": _IDENTITY_FIELDS, "text": _BODY_FIELDS}
 
 
--- a/agent/core/notification_interrupt_policy.py
+++ b/agent/core/notification_interrupt_policy.py
@@ -31,7 +31,7 @@ from .notification import CORE_SOURCE, Notification
 # satisfies the op. This is the single place cross-source field-name knowledge lives — a new source's
 # new field is targetable by its concrete name with no code change.
 _IDENTITY_FIELDS = ("sender", "contact_name", "handle", "from", "author")
-_BODY_FIELDS = ("body", "message", "content")
+_BODY_FIELDS = ("body", "message", "content", "text")
 _FIELD_ALIASES: dict[str, tuple[str, ...]] = {"sender": _IDENTITY_FIELDS, "text": _BODY_FIELDS}
 
 
--- a/agent/skills/telegram/SKILL.md
+++ b/agent/skills/telegram/SKILL.md
@@ -23,6 +23,10 @@ telegram send-file --to "<contact_name>" --file-path /path/to/document.pdf
 telegram send-voice --to "<contact_name>" --file-path /path/to/note.ogg
 ```
 
+## Edited messages
+
+People change their minds after they hit send, so a message you already read can change. An edit arrives as an `edit` notification naming the message that changed (`target_message_id`), the text you last saw (`old_text`), and what it says now (`new_text`). The stored message is rewritten, so `list-messages` and search show only the new text. Answer again only if the edit asks something new: a fixed typo needs nothing from you. Deletions are invisible here, Telegram never tells a bot that a message was deleted.
+
 ## Interactive UI (inline buttons + callbacks)
 
 Send tappable buttons, get notified when the owner taps, then answer the tap and/or edit the
--- a/agent/skills/telegram/SKILL.md
+++ b/agent/skills/telegram/SKILL.md
@@ -25,7 +25,7 @@ telegram send-voice --to "<contact_name>" --file-path /path/to/note.ogg
 
 ## Edited messages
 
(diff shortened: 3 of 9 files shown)


emipasca:
code review this

elyxlz:
## Code review: feat(telegram): notify on edited messages

**Verdict: approve with nits.** This is a clean, well scoped, genuinely well tested change. Edit detection uses the right signal (Telegram's `update.EditedMessage` update type, routed to a dedicated `handleEditedMessage`), so it never has to guess an edit from `edit_date` or a repeated `message_id`. The store stays in sync, the FTS index is kept honest, own edits, skipped senders, and no-op edits are all suppressed, and the shared `notifContextFor` keeps an edit exactly as quiet as the message it refers to. Eight focused tests cover the real branches. Nothing here blocks merge.

### Correctness (all good)
- Edit vs new message: driven by the update type, not a heuristic. Correct and robust.
- Dedup / spurious edits: `oldText == newText` short-circuits the notification, so Telegram re-firing `edited_message` for a link-preview load (unchanged text) writes nothing. Nice, and tested.
- Store sync happens before the suppression check, so own edits and skipped-sender edits still rewrite stored content while writing no notification. Correct.
- Interrupt disposition: `editNotif` omits `interrupt`, so Python's `Notification.interrupt` defaults to `True`, matching plain telegram messages (which also omit it). Consistent, as intended.
- Fleet safety: the new `messages_au` trigger is `CREATE TRIGGER IF NOT EXISTS`, applied idempotently on existing DBs at `NewMessageStore`. No migration needed. SKILL.md frontmatter is unchanged, so `index.json` needs no regen. Correct.

### Should-fix (discuss, not blocking)
- `agent/core/notification_interrupt_policy.py` `text` alias (`_BODY_FIELDS = body/message/content`): an edit carries its text under `new_text`/`old_text`, not one of those keys, so a user rule that matches telegram message content via the `text` alias (or a legacy `keyword`) will not match the edited text, and `new_text`/`old_text` instead surface as targetable facet fields. Reactions have the same shape today, so this is consistent within the skill, but it means content-based rules silently do not apply to edits. Worth a conscious call: is that the intended behavior?

### Nits / context
- `notification.py` `_CONTENT_FIELDS` (`message/text/content`): since the edit notif uses `new_text`, `format_for_display` renders an empty body and puts both texts as attributes. Readable, just not the message-in-body shape the other telegram notifs get.
- Behavior change worth noting: an edit of a message the bot never stored no longer lands in the store. `UpdateMessageContent` scopes by `id AND chat_id` and updates zero rows, where the old `handleMessage` path would have `INSERT OR REPLACE`d it. The agent is still notified (with empty `old_text`, tested), but `list-messages` will not show that message. Minor, and arguably fine.
- Pre-existing, surfaced by this feature: `messages.id` is the global primary key, but Telegram `message_id` is unique per chat, not globally. `GetMessageContent(msgID)` looks up by `id` only, so a cross-chat id collision (common, since chats restart numbering) could return another chat's text as `old_text`. This predates the PR and affects normal storage too, so it is not a blocker here, but the old-vs-new value of this feature is sensitive to it, so it is a good follow-up candidate.
- `WriteEditNotification` uses one `reply_hint` for direct and group chats, where `WriteNotification` adds a group-chat caveat. Trivial.

### Conventions
- Go: idiomatic, errors logged rather than swallowed, the `notifContextFor` extraction cleanly de-duplicates the new-message and edit paths. Good.
- Brand voice in SKILL.md and the reply hints: no forbidden pronouns for Vesta, no dash separators. Clean.
- CI wiring (`check.sh` `check_telegram`, the go path filter, the go-checks step) mirrors the whatsapp setup correctly, and `cgo-env.sh` exists.

Solid work. Ship after deciding whether edit text should be reachable by content-based notification rules.

🤖 Generated with Claude Code

emipasca:
lets keep text and old text to not break backwards compatability

elyxlz:
Renamed the edit notification's current-text field from `new_text` to `text` and kept `old_text`, so a content rule using the `text` field/alias now matches an edited message the same way it matches a plain message. Also extended the `text` alias in the interrupt policy to reach a literal `text` field (it previously only expanded to body/message/content), and `text` is already a rendering content field so edit notifications keep a non-empty body. Updated the Go tests to assert on `text`/`old_text`, added a Python test for text-alias matching an edit, and updated the SKILL.md field reference. Go tests could not be run here (no Go toolchain); Python notification tests pass.

🤖 Generated with Claude Code

emipasca:
actually, ignore that, lets keep the final text as part of the notifaction body and only hve the old text field

elyxlz:
Reworked per the latest note: an edit notification now puts the new text in the `message` body field a plain telegram message uses, keeping only `old_text` as an extra, so an edit reads as a normal message and a content rule matches it naturally. Dropped the dedicated `text` facet field and reverted the earlier `_BODY_FIELDS` text-alias extension it required. Updated the Go and Python tests (edit body via `message`, `old_text` for the prior content) and the SKILL.md. Python `test_notification_rules.py` + `test_notifications.py` pass and `./check.sh guards` is green; the Go tests were not run (no Go toolchain in this environment).

🤖 Generated with Claude Code

emipasca:
do this alwasy for the similar pr on the whatsapp notifciation

PR #1294, opened 2026-07-16
fix(whatsapp): make the CLI's own help tell the truth

emipasca:
Both halves of this were confirmed against a real agent's logs: Luna tried to place a WhatsApp call on 0.1.177 and had to reverse-engineer the commands. Her transcript is quoted below.

### 1. `whatsapp --help` hid 15 of the 35 commands

`printUsage` was a hand-maintained list that drifted. Missing entirely: the **whole voice-calling feature** (`call`, `say`, `hangup`, `call-status`), plus `pair-phone`, `send-audio`, `revoke-message`, `archive-all-chats`, `get-group-invite-link`, `set-group-photo`, `list-received-contacts`, `daemon-status`, `link-start/status/stop`. It also advertised `search-contacts` (an alias, not a command) and listed `update-group-participants` and `serve` twice.

#1149 added the call commands to the registry and to SKILL.md, but nobody updated the usage text. Now it is generated from the registry, with aliases and positionals, so a command cannot ship without appearing.

### 2. `whatsapp <command> --help` never worked

Commands run in the daemon, so `--help` arrived as just another flag: the flag package printed usage to the **daemon's** stderr and returned `flag.ErrHelp`, which the handler returned as its error.

```
# before
$ whatsapp call --help
{"error":"flag: help requested"}

# after
$ whatsapp call --help
{"usage":"Usage of call:\n  -to string\n    \tWho to call (contact name, phone, or JID)"}
```

This matters most for the five commands whose flags SKILL.md deliberately does not spell out, pointing at `--help` instead: `send-audio`, `create-group`, `set-group-photo`, `get-group-invite-link`, `update-group-participants`. For those the agent had no documented signature **and** no working way to discover one.

### The part that is not just a bad message

Seven handlers took `_ []string` and ignored their args, so `--help` never reached a parser and the command simply **ran**:

| command | what `--help` did |
|---|---|
| `clear-all-chats` | **wiped the local message database** |
| `archive-all-chats` | archived every chat |
| `hangup` | dropped the live call |
| `link-stop` | stopped an in-progress link |
| `call-status`, `daemon-status`, `link-status` | harmless (read-only) |

Asking `clear-all-chats` what it takes destroyed the message history. The same hole meant a typo'd flag was silently ignored and the command ran anyway.

## What it actually cost, from Luna's logs (Jul 15, 21:57)

```
21:57:05  whatsapp call --help | head -30 ; whatsapp --help | grep -iE 'call|say|hangup'
21:57:15  whatsapp | grep -iE 'call|say|hangup|voice'
21:57:49  whatsapp call            <- run it bare, read the error to learn the flag
21:58:20  whatsapp call --to +44...
21:59:47  whatsapp say --to +44... "hey! it actually works"   <- wrong, say has no --to
21:59:59  whatsapp say "hey! it actually works"               <- corrected
```

In her own words:

> *"the help output doesn't list any voice call commands like `call`, `say`, `hangup`, or `call-status`, yet `call-status` actually worked when I tried it earlier. This suggests the commands might exist but aren't shown in the usage documentation"*

> *"Syntax is `whatsapp call --to`"* — inferred from the bare command's error message, three minutes and six probes after starting.

She then assumed `say` mirrored `call` and took `--to`, which it does not, and sent a spoken line to nowhere before correcting. Against the new output, her exact grep now matches:

```
$ whatsapp --help | grep -iE 'call|say|hangup'
  call <to>
  say <text>
  hangup
  call-status
```

## How

- `parseFlags(fs, args)` turns `flag.ErrHelp` into a `helpRequest` carrying the usage the FlagSet already writes. All 27 flag-parsing handlers route through it (one line each; the idiom was already uniform).
- `parseNoFlags(name, args)` gives the seven argless handlers a FlagSet, so they answer `--help` and refuse unknown flags.
- `executeCommand`, the single dispatch seam, returns usage as the command's result. Asking what a command takes is a question, not a failure.
- Commands declaring no flags report `hangup takes no flags` rather than a bare `Usage of hangup:` header with nothing under it.
- `printUsage` is generated from the registry.

**Zero drift, both halves**: the per-command usage comes from the real `FlagSet`, the command list from the real registry. Neither can fall out of sync the way a hand-written signature does.

**SKILL.md is unchanged.** It has been promising `--help` all along; the code now keeps the promise. That preserves the deliberate token trade behind `perf(skills): trim defaults to shrink agent token usage` rather than silently reversing it.

A genuine parse error (`flag provided but not defined: -x`) still returns as an error, unchanged.

## Tests

`TestEveryCommandAnswersHelpRatherThanRunning` and `TestUsageListsEveryCommandTheCLIAccepts` both drive the **real registry**, so a new command cannot be added without inheriting either guarantee. The test client is deliberately built with a nil whatsmeow connection: any command that runs its body instead of answering fails loudly rather than quietly passing.

All three mutation-checked rather than assumed:

- reverting `parseFlags` to hand back `ErrHelp` fails the registry-wide help test on every command;
- reverting `clear-all-chats` to ignore its args fails `TestHelpOnClearAllChatsLeavesTheMessagesAlone` by panicking inside `DeleteChat`, which is itself the proof the body executed;
- omitting `call`/`say` from the usage list fails `TestUsageListsEveryCommandTheCLIAccepts`.

`./check.sh whatsapp` passes, run in the `vesta:local` image (this box has no Go toolchain).

## Not verified end to end

Driven through `executeCommand` and `printUsage` rather than a live daemon over its socket, so the JSON has not been observed coming back out of a real `whatsapp call --help` against a paired account.

## Note

Telegram's CLI is a separate implementation and was not examined for either hole. Worth a follow-up.

---

## Update: restructured after code review

A workflow-backed review at high effort found **10 confirmed findings** against the first cut of this PR, several of them regressions it introduced. All are fixed; the diagnosis was a single mistake in layering.

**Help is static text derived from a FlagSet, but the daemon was made to produce it**, so the answer had to travel back out through dispatch, the read-only gate, and the socket. That leaked in five directions:

| finding | what it did |
|---|---|
| `clear-all-chats help` | **still wiped the database.** Go's flag package reads a bare `help` as a positional, so Parse returned nil and the body ran. `isHelpArg` blesses that spelling, so the CLI taught the one form that still destroyed data. The tests only passed `--help`, so they could not catch it. |
| `whatsapp send Bob '-h'` | reported usage and **exited 0** — a regression this PR introduced. The agent read a silent no-op as a delivered message. Previously it failed loudly at exit 1. |
| `whatsapp link --help` | burned a rate-limited pairing attempt (ban risk); `serve --help` ran the daemon forever. Both dispatch outside the registry. |
| read-only + `--help` | refused help for all ~19 write commands: the gate treated the question as the action. |
| daemon down + `--help` | unavailable, exactly when an agent is most likely reaching for it. |

**The fix.** Every handler declares its FlagSet and parses *before* touching the client, so `main` reports a command's flags by running it **with no client at all**: no daemon, no socket, no command body, no read-only gate, and plain text instead of a JSON-escaped one-liner. Only the first argument is read as help, so an argument whose own text is `-h` stays an argument and fails loudly carrying the usage, as it did before.

That invariant is verified across the whole registry (`TestEveryCommandReportsItsFlagsWithoutAClient`): a handler that reached for the client before parsing would panic rather than pass quietly.

Two smaller findings also fixed: a rejected flag now answers with the flag list the FlagSet already wrote (it previously existed nowhere), and usage no longer offers `link-start`/`link-status`/`link-stop`/`daemon-status`, which the `link` and `daemon` wrappers own and which bypass their rate-limit and QR-page handling when called directly.

**One finding not acted on.** The review flagged `commandSignature` rendering flag-only commands (`send-audio`, `create-group`) as bare names. On comparison that is not a regression: the old hand-written list showed those bare too, and did not list `send-audio` at all. With per-command `--help` now working and the usage header pointing at it, the discovery path is complete.


emipasca:
fix ci

elyxlz:
CI triage: not a real defect, environmental flake.

Failing step: `test-integration` -> "Run integration tests". The step hit its 18-minute cap (`Timeout of 1080000ms hit`), was retried, and the retry was canceled (`The operation was canceled.`). The runner was struggling: "Install Rust" logged `rustup install failed after 3 attempts` and rust-cache errored, and every Docker-lifecycle test ran "for over 60 seconds". Failures were inconsistent across attempts (`rename::rename_basic` and `upstream::agent_attaches_to_the_upstream_through_the_mounted_repo` both passed on attempt 1 and failed on the retry), the signature of timeout-induced flakiness, not a defect.

None of the failing tests (backup, rename, upstream container lifecycle) touch the whatsapp Go CLI. `go-checks` (which does cover this PR) passes. No code change made. Re-ran the failed job.

🤖 Generated with Claude Code
Do NOT merge this PR.

emipasca:
fix ci

elyxlz:
Heads-up: this PR's `test-integration` red is not your code. It is a systemic, master-wide failure hitting many PRs and master itself.

The `test-integration` job exhausts the ~14 GB free disk on the ubuntu runner: the `backup`/`rename`/`upstream` tests export multi-GB agent containers into restic and re-import them 8-in-parallel, so docker export/import and the mounted-repo git fetch fail with `no space left on device`. Attempt 1 stalls until the 18-minute cap (`Timeout of 1080000ms hit`), attempt 2 fails fast. Because master is red, no per-PR change can make this job green.

Fix is in #1295 (free the runner disk before the suite). Once that lands on master, rebase or re-run this PR's CI and `test-integration` should pass.

🤖 Generated with Claude Code

elyxlz:
c should be fixed rebase, and then try and merge

elyxlz:
Brought the branch current with master (merge, no force-push) to pick up the #1295 disk-cleanup fix, so test-integration re-runs without the runner disk exhaustion.

The refreshed CI surfaced one real, unrelated failure: master's new `guards` convention check (from #1291) flagged a 10-line comment block in `whatsapp/cli/notifications.go` (cap is 8). That comment predates the guard and is byte-identical to master, so master itself trips it. Trimmed it to 7 lines with the meaning intact (comment-only, no Go behavior change); `./check.sh guards` is green locally.

Auto-merge (squash) is armed, so this lands once the now-fixed required checks pass.

🤖 Generated with Claude Code

PR #1313, opened 2026-07-17
feat(web): persist the dashboard/chat divider position

elyxlz:
Fixes #1312

## What

The dashboard/chat split reset to the same ratio on every launch. It now persists.

`DesktopPanelView` wires its `ResizablePanelGroup` to `useDefaultLayout` from `react-resizable-panels`, the library's purpose built API for exactly this: it reads the stored layout through `useSyncExternalStore` and hands back a `defaultLayout`, so restoring needs no effect of our own. Storage defaults to `localStorage`, keyed by the group id plus the panel id set.

Three details worth flagging:

- **Stable panel ids.** The panels now carry explicit `id`s (`dashboard`, `chat`). Without them the hook keys on generated `useId` values, which change per mount, so nothing would ever restore.
- **`panelIds` tracks `chatCollapsed`.** The chat panel is conditionally rendered; passing the rendered set means the collapsed and expanded states persist as separate layouts instead of fighting over one entry.
- **`onlySaveAfterUserInteractions`.** Only a real drag or keyboard resize is saved, so mount and imperative layout changes cannot quietly overwrite what the user chose.

The no-preference default moves from 70/30 to 33/67 (roughly one third dashboard, two thirds chat) per the issue's nice to have. Existing users see this default once, then their first drag sticks.

## Tests

None, deliberately, and I want to be explicit about why rather than ship a test that only looks like one.

I probed this before deciding: in jsdom both panels render `flex: 50 1 0px` whether or not a layout is stored. `ResizeObserver` is stubbed to a no-op in `vitest.setup.ts` and jsdom has no layout engine, so the library never applies `defaultSize` or `defaultLayout` at all. A DOM level restore assertion would pass against any wiring, including no wiring.

The remaining option was a hook level test driving `useDefaultLayout` directly, but that exercises the library (which ships its own tests) rather than this component's wiring, and would still pass if the component were unwired entirely. The persistence logic here is the library's; what is ours is the wiring, which is only observable with a real layout engine.

## Verification

`./check.sh web` passes: 0 eslint errors (the 56 warnings are pre-existing on master), prettier clean, tsc clean, 185 vitest tests green.

Manual check still worth doing on a real window: drag the divider, relaunch, confirm the ratio holds, and confirm collapsing and reopening chat keeps its own split.


emipasca:
fix the merging block

elyxlz:
The merge block was the branch-protection requirement that the branch be up to date with master (all checks were already green, no review required). Brought the branch current by merging origin/master (clean, no conflicts), verified ./check.sh web passes, and pushed. Fresh CI is now running on the up-to-date branch and squash auto-merge is armed, so it will land automatically once merge-gate-ci goes green.

🤖 Generated with Claude Code

PR #1318, opened 2026-07-17
fix(browser): bound the bidi request path with a timeout

elyxlz:
Refs #1305

## Scope

This lands **fix 1 only** from the issue: a timeout in the bidi/send path, so a withheld BiDi response surfaces as a diagnosable error instead of an unbounded hang. It is correct regardless of why the browser goes quiet.

It deliberately does **not** attempt fix 2, the `open`/`new_tab` root cause. The issue is explicit that the root cause is not established: it lives in Camoufox's compiled Firefox fork, and the correlation with the patched window-open path is unverified. Guessing at a mechanism there would be inventing one.

So this closes the unbounded-hang class. `browsingContext.create` will now fail in about 60s with a message naming the method, rather than blocking forever. **The `open`/`new_tab` root cause stays open**, which is why the issue is referenced, not closed. `browser navigate` remains the working path for navigation.

## What changed

Both waits in the request path were unbounded, in two different processes:

- `bidi.py`: `send()` awaited its response future with no bound. It now awaits under a named `BIDI_RESPONSE_TIMEOUT_S` (60s) and raises `BidiError("timeout", "no response to 'browsingContext.create' within 60.0s")`. Raising `BidiError` rather than a bare `TimeoutError` is load-bearing: a bare `TimeoutError` is an `OSError` subclass whose `str()` is empty, so it would have surfaced through the daemon as `{"error": ""}`. `BidiError` instead rides the daemon's existing `except BidiError` channel and reaches the CLI with the method named.
- `admin.py`: `send()` built a socket and looped on `s.recv()` with no `settimeout`, the one function in that file that did not. It now calls `s.settimeout(DAEMON_RESPONSE_TIMEOUT_S)`, matching the pattern its neighbours already use, and converts the timeout into a `RuntimeError` naming the request.

The abandoned future is dropped from the pending map in a `finally`, so a wedged browser cannot accumulate them.

### On the two timeout values

60s / 90s, generous by design. `goto()` issues `browsingContext.navigate` with `wait="complete"`, so one legitimate command can cover a full page load; the skill's own `wait_for_load` default is 15s and the CDP backend's internal load bound is 30s, so 60s leaves real headroom and only fires when the browser has genuinely gone silent.

The socket bound sits **above** the daemon's own per-command bound on purpose. The daemon's error names the method, so letting it win the race surfaces the better message; a CLI-side timeout firing first would mask it with a blunter one.

## Tests

`tests/fake_bidi.py` gains a `withhold` set: methods the fake server accepts and never answers, reproducing exactly what Camoufox does to `browsingContext.create`. That keeps the high-fidelity fake faithful instead of reaching for a mock.

- a withheld response raises `BidiError` with code `timeout`, naming the method
- the timed-out request is dropped from the pending map
- one silent command does not wedge later commands on the same client
- `admin.send` surfaces a `RuntimeError` when the daemon accepts and stays quiet

Verified these are real red/green tests, not vacuous: reverting only the bounding logic while keeping the constants makes all four fail. Each test carries its own hang guard (an outer `wait_for`; a thread join for the socket path), so that regression fails in seconds with a clear message rather than hanging the CI job, which is what the unbounded code does today.

## Checks

`./check.sh agent` and `./check.sh guards` both pass.

### DIFF
--- a/agent/skills/browser/cli/src/vesta_browser/admin.py
+++ b/agent/skills/browser/cli/src/vesta_browser/admin.py
@@ -18,6 +18,9 @@ from .launcher import RunningCamoufox, launch
 SESSION_FILE_PREFIX = "/tmp/vesta-browser-"
 GRACEFUL_EXIT_POLLS = 25
 GRACEFUL_POLL_INTERVAL_S = 0.2
+# Above the daemon's own per-command bound, so a withheld browser response surfaces as the
+# daemon's error naming the method rather than as a blunter socket timeout out here.
+DAEMON_RESPONSE_TIMEOUT_S = 90.0
 
 
 def _session_name(name: str | None = None) -> str:
@@ -241,18 +244,32 @@ def ensure_daemon(wait_s: float = 30.0, name: str | None = None) -> None:
     raise RuntimeError(f"daemon {session!r} did not come up within {wait_s}s. Last log line: {tail or '(none)'}")
 
 
+def _req_label(req: dict) -> str:
+    """Name a request for error messages: its BiDi method, or its control meta."""
+    if "method" in req:
+        return repr(req["method"])
+    if "meta" in req:
+        return f"meta {req['meta']!r}"
+    return "request"
+
+
 def send(req: dict, name: str | None = None) -> dict:
     """Low-level sync request to the daemon. Raises on error."""
     s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-    s.connect(socket_path(name))
-    s.sendall((json.dumps(req) + "\n").encode())
-    data = b""
-    while not data.endswith(b"\n"):
-        chunk = s.recv(1 << 20)
-        if not chunk:
-            break
-        data += chunk
-    s.close()
+    s.settimeout(DAEMON_RESPONSE_TIMEOUT_S)
+    try:
+        s.connect(socket_path(name))
+        s.sendall((json.dumps(req) + "\n").encode())
+        data = b""
+        while not data.endswith(b"\n"):
+            chunk = s.recv(1 << 20)
+            if not chunk:
+                break
+            data += chunk
+    except TimeoutError:
+        raise RuntimeError(f"daemon did not respond to {_req_label(req)} within {DAEMON_RESPONSE_TIMEOUT_S}s") from None
+    finally:
+        s.close()
     resp = json.loads(data)
     if "error" in resp:
         raise RuntimeError(resp["error"])
--- a/agent/skills/browser/cli/src/vesta_browser/admin.py
+++ b/agent/skills/browser/cli/src/vesta_browser/admin.py
@@ -18,8 +18,7 @@ from .launcher import RunningCamoufox, launch
 SESSION_FILE_PREFIX = "/tmp/vesta-browser-"
 GRACEFUL_EXIT_POLLS = 25
 GRACEFUL_POLL_INTERVAL_S = 0.2
-# Above the daemon's own per-command bound, so a withheld browser response surfaces as the
-# daemon's error naming the method rather than as a blunter socket timeout out here.
+# Above the daemon's own per-command bound, so its error (which names the method) wins the race.
 DAEMON_RESPONSE_TIMEOUT_S = 90.0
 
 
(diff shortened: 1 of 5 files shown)


elyxlz:
code review

elyxlz:
## Code review: bound the BiDi request path with a timeout

**Verdict: approve-with-nits.** The fix is correct, well-scoped, and unusually well-tested. Timeout sits at the right layer, raises a typed error, cleans up the pending future, and leaves the connection usable. Two small considerations below, none blocking.

### What the change does (confirmed)
- `bidi.py:send` now wraps every BiDi request in `asyncio.wait_for(future, BIDI_RESPONSE_TIMEOUT_S=60.0)`; on timeout it raises `BidiError("timeout", ...)` naming the method and, in `finally`, drops the abandoned future from `_pending`. Applied at the single request/response chokepoint, so every BiDi command inherits it.
- `admin.py:send` bounds the client to daemon unix-socket read at `DAEMON_RESPONSE_TIMEOUT_S=90.0` (above the 60s bound so the daemon's method-naming error wins the race, per the comment).
- Tests: hung request times out naming the method, the timed-out future is dropped from `_pending`, a later command still succeeds on the same client, and `admin.send` surfaces an error instead of hanging. Ran `pytest tests/test_bidi.py tests/test_admin.py`: 28 passed.

### Correctness (good)
- Named constants, no magic numbers. `BidiError(code="timeout")` is typed, not stringly. The comment on the `raise` explains why it maps to `BidiError` not `TimeoutError` (the latter stringifies empty and the daemon would relay `{"error": ""}`), which is a genuine non-obvious mechanic worth the line.
- Cleanup is correct: `finally: self._pending.pop(command_id, None)` runs on success (reader already popped, hence the default), timeout, and cancellation. No unshielded await in `finally`. The reader's late `_resolve` no-ops because the id is gone from `_pending`. Connection is not poisoned: `test_a_withheld_response_does_not_block_later_commands` proves the next command works.
- The two bounds are complementary, not a redundant guard on one root cause: they cover different sockets (daemon to browser websocket vs client to daemon unix socket), so a wedged daemon is caught even when the browser is fine. One-fix-per-root-cause holds.

### Should-consider (nit)
- **60s caps real navigations, not only wedged commands.** `helpers.py:112` sends `browsingContext.navigate` with `wait="complete"`, so the command response only returns once the page finishes loading. A genuinely slow-but-valid page load beyond 60s would now fail with a `timeout` BidiError. 60s is generous versus industry defaults (Playwright's is 30s), so this is almost certainly fine, but the bound is uniform across all commands including the one long-legit operation. Worth a conscious "60s is enough for a slow page" sign-off rather than an accident of picking one number.

### Nit
- `admin.py` uses `s.settimeout`, which is per-operation (each `recv` gets a fresh 90s), not a total deadline; a daemon dribbling bytes slowly could exceed 90s overall. Not realistic here, noting for completeness.

### Conventions (clean)
- No `getattr`/`.get()`/`hasattr`/`Any`. Label extraction uses an `in` check (`req["method"] if "method" in req else req["meta"]`). `_pending.pop(command_id, None)` is not the banned `.get()` fallback and the default is load-bearing on the success path. Comments are one line each, under the cap; no lint escapes. Brand voice clean (no she/her/it, no dash separators).

🤖 Generated with Claude Code

elyxlz:
address fixes simply and merge

elyxlz:
Addressed the one review consideration (uniform 60s BiDi bound killing a slow-but-valid navigation) and armed auto-merge.

**The fix (minimal):** navigate and reload send `wait="complete"`, so their response only lands once the page finishes loading, which is exactly the operation the flat 60s per-command bound could kill. `bidi.send` now selects the timeout by method: the two wait-for-load methods (`browsingContext.navigate`, `browsingContext.reload`) get `NAVIGATE_RESPONSE_TIMEOUT_S = 90.0`; every other command keeps `BIDI_RESPONSE_TIMEOUT_S = 60.0`. The decision stays entirely inside `send` (the owner of the timeout), no per-call plumbing through the daemon, no redundant guard.

To preserve the ordering invariant (the daemon's method-naming error must win the race against the client socket read), `DAEMON_RESPONSE_TIMEOUT_S` is lifted 90.0 -> 120.0 so it still sits above the largest per-command bound.

**Test:** added `test_navigate_gets_the_longer_wait_for_load_bound`, which withholds the navigate response and asserts the error names the navigate bound, not the generic one (proves navigate took the longer path). All browser CLI tests pass (`pytest tests/test_bidi.py tests/test_admin.py`: 29 passed), guards pass, and `./check.sh agent` is green (a one-off `test_register_service` port flake cleared on rerun; it shares no code path with this change).

The minor `admin.py` per-operation-vs-total-deadline note was left as-is: it was a non-defect observation and any change there would be speculative.

**Merge status:** squash auto-merge armed; it lands once `merge-gate-ci` goes green.

🤖 Generated with Claude Code

PR #1320, opened 2026-07-17
fix(vestad): stop exporting UV_PROJECT_ENVIRONMENT so skill uv runs cannot clobber the engine venv

elyxlz:
Fixes #1310

## Root cause: `UV_PROJECT_ENVIRONMENT`, not `VIRTUAL_ENV`

The issue fingers `VIRTUAL_ENV`. It is a red herring, and removing it would have fixed nothing. Verified both against uv 0.11.3:

* `VIRTUAL_ENV` is **ignored** by uv for project commands. It warns `does not match the project environment path .venv and will be ignored` and the skill correctly gets its own `.venv`. Nothing in this repo ever exported it: it appears in the agent's env because `uv run` sets it for its child, which is why it was visible on the box.
* `UV_PROJECT_ENVIRONMENT=/root/agent/.venv` is the destroyer. It retargets **every** uv project, so `uv run` in a skill's `cli/` adopts the engine venv and re-syncs it, deleting and recreating it whenever the skill pins another Python. It was exported in two places: `ENV` in `vestad/Dockerfile` and `export` in `agent_container_entrypoint_cmd`.

Reproduced in a container with a read-only core mount, old entrypoint:

```
engine venv: version_info = 3.14
-- agent runs the skill's tests (the encouraged action) --
Removed virtual environment at: /root/agent/.venv
engine venv now: version_info = 3.11
-- next restart: --
ModuleNotFoundError: No module named 'rich'      # stands in for claude_agent_sdk
```

## Why the export existed (verified, not assumed)

Removing it outright would have made the bricking worse. `/root/agent/core` is a **read-only** bind mount, so without the variable uv defaults the engine's project env to `/root/agent/core/.venv` and dies:

```
error: failed to create directory `.../engine/.venv`: Permission denied (os error 13)
```

So the variable is genuinely needed, but only by the engine's own uv commands. It was exported globally because ad hoc `uv run --project core <tool>` also had to resolve the engine venv.

## The fix

* Scope `UV_PROJECT_ENVIRONMENT` to the engine's own `uv sync` (entrypoint + Dockerfile `RUN`), never exported.
* Put `/root/agent/.venv/bin` first on PATH. `uv run` was already doing this implicitly; making it explicit is what replaces the global export. `ruff`, `pytest`, `ty` and `python` are engine deps, so they still resolve as plain commands, and PATH does **not** make uv adopt the venv as its target (verified).
* Launch the engine with `.venv/bin/python -m core.main` rather than `uv run`, which would carry the variable straight back into the agent's env and every shell it spawns. This also collapses the legacy layout branch, since both layouts put the venv at the same path.

Same scenario, new entrypoint:

```
agent env: UV_PROJECT_ENVIRONMENT=[] VIRTUAL_ENV=[]
Creating virtual environment at: .venv     # the skill's own
engine venv now: version_info = 3.14       # untouched
python -> /root/agent/.venv/bin/python     # pinned tools still resolve
-- next restart: --
ENGINE BOOTED OK
```

## Existing boxes: no migration needed

The export lives only in vestad-owned places, never on the agent's disk. `/root/.bashrc` is truncated by the image and nothing writes the variable there, so there is no persisted per-agent state to converge:

* The entrypoint is the container's `cmd`, and `container_needs_rebuild` compares it against `agent_container_entrypoint_cmd()`. This change drifts every existing container, so reconcile snapshots and recreates each one.
* A rebuild goes through `docker export | docker import`, which drops image `ENV`, and `create_container` supplies no `env`. Verified directly: the imported container's `UV_PROJECT_ENVIRONMENT` is empty and PATH resets to the daemon default. That is also why the entrypoint re-exports PATH. So no box can inherit a stale value from its old image, and no `unset` guard is warranted.
* Already-damaged boxes repair themselves: the boot `uv sync --frozen --project core` restores a clobbered venv (verified), which is the same command the issue documents as recovery.

## Also updated

Two skill docs stated commands that this change would otherwise break or make false:

* `upstream-pr`: `uv run --project core ruff` would now hard-fail against the read-only mount, so it becomes plain `ruff` from `~/agent` (still the pinned ruff, via PATH). Adds the terse hazard plus recovery where it tells agents to run a skill's tests, per the issue's mitigation (3).
* `skills-registry`: cited `VIRTUAL_ENV=~/agent/.venv` as the reason for its `uv pip install -e` warning. The advice stands; the mechanism is now PATH.

Fix (2) from the issue is deliberately skipped: per-skill venvs are what (1) produces for free, and `.venv/` is already gitignored at any depth, so no box gets a dirty tree.

## Checks

`./check.sh vestad` (207 passed), `./check.sh agent`, `./check.sh guards` all green. Adds `entrypoint_never_exports_uv_project_environment_into_the_agent` as the regression guard.

### DIFF
--- a/agent/skills/skills-registry/SKILL.md
+++ b/agent/skills/skills-registry/SKILL.md
@@ -38,12 +38,11 @@ your edits and upstream updates take effect on the command's next run:
 uv tool install --editable ~/agent/skills/<name>/cli
 ```
 
-Never install a skill's CLI with `uv pip install -e` or `pip install -e`. You run
-with the engine venv active (`VIRTUAL_ENV=~/agent/.venv`), so an editable pip
-install drops the skill's command into `~/agent/.venv/bin`, which sits ahead of
-`~/.local/bin` on PATH and shadows the real tool: the command and its daemon then
-break with an import error. `uv tool install` keeps the CLI isolated and on
-`~/.local/bin`, so use it every time.
+Never install a skill's CLI with `uv pip install -e` or `pip install -e`. Run from
+`~/agent` those resolve the engine venv, dropping the skill's command into
+`~/agent/.venv/bin`, which leads your PATH and shadows the real tool: the command
+and its daemon then break with an import error. `uv tool install` keeps the CLI
+isolated and on `~/.local/bin`, so use it every time.
 
 ## Notes
 
--- a/agent/skills/upstream-pr/SKILL.md
+++ b/agent/skills/upstream-pr/SKILL.md
@@ -113,14 +113,7 @@ upstream-pr --token-only
 
 ## Running a skill's tests
 
-Each skill CLI is its own uv project, so run its tests from its own directory: `cd ~/agent/skills/<name>/cli && uv run pytest`. uv builds a `.venv` there (gitignored) and leaves the engine venv at `~/agent/.venv` alone.
-
-If uv ever prints `Removed virtual environment at: /root/agent/.venv`, it retargeted the engine venv instead (something exported `UV_PROJECT_ENVIRONMENT`). The running agent survives on already-imported modules, so nothing looks wrong until the next restart fails. Repair it before you restart, and re-run the tests with `uv run --isolated`:
-
-```bash
-UV_PROJECT_ENVIRONMENT=~/agent/.venv uv sync --frozen --project ~/agent/core
-cd ~/agent && .venv/bin/python3 -c "import core.main"   # verify the restart path
-```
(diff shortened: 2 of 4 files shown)


elyxlz:
less loc less comments less fluff less prompt please

elyxlz:
Tightened. Cut the new upstream-pr "Running a skill's tests" section from ~10 lines to 1 (dropped the UV_PROJECT_ENVIRONMENT repair recipe, which the fix makes moot), and trimmed every comment across the entrypoint (PATH, sync step, launch, the new test) plus the Dockerfile down to the irreducible WHY. Net roughly 12 fewer lines, no repeated paragraph-length explanations. The fix behaves identically: UV_PROJECT_ENVIRONMENT stays scoped to the core sync and is never exported, the engine venv leads PATH, and both entrypoint tests are unchanged and green (clippy + tests pass).

🤖 Generated with Claude Code

PR #1321, opened 2026-07-17
fix(microsoft): only advance the monitor watermark across successfully polled windows

elyxlz:
Fixes #1311

## The bug

`monitor.py` wrote `new_check_time` as the last-check watermark at the end of **every** cycle, including cycles whose polls fetched nothing because auth was broken. The poll helpers swallowed their own exceptions (`except Exception as e: logger.error(...)`), so a failed poll never reached the outer handler: the loop committed the watermark as if the window had been read, and the next cycle's `receivedDateTime gt <advanced watermark>` skipped the unread window forever. The daemon logged `Completed check cycle` throughout. On a real box this silently dropped two emails that needed a reply during a 26h token outage.

The `first_run` catch-up did not help: it only fires against a stale state file, i.e. only when the process was down. While the process is alive and auth is broken, the watermark is diligently advanced the whole time.

## The fix

The invariant: a watermark only advances across a window that was actually read. Poll helpers now return whether they read their window, and the loop records `new_check_time` only for the polls that succeeded; a failed poll parks its watermark so the next healthy cycle re-reads the window.

## Granularity: per (account, source), not one global watermark

Each poll unit (`mail:<account>`, `calendar:<account>`, `teams:<account>`, `channels:<account>`) carries its own watermark.

A single global watermark is simpler, but `notifications.write_notification` does **not** dedupe by message id, so re-reading is not free: with one global watermark, one broken account parks the watermark for everyone, and every healthy account re-reads and re-notifies its window every 45s until the broken one heals. That trades silent mail loss for a notification storm. Per-account keeps a dead account's window parked without touching a healthy account's.

Mail and calendar are separate units within an account for the same reason: they are separate endpoints with separate failure modes, and a permanently failing calendar (a missing `Calendars.Read` scope 403s forever) would otherwise re-notify that account's mail on every cycle.

Two deliberate non-failures, both preserving documented behavior:
- Teams **channel enumeration** failure keeps reporting the window read. It is the documented graceful degrade for the many accounts with no channel access, and parking there would only flood them if access ever appeared. A missing token, which loses messages the account can otherwise read, does report the window unread.
- A single team/channel failure after successful enumeration stays best-effort ("never fatal"), since a private channel the user cannot read would otherwise park the whole unit permanently.

## Bounded recovery

A unit's re-read window is clamped to `_MAX_CATCHUP` (7 days), so a long-dead account cannot flood the user when it heals. The clamp is persisted on each failed poll, so staleness cannot grow without bound. The issue's 26h outage recovers in full.

Catch-up is now intrinsic: `catching_up` is derived per unit from its own gap (`_CATCHUP_GAP_SECONDS`), which deletes the `first_run` flag and means recovered mail is tagged `missed` whether the process was down **or** the token was dead. The old flag could only ever see the former.

## Fleet impact

`state.txt` becomes `{"last_cycle": iso, "units": {...}}`. A box on the old format holds a bare ISO timestamp, which reads as `last_cycle` with no units, so every unit resumes from exactly where the old monitor left off: no migration, no re-notify burst on update, and the offline catch-up still works. The path is unchanged.

## Tests

`skills/microsoft/cli/tests/test_monitor.py`, driving the real `run()` loop over a real state file:
- a cycle whose poll failed does not advance the watermark, and the next successful cycle re-reads the window and produces the notification that was previously dropped (the issue's exact case)
- a broken account does not make a healthy one re-notify, and the healthy account's watermark keeps advancing
- recovery reads at most the catch-up window, and the recovered mail is tagged `missed`
- a legacy bare-timestamp state file is honored as the starting watermark
- the mail and calendar polls each report a window they could not read

Both watermark tests were confirmed red against the old unconditional-advance behavior.

`./check.sh agent` and `./check.sh guards` pass.

### DIFF
--- a/agent/skills/microsoft/cli/src/microsoft_cli/monitor.py
+++ b/agent/skills/microsoft/cli/src/microsoft_cli/monitor.py
@@ -1,7 +1,9 @@
 import json
 import re
 import time
+from collections.abc import Callable
 from datetime import UTC, datetime, timedelta
+from functools import partial
 from pathlib import Path
 from typing import NotRequired, TypedDict
 from zoneinfo import ZoneInfo
@@ -57,13 +59,11 @@ def _read_state(path: Path, now: datetime) -> MonitorState:
         parsed = json.loads(raw)
     except json.JSONDecodeError:
         return MonitorState(last_cycle=raw, units={})
-    if not isinstance(parsed, dict) or "last_cycle" not in parsed or "units" not in parsed:
-        raise ValueError(f"Malformed monitor state in {path}: {raw[:100]}")
-    last_cycle = parsed["last_cycle"]
-    units = parsed["units"]
-    if not isinstance(last_cycle, str) or not isinstance(units, dict):
-        raise ValueError(f"Malformed monitor state in {path}: {raw[:100]}")
-    return MonitorState(last_cycle=last_cycle, units={str(unit): str(watermark) for unit, watermark in units.items()})
+    match parsed:
+        case {"last_cycle": str(last_cycle), "units": dict(units)}:
+            return MonitorState(last_cycle=last_cycle, units={str(unit): str(watermark) for unit, watermark in units.items()})
+        case _:
+            raise ValueError(f"Malformed monitor state in {path}: {raw[:100]}")
 
 
 def _write_state(path: Path, state: MonitorState) -> None:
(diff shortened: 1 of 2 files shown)


elyxlz:
LESS LOC

elyxlz:
Tightened, roughly 13 fewer lines in monitor.py (44 removed / 31 added), behavior and test coverage unchanged.

- Folded `_unit_window` + `_record_poll` into one `_poll_unit`: it computes the window a unit still owes and advances the watermark only across a window the poll actually read. Each unit in `run()` is now a single `_poll_unit(..., partial(poll_fn, ...))` call instead of the three-line unit/window/record dance (the two calendar polls took a param reorder so `(last_dt, catching_up)` trail for `partial`; the two test calls were updated to match).
- `_read_state` now parses via a `match` on the JSON shape, dropping the duplicated `isinstance`/`raise` pair.

All 210 microsoft CLI tests pass; ruff check + format clean. The watermark still only advances across successfully-polled windows, so a failed poll never skips unread items.

🤖 Generated with Claude Code

PR #1385, opened 2026-07-18
fix(browser): widen try so a cancel on ws.send clears the pending entry

elyxlz:
## Summary

Both browser transports registered the pending future *before* the websocket send, outside the `try` whose `finally` pops it:

```python
self._pending[command_id] = future
await self._ws.send(...)   # cancel here leaks
try:
    return await asyncio.wait_for(future, ...)
finally:
    self._pending.pop(command_id, None)
```

A cancellation delivered while awaiting `ws.send(...)` never entered the `try`, so the `finally` never ran and the entry stranded. `websockets` applies write backpressure, so `ws.send` can suspend when the send buffer fills, which is exactly when a cancel is most likely to land on it.

## Fix

Widen the `try` one line up in both `bidi.py` and `cdp_backend.py` so the send is inside it and cleanup is symmetric with registration. Registration stays *before* the send: moving it after would race the read loop, which drops any reply whose id is not yet in `_pending`. The two backends are kept mirrored, as #1318/#1342 deliberately keep them.

## Testing

Added `test_cancel_landing_on_the_send_is_dropped_from_pending` to both `test_bidi.py` and `test_cdp_backend.py`. Each drives `send()` against a websocket whose `send` never returns (backpressure), asserts the entry is registered, cancels the request while it is blocked on `ws.send`, and asserts no pending entry remains. Both tests fail against the unpatched code (the future strands as `{1: <Future pending>}`) and pass with the fix. Full `skills/browser/cli` suite green (202 passed); `./check.sh guards` green.

Fixes #1349

### DIFF
--- a/agent/core/uv.lock
+++ b/agent/core/uv.lock
@@ -229,20 +229,20 @@ wheels = [
 
 [[package]]
 name = "claude-agent-sdk"
-version = "0.2.118"
+version = "0.2.122"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "anyio" },
     { name = "mcp" },
     { name = "sniffio" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/1b/bb/86d5dd732721ba96ffb004c0c573689d52dbb7829415cdbc9837a0bcb6b2/claude_agent_sdk-0.2.118.tar.gz", hash = "sha256:4d7f6aacbc4caf9ee585c9b3f53b8ed2be6fe376cdfeb77dbfb37948e48dfdab", size = 268653, upload-time = "2026-07-14T06:48:40.334Z" }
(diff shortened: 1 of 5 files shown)


elyxlz:
fix conflicts then merge

elyxlz:
Resolved conflicts and armed squash auto-merge.

Only one file conflicted: `agent/skills/browser/cli/tests/test_cdp_backend.py`, where this PR and master's #1318 each appended a new test at the same spot. Kept both: this PR's `BlockingSendWs` + `test_cancel_landing_on_the_send_is_dropped_from_pending`, and master's `test_wedged_domain_enable_propagates_timeout_instead_of_being_swallowed`.

The source files auto-merged cleanly: the PR head was already based on a master carrying #1318's `asyncio.wait_for` + `finally` cleanup, so the widened try and the timeout/finally coexist as intended. `ws.send` now sits inside the try, and the pending entry is cleared exactly once on every path (send-cancel, timeout, normal) via the single `finally: self._pending.pop(command_id, None)`. No double-pop, no leak. The widened try is not redundant: without it a cancel landing on `ws.send` never enters the try, so the finally never runs and the entry strands, which the new cancel-on-send tests prove.

Checks: 37 browser BiDi + CDP tests pass (incl. both cancel-on-send tests and the wedged-domain timeout test); ruff check + format clean on the resolved file.

Merge: squash auto-merge armed, lands when required checks go green.

🤖 Generated with Claude Code

PR #1477, opened 2026-07-25
fix(vestad): defer scheduled backups while the agent is mid-task

elyxlz:
Fixes #1466

## Problem

Scheduled backups stop the container for the restic snapshot, restarting the agent even mid-task and destroying in-flight state (the reported incident: an SPID-authenticated browser session dropped mid-flow by a backup-triggered restart).

## Fix

Gate the auto-backup loop on agent idleness, using the live activity state vestad already taps from each agent's WS (`AgentStatusCache`, the same `"idle"`/`"thinking"` signal the app roster shows). No new protocol, no extra agent call: the busy-check is a read of the existing in-memory cache.

- Before taking a due scheduled backup, `backup::gate_scheduled_backup` (pure, unit-tested) decides: idle or unknown activity -> run; busy -> defer; busy past the cap -> run anyway.
- Deferral parameters: re-check every 5 minutes (`BUSY_RECHECK_INTERVAL_SECS`), hard cap of 6 hours total deferral (`MAX_BUSY_DEFERRAL_SECS`) so a permanently-busy agent still gets backed up. Deferrals and cap-expiry are logged.
- An agent with no activity entry (stopped, or its tap unreachable) is not mid-task, so the backup proceeds: the failure mode of the busy-check is designed out rather than special-cased.
- Manual and pre-restore backups are untouched and stay immediate.

## Tests

Six unit tests on the pure gate in `backup.rs` (idle, unknown activity, first busy sighting, under cap, cap expiry, busy-then-idle), in the fast `cargo test -p vestad --bins` tier. `./check.sh vestad` green locally.

### DIFF
--- a/vestad/src/agent_status.rs
+++ b/vestad/src/agent_status.rs
@@ -231,6 +231,13 @@ impl AgentStatusCache {
         self.activity_rx.clone()
     }
 
+    /// Snapshot of each tapped agent's live activity state ("idle"/"thinking"). Agents that are
+    /// stopped or unreachable have no entry. The auto-backup scheduler reads it to defer a
+    /// scheduled backup while an agent is mid-task.
+    pub fn activity(&self) -> HashMap<String, String> {
+        self.activity_rx.borrow().clone()
+    }
+
     /// Snapshot of each alive agent's IANA timezone, as last reported on its connect snapshot.
     pub fn timezones(&self) -> HashMap<String, String> {
         self.timezones_rx.borrow().clone()
--- a/vestad/src/backup.rs
+++ b/vestad/src/backup.rs
@@ -17,6 +17,42 @@ const MIN_DISK_SPACE_BYTES: u64 = 1_000_000_000; // 1 GB
 const DISK_SPACE_MARGIN_BYTES: u64 = 500_000_000; // 500 MB margin above container size
 pub const BACKUP_STOP_TIMEOUT_SECS: i32 = 30;
 pub const MIN_AGE_FOR_BACKUP_SECS: u64 = 6 * 3600;
+/// How soon the auto-backup loop re-checks a busy agent it deferred.
+pub const BUSY_RECHECK_INTERVAL_SECS: u64 = 5 * 60;
+/// Hard cap on total deferral: a permanently-busy agent is backed up anyway past this.
+pub const MAX_BUSY_DEFERRAL_SECS: u64 = 6 * 3600;
+
+/// Whether a scheduled backup should run now or wait for the agent to go idle.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ScheduledBackupGate {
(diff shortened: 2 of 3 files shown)


elyxlz (commented):
Reviewed the gate table, activity source, map hygiene, sleep loop, manual/pre-restore paths, and tests. The pure gate is correct and exhaustively tested, manual and pre-restore backups are untouched, and the deferral map is pruned against the live agent list each cycle. One finding survives adversarial checking:

**Deferral state leaks across the daily-window rollover (`vestad/src/serve.rs:2588`, `:2677`)**, severity: medium.

The `current_hour < target_hour` skip at serve.rs:2588 runs before the gate, so once a deferral spans midnight the loop stops re-checking it (every cycle from 00:00 until the next window just `continue`s), while the `busy_deferrals` entry survives untouched. For any user-configured `backup.hour >= 18` (cap of `MAX_BUSY_DEFERRAL_SECS` = 6h crosses midnight), a busy agent produces two defects from that one root:

1. **The cap is silently defeated and that day's daily is dropped.** Deferrals re-check 20:00-23:55, then the window closes before the 6h cap can fire; `has_daily_today` rolls over at midnight, so the deferred backup is never taken. The cap's contract ("a permanently-busy agent is backed up anyway past this") does not hold in this configuration.
2. **The stale epoch recreates incident #1466 the next day.** The surviving entry (epoch from day 1) makes the *first* busy sighting in day 2's window read as `now - since >= 6h` -> `RunDeferralExpired` -> immediate stop + backup of an agent that may be two minutes into a fresh task, which is exactly the mid-task restart this PR exists to prevent.

Default `hour = 4` never hits this (cap fires by ~10:00), so it needs a non-default evening hour, but the hour is a plain user setting and the failure is deterministic once configured.

Suggested fix: scope deferral state to a single daily window: clear `busy_deferrals` in the `current_hour < target_hour` branch (one line; fixes the stale-epoch misfire, and the next window starts a fresh honest deferral). If you also want the cap's guarantee to hold across midnight, let the loop keep re-checking while `busy_deferrals` is non-empty instead of clearing (deliberate design choice, either is defensible; dropping a daily silently is the one outcome to avoid). Since the bug lives in the untested `spawn_auto_backup_task` wiring, consider extracting the window+deferral decision into a pure helper next to `gate_scheduled_backup` so the fix carries a test.

Non-blocking notes, no change requested:
- In-memory `busy_deferrals` loss across a vestad restart is fine (vestad stops all agents on its own shutdown, so the in-flight work the epoch protected is already gone); a one-line comment saying so would save the next reader the analysis.
- Residual known gap in the activity source: a transient `GET /config` failure flips the agent to `Starting`, which tears down the tap and removes its activity entry (agent_status.rs:394), and a WS drop resets it to `"idle"` (agent_status.rs:544), so a busy agent can briefly read as backup-safe. Windows are seconds wide against a 5min-1h gate cadence and the None->Run default is the right trade-off for stopped agents; just noting it is not airtight.

PR #1479, opened 2026-07-25
feat(email): undo-send hold window for email-client and microsoft send verbs

elyxlz:
Addresses #1446 (CLI layer only; the native-client "Undo send" toast and the always-on config default are follow-up work).

## What

An opt-in Gmail-style undo-send window on the send verbs of both email skills:

- **`--hold <secs>`** on `email-client-send` (send / reply / forward) and on `microsoft email send / reply / forward`. Instead of dispatching, the send is fully resolved and validated up front (recipients, reply threading, attachments), persisted to a pending queue, and the command prints a cancel token.
- **Cancel / list**: `email-client pending list` + `email-client pending cancel <token>`, and `microsoft email list-pending` + `microsoft email cancel --token <t>`.
- **Dispatch** folds into each skill's existing daemon, no new process:
  - email-client: the poll daemon's supervisor tick (~10s) sends due entries (`poll_daemon.dispatch_held`).
  - microsoft: the serve daemon's monitor cycle (~45s) sends due entries (`monitor._dispatch_held_sends` -> `pending.dispatch_due`), replaying the stored backend choice through the existing Graph/OWA-REST router.
- Default unchanged: without `--hold` every send dispatches immediately.

## Design

- One small queue module per skill, mirroring how these skills already duplicate small shapes rather than sharing a library: `agent/skills/email-client/pending_sends.py` (dependency-free, like `daemon_lifecycle.py`) and `microsoft_cli/pending.py`.
- Queue entries are typed dataclasses persisted one JSON file per token, written atomically (tmp + rename): `$EMAIL_CLIENT_DIR/pending-sends/<token>.json` and `~/.microsoft/pending-sends/<token>.json`.
- email-client stores the fully composed message (resolved recipients, threading headers, base64 attachments), so dispatch is deliver + mark-answered + sent-sync with no re-compose; microsoft stores the resolved command payload (`MailDraft`, ids, pinned backend).
- **Race-free cancel**: dispatch claims an entry by atomically renaming it to `<token>.sending`; cancel unlinks `<token>.json`. Exactly one side can win.
- **No double-send**: a failed dispatch is parked as `<token>.failed` (error recorded inside), never retried, and raises an interrupting `send_failed` notification through the skill's existing notification writer. A `.sending` file left by a crash mid-SMTP is likewise kept, not retried.
- `EMAIL_DRAFT_ONLY` still refuses a held send at the verb (it is a transmit). `--hold` is mutually exclusive with `--draft`/`--dry-run`, and email-client refuses `--hold` when the poll daemon is not running (nothing would ever fire it).
- New skill files need no install migration: email-client modules are imported through the resolved symlink parent (existing mechanism), and the microsoft CLI ships as a package.

## Tests

- `agent/skills/microsoft/cli/tests/test_pending.py`: hold enqueues without touching either backend, up-front validation, draft-only guard, cancel prevents dispatch, dispatch after fire-at sends the exact stored payload per verb (faked send client), dispatch before fire-at leaves the entry, list-pending ordering, failed dispatch parking.
- `agent/skills/email-client/tests/test_pending_sends.py`: queue mechanics (round-trip incl. attachments, claim-due boundary, cancel-vs-claim), `--hold` CLI path (enqueue without SMTP, daemon-required, flag exclusivity), `send_held` delivery incl. attachment reconstruction, daemon dispatch success + failure notification.
- No sleeps anywhere: dispatch takes explicit `now` / entries carry explicit fire-at values.

Note: the microsoft daemon's dispatch granularity is its 45s monitor cycle, so a send can go out up to ~a cycle after the window elapses (documented in the skill reference). Skills are discovered from disk, so there is no index to regenerate.

### DIFF
--- a/agent/skills/email-client/SKILL.md
+++ b/agent/skills/email-client/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: email-client
-description: Personal email over IMAP/SMTP for any provider (Gmail, Outlook, Yahoo, iCloud, Fastmail, generic IMAP). Multi-account: read an inbox, send/reply/forward, save drafts, manage messages and folders, handle attachments, and get paged on new mail. Calendar over CalDAV with the same credential (Gmail, iCloud, Fastmail, any CalDAV server): list/create/update/delete/respond to events. Requires the poll daemon for notifications.
+description: Personal email over IMAP/SMTP for any provider (Gmail, Outlook, Yahoo, iCloud, Fastmail, generic IMAP). Multi-account: read an inbox, send/reply/forward (with an optional undo-send hold window), save drafts, manage messages and folders, handle attachments, and get paged on new mail. Calendar over CalDAV with the same credential (Gmail, iCloud, Fastmail, any CalDAV server): list/create/update/delete/respond to events. Requires the poll daemon for notifications.
 ---
 
 # Email Client
@@ -46,7 +46,7 @@ One-time per account, ~2 minutes. See [SETUP.md](SETUP.md) for install, every au
 Two binaries plus a daemon:
 
 - `email-client` - read (`list-folders`, `list`, `get`, `search`, `attachments`, `status`), manage messages (`mark`, `move`, `archive`, `delete`), manage folders (`folder create/rename/delete/subscribe`), and choose notify folders (`notify list/add/remove`)
-- `email-client-send` - outbound mail (send, reply, forward, save draft)
+- `email-client-send` - outbound mail (send, reply, forward, save draft, `--hold` undo window; cancel via `email-client pending`)
 - `poll_daemon.py` - watches each account's chosen folders (INBOX by default) and writes a notification per new message
 
 Omit `--account` on any command to use the default account from `accounts.json`. All commands accept `--account` and (where relevant) `--folder` (default `INBOX`).
@@ -170,6 +170,19 @@ email-client-send --account personal --forward-uid 999 --to recipient@example.co
 ```
 
 A draft does not contact SMTP and does not flag the original `\Answered` (nothing was sent). `--dry-run` previews the draft without writing it. The Drafts folder is auto-detected (see below).
+
+### Undo window (hold a send)
+
+Pass `--hold <secs>` on any send, reply, or forward to keep an undo window: the message is fully composed and validated now (recipients, threading, attachments), then held; the command prints a cancel token, and the poll daemon sends it once the window elapses uncancelled (so the daemon must be running). Not combinable with `--draft` or `--dry-run`.
+
+```bash
+email-client-send --account personal --to recipient@example.com --subject "Hi" --body "..." --hold 60
+email-client-send --account personal --reply-to-uid 12345 --body "thanks, will do" --hold 30
+email-client pending list                # held sends: token, recipient, subject, seconds until fire
+email-client pending cancel <token>     # abort a held send before it fires
+```
+
+Use it whenever a send feels irreversible-risky (fast back-and-forth drafting, changed recipients or figures): a mis-send becomes recoverable for the length of the window. A held send that fails at dispatch is kept on disk (never retried, no double-send risk) and raises a `send_failed` notification so you can recover it.
 
 ### Draft-only mode
 
--- a/agent/skills/email-client/imap_client.py
+++ b/agent/skills/email-client/imap_client.py
@@ -45,6 +45,7 @@ from imap_tools import AND, MailBox, MailMessageFlags
 # importable without adding a new symlink for each one.
 sys.path.insert(0, str(pathlib.Path(__file__).resolve().parent))
 import daemon_lifecycle
+import pending_sends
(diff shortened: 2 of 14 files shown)


elyxlz (commented):
Reviewed as a money-path change (real outbound mail): traced both send paths end to end, verified default no-hold behavior, payload parity, and the claim/cancel protocol, and ran both new suites (22 + 13 pass). The rename-claim design is sound and the at-most-once choice (park, never retry) is the right one. Four findings survive adversarial checking, ordered by severity.

**1. HIGH: a cancel racing dispatch at the read (not the rename) kills the poll daemon.**
- `agent/skills/email-client/pending_sends.py:110-111`: `claim_due` handles the rename race (line 117) but `load(path)` between the glob and the rename is unguarded. A `pending cancel` unlinking the file in that window raises `FileNotFoundError` out of `claim_due`; reproduced locally. In `poll_daemon.py:372` `dispatch_held` runs bare in the supervisor loop (the only try/except wraps `desired_workers()`, line 361), so the exception exits the loop and the whole daemon dies: all mail notification watching stops. The module docstring's "cancel and dispatch can never both win" holds for the rename but not for this read.
- Same window in `microsoft/cli/src/microsoft_cli/pending.py:98-100` (`claim_due`) and `:79` (`list_pending`), where monitor.py:692's broad except degrades it to an aborted 45s cycle, and in email-client `pending_sends.py:100` / `imap_client.py:820` (`pending list` tracebacks when the daemon claims an entry mid-listing).
- Fix: in both skills' `claim_due` and `list_pending`, treat a vanished file as already-cancelled/claimed (catch `FileNotFoundError` around `load`, skip). Separately worth considering: any other per-entry error also kills the email-client daemon through the same bare call (e.g. a version-skewed entry shape raising `TypeError` from `HeldSend(**data)`; daemons keep running across fleet updates), so guarding the `dispatch_held` call the way `_maybe_probe` guards itself (poll_daemon.py:275) would keep one bad queue file from taking down mail watching.

**2. MEDIUM: orphaned `.sending` entries are silent lost mail, and the claim-all shape widens the window.**
- `pending_sends.py:109-120` / `pending.py:96-113`: `claim_due` renames every due entry to `.sending` before the first send is attempted, so a daemon death while sending entry 1 orphans entries 2..n that were never even tried.
- Nothing ever surfaces an orphan: `pending list` / `list-pending` glob `*.json` only, no notification fires (`daemon_died` says nothing about held sends; SIGKILL writes nothing). The user believes the mail went out and no surface disagrees.
- The PR's "a `.sending` file left by a crash is kept, not retried" claim has no test in either suite.
- Fix: claim each entry immediately before its own send attempt (shrinks the orphan window to the genuinely ambiguous mid-send crash), and surface `.sending` / `.failed` leftovers in the pending list output so a lost send is at least discoverable. Add a test pinning that a pre-existing `.sending` file is never re-dispatched.

**3. MEDIUM: microsoft held payload is not "fully resolved": attachments are paths, read only at dispatch.**
- `cli.py:700-702` checks existence at hold time, but the entry stores `MailDraft.attachments` as path strings and `email.py` (`_attach_files` / `_split_attachments`) reads the files at dispatch. A file deleted during the window fails the send (parked + notified, acceptable), but a file modified or replaced during the window silently sends different bytes than what was validated, and this diverges from email-client, which snapshots base64 content into the entry (`smtp_send.py:_held_from_outbound`). PR body and `references/email.md` both say "fully resolved".
- Fix: snapshot attachment bytes into the entry (or copy into the queue dir) like email-client does, or at minimum correct the docs.

**4. LOW: microsoft `dispatch_due` builds `_send_fns` outside the per-entry try.**
- `pending.py:151`: an unknown/corrupt `command` raises `ValueError` after the entry is already claimed, so it is neither parked as `.failed` nor notified, leaves a `.sending` orphan, and skips the remaining due entries that cycle. Move the `_send_fns` call inside the try so it parks + notifies like any other dispatch failure.

Nits (no change requested if you disagree):
- `fail()` writes the `.failed` file non-atomically (`pending_sends.py:128`, `pending.py:121`) while `save()` documents tmp + rename; a crash mid-write leaves a torn `.failed` alongside the `.sending`.
- Dispatch parity: a held AUTO entry goes through `backend.run(AUTO, ...)` (`pending.py:153`), which on a Graph permission failure retries OWA REST even with no captured token, so the `send_failed` notification can carry the misleading REST error; the immediate path's `_route` (cli.py:609-611) calls Graph only in that case.

Verified clean: default no-hold path is byte-identical on both skills (hold=None short-circuits before any behavior change); email-client dispatch replicates the immediate path's post-send steps exactly (mark-answered + sent-sync, bcc-stripped sent copy); held vs immediate kwargs match field for field on all three microsoft verbs; draft-only refuses held sends (tested); daemon liveness is enforced on both (email-client's explicit check, microsoft's global `_require_daemon`); credentials are fetched fresh at dispatch on both paths; tests use explicit fire-at values, no sleeps.

PR #1482, opened 2026-07-25
fix(browser): repair Camoufox's search-config stub so search init cannot wedge launch

elyxlz:
Addresses #1445

## Root cause

The fatal `JSON error: missing field `recordType` at line 1 column 114` is not remote data, a corrupt profile, or the pin. Camoufox's own `no-search-engines.patch` hardcodes `SearchEngineSelector.#getConfiguration` (inside `omni.ja`) to return one stub record from the pre-v2 search-config era:

```js
if (true) {
  return [{ "appliesTo": [{ "default": "yes", "included": { "everywhere": true }, "webExtension": { "id": "none@mozilla.org" } }] }];
}
```

FF150+ feeds that JSON to the Rust search selector (`RustSearch.sys.mjs` uniffi bindings), whose serde parser requires `recordType`. `JSON.stringify({data: [stub]})` is exactly 116 chars; serde reports the missing field at column 114, matching the issue byte for byte. SearchService init then fails permanently and retries forever.

Two red herrings from the issue, checked and cleared:

- The "beta.25 installed vs alpha.26 pinned" mismatch is cosmetic: release tag `v150.0.2-beta.25` genuinely ships `camoufox-150.0.2-alpha.26-lin.x86_64.zip` (per-arch asset build numbers differ from the tag upstream). The pinned sha256 matches the live asset.
- A pin bump does not help: `v152.0.4-beta.28` (latest) carries the identical stub and fails identically (verified by launching it). The stub is unchanged on camoufox master, so no upstream build fixes this.

## Fix

`repair_search_stub` in `launcher.py` rewrites the stub inside `omni.ja` to a minimal valid search-config-v2 record set (one DuckDuckGo engine + a `defaultEngines` record). It runs on fresh extraction (before the atomic publish) and on every `ensure_camoufox` hit of an existing install, so fleet boxes that already extracted the broken build heal on their next launch with no migration. Unknown future layouts (member missing, stub gone) are a no-op, and the rewrite is atomic (`tmp` + `replace`).

## Verified in this environment (x86_64, headless, real builds)

- Unpatched v150 (pinned asset, sha-verified): reproduces the exact `recordType` error from the issue on a fresh profile.
- Unpatched v152.0.4-beta.28: identical error, confirming pin bumps are not a fix.
- Patched v150 and v152: zero `recordType` errors; the skill's real flow (launcher launch, BiDi `session.new`, adopt initial context, `browsingContext.navigate` to example.com, `captureScreenshot`) succeeds and the process stays alive.
- Full shipped path end-to-end: `ensure_camoufox()` with no override (real download of the pinned asset, sha verify, extract, repair) then launch + navigate: success.
- A profile that had already run the broken build picks up the repair (Firefox invalidates its startupCache when omni.ja changes); no cache clearing needed.
- Stock Firefox over the same BiDi flow works in this environment (control).

## Verified vs inferred

Verified: the search-component crash, its root cause, and that the repaired build launches and navigates through the skill's real code path. Not reproduced here: the reporter's hard process death ("Exiting due to channel error" killing the browser right after BiDi). In this environment the unpatched browser survives the search failure with the initial-context flow, so that death has an environmental component in the reporter's container; the search-init failure it follows is what this PR removes. Hence "Addresses", not "Fixes". One upstream deficiency noted while testing: Camoufox's BiDi `browsingContext.create` never answers (stock Firefox answers fine); the skill never calls it (it adopts the initial context), so nothing to change, recorded here for future debugging.

### DIFF
--- a/agent/skills/browser/cli/src/vesta_browser/launcher.py
+++ b/agent/skills/browser/cli/src/vesta_browser/launcher.py
@@ -125,6 +125,74 @@ def _extract_preserving_mode(zip_path: Path, dest: Path) -> None:
                 (dest / info.filename).chmod(mode)
 
 
+_OMNI_SEARCH_SELECTOR = "moz-src/toolkit/components/search/SearchEngineSelector.sys.mjs"
+# Camoufox's no-search-engines patch, verbatim from its shipped omni.ja (v150 and v152).
+_BROKEN_SEARCH_STUB = """\
+      return [
+        {
+          "appliesTo": [{
+            "default": "yes",
+            "included": {
+              "everywhere": true
+            },
+            "webExtension": {
+              "id": "none@mozilla.org"
+            }
+          }],
+        },
+      ];"""
+_REPAIRED_SEARCH_STUB = """\
+      return [
+        {
+          "recordType": "engine",
+          "identifier": "ddg",
+          "base": {
+            "classification": "general",
+            "name": "DuckDuckGo",
+            "urls": {
+              "search": {
+                "base": "https://duckduckgo.com/",
+                "searchTermParamName": "q"
+              }
+            }
+          },
+          "variants": [{ "environment": { "allRegionsAndLocales": true } }]
+        },
+        {
+          "recordType": "defaultEngines",
+          "globalDefault": "ddg",
+          "specificDefaults": []
+        },
+      ];"""
+
+
+def repair_search_stub(home: Path) -> None:
+    """Fix Camoufox's search-config stub inside omni.ja so SearchService can start.
+
+    Camoufox hardcodes SearchEngineSelector's configuration to one stub record that
+    predates search-config-v2; FF150+ hands it to the Rust selector, which requires
+    `recordType` and fails search init permanently (fatal in headless boxes: every
+    open then hangs, issue #1445). Rewriting the stub to a minimal valid config
+    heals fresh extractions and already-installed builds alike; Firefox invalidates
+    the profile startupCache on its own when omni.ja changes."""
+    omni = home / "omni.ja"
+    if not omni.is_file():
+        return
+    with zipfile.ZipFile(omni) as z:
+        if _OMNI_SEARCH_SELECTOR not in z.namelist():
+            return
+        source = z.read(_OMNI_SEARCH_SELECTOR).decode()
+    if _BROKEN_SEARCH_STUB not in source:
+        return
+    repaired = source.replace(_BROKEN_SEARCH_STUB, _REPAIRED_SEARCH_STUB)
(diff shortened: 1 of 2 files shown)


elyxlz (commented):
Reviewed the omni.ja repair for crash safety, idempotence, the LEGACY marker, the startupCache claim, fingerprint impact, test hermeticity, and the sha pins. One finding survives; everything else checks out (both arch pins match GitHub's own asset digests, the LEGACY condition is concrete and evaluable against `CAMOUFOX_RELEASE_TAG`, unknown layouts correctly fail open, tests are hermetic and pass).

- [ ] `agent/skills/browser/cli/src/vesta_browser/launcher.py:188` (medium): the heal-path rewrite uses one fixed tmp path, `omni.ja.tmp`, with no serialization. Two concurrent launches against a still-broken install (exactly the first-launch-after-upgrade moment this heal exists for, and the rewrite window is seconds: a full re-deflate of omni.ja) both pass the `_BROKEN_SEARCH_STUB in source` check and both open the same tmp path with mode "w": the second open truncates the inode the first is still writing through, so the two writers interleave on one file. The first finisher renames that shared inode over `omni.ja` (the second writer keeps writing into the live `omni.ja` through its fd, and its own `tmp.replace` then raises FileNotFoundError), so `omni.ja` can end up a corrupt zip. The failure is then permanent: every later launch raises `BadZipFile` at `zipfile.ZipFile(omni)` inside `repair_search_stub` before `ensure_camoufox` returns, and since `home/camoufox` still exists nothing ever re-extracts: a bricked install of the same wedged-launch class this PR removes. This function's siblings already treat concurrent launches as real (the staging keep-theirs publish at launcher.py:224, the flock in `_ensure_xvfb`), so the heal path should too. Fix at the root cause, no `BadZipFile` guard needed afterward: give each writer its own tmp (e.g. `omni.with_name(f"omni.ja.{os.getpid()}.tmp")`); the repaired content is identical across writers, so last-rename-wins is safe and `Path.replace` stays atomic. Worth also sweeping stale `omni.ja*.tmp` leftovers, since a crash mid-rewrite currently strands one forever (after repair the function no-ops before reaching the tmp path).

Verified clean, for the record: crash mid-rewrite before the rename self-heals (mode "w" truncates the partial tmp, `omni.ja` untouched, next run repairs again); repeated runs no-op byte-identically; a future build with a different stub falls through untouched (fail open is right here, the browser is no worse off than today); a running Firefox keeps its mmapped old inode across the rename on Linux; the injected DuckDuckGo search-config-v2 record is consumed only by SearchService internals via the hardcoded `#getConfiguration` (no remote-settings fetch, not content-observable) and the skill never drives the urlbar.

PR #1486, opened 2026-07-25
fix(agent): flip graceful_shutdown even when the crash-persist fails, so a dead processor never wedges the agent

elyxlz:
Fixes #1363

## Root cause

The incident log sequence pins it: `compact_result success`, `state → idle`, `Client session closed`, all in the same second, then the agent busy-skipping every 2s with no recovery until a manual container restart. The new regression test reproduces that exact log sequence on master.

The chain, restart=false only:

1. The restart=false drain tail is the one place that writes a live follow-up notification after compacting (`drain_compaction_request` -> `drop_core_notification` -> `atomic_write_text`). On a failing disk that write raises `OSError`, and nothing between the drain tail and the processor loop catches it (the drain's catch wraps only `compact_session`), so `message_processor` unwinds, logging `Client session closed` on its way out. That matches the incident timestamps: the only code between "state → idle" and the session close is that write.
2. `handle_processor_done`, whose whole job is "the agent never wedges silently", then runs `state_store.save_state` **before** `state.graceful_shutdown.set()`. On the same failing disk that persist raises too. An exception escaping a done-callback is swallowed by asyncio's callback handler, so the flip on the next line never happens.
3. With `graceful_shutdown` never set, `run_vesta` keeps waiting, nothing restarts the agent, and `monitor_loop` stays alive queueing turns nothing will ever consume. `not queue.empty()` keeps the proactive-check busy-skip firing every tick (the 20:02 start in the incident is the 60-minute proactive interval coming due), and the agent is unresponsive forever.

`restart=true` (the nightly dreamer) never touches this: its tail hands off to vestad, which restarts the container externally regardless of agent-side state, which is why the dreamer resumes cleanly every night while the in-place path wedged.

## Fix

One layer, at the invariant's owner: `handle_processor_done` now guards the persist so `graceful_shutdown.set()` always runs. A crash whose reason cannot be persisted still restarts the agent (the reason falls back to the generic crash label on the next boot); it no longer wedges it.

The drain-tail write failure itself keeps its existing semantics: an unexpected exception unwinds the processor and the agent crash-restarts via the callback, which is the designed recovery. No second defensive layer added in the drain.

## Secondary (busy-skip log spam)

Already fixed on master by #1252: the skip line is logged once per due window (`proactive_defer_logged`), not every tick, so this PR adds nothing there.

## Tests

- `test_inplace_nap_on_failing_disk_restarts_instead_of_wedging` (test_interrupts.py): drives `message_processor` end to end through a restart=false nap on a failing disk, with the done-callback attached exactly as `main.py` wires it, and asserts the crash flips `graceful_shutdown` with a `crash:` reason instead of wedging. Fails on master (times out waiting for the flip, after reproducing the incident's log sequence verbatim).
- `test_handle_processor_done_failing_state_write_still_triggers_restart` (test_processor.py): unit regression on the callback itself. Fails on master.
- `test_processor_serves_turns_after_inplace_nap` (test_interrupts.py): pins the healthy restart=false contract, the processor returns to serving turns on the same session after an in-place drain.

`./check.sh agent` and `./check.sh guards` green locally.

### DIFF
--- a/agent/core/main.py
+++ b/agent/core/main.py
@@ -60,7 +60,13 @@ def handle_processor_done(task: asyncio.Task[None], *, name: str, state: vm.Stat
         else:
             logger.error(f"{name} exited without error, restarting")
             state.persisted.last_restart_reason = f"crash: the {name} exited silently"
-    state_store.save_state(state.persisted, config)
+    # The flip is the wedge-proof invariant: the same failing disk that killed the task makes
+    # this persist raise, and an exception escaping this callback is swallowed by asyncio, so
+    # an unguarded persist leaves the agent up but never serving again (issue #1363).
+    try:
+        state_store.save_state(state.persisted, config)
+    except OSError as e:
+        logger.error(f"failed to persist restart reason: {e}")
     state.graceful_shutdown.set()
 
 
--- a/agent/tests/test_interrupts.py
+++ b/agent/tests/test_interrupts.py
@@ -602,6 +602,93 @@ async def test_compact_session_query_send_timeout_fails_compaction():
     assert state.turn is None
 
 
+# --- In-place (restart=false) compaction through the processor ---
+
+
+def _processor_nap_harness(tmp_path):
+    """Stream harness wired for driving message_processor through an in-place nap; the follow-up
+    notification write is pointed at the test's tmp dir, never the real agent home."""
+    state, config, mock_client, _, message_queue, _ = make_stream_harness()
+    config.agent_dir = tmp_path / "agent"
+    mock_client.__aenter__ = AsyncMock(return_value=mock_client)
+    mock_client.__aexit__ = AsyncMock(return_value=None)
+    return state, config, mock_client, message_queue
+
+
+async def _drive_nap(state, mock_client, message_queue, queue):
+    """One turn that requests an in-place nap, then the compaction stream that completes it."""
+    from claude_agent_sdk import SystemMessage
+
+    await queue.put(vm.QueuedTurn("please take a nap", True, []))
+    await wait_for_condition(lambda: mock_client.query.await_count >= 1, message="nap turn never queried")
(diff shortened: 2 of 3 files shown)


elyxlz (commented):
Review of the fix against issue #1363. Root cause chain re-derived and confirmed: the restart=false drain tail write (agent/core/loops.py:396, drop_core_notification -> cfg.atomic_write_text) has no catch between the idle-tick call site (loops.py:436) and message_processor, so an OSError unwinds the processor through client_session; master's unguarded save_state in the done-callback (before the flip) then swallows the flip when the same disk fails the persist. Verified the two regression tests pass on the branch and fail with the fix hunk reverse-applied. No second guard was added in the drain, which is right.

One finding:

- [ ] agent/core/main.py:66-70 (medium): the flip is guarded, not unconditional. `except OSError` makes graceful_shutdown.set() survive only the exception type this incident happened to throw; anything else escaping state_store.save_state (or a future edit to it) still skips the flip and re-wedges the agent, the exact failure mode this PR exists to remove. The code comment claims "the flip is the wedge-proof invariant", but the code enforces "the flip survives OSError". This callback is also the single backstop for every other unguarded save in the crash paths (e.g. the saves in _run_one_turn unwind into the processor and land here), so enumerating exception types in the last-resort supervisor repeats the assumption that caused the incident ("the persist cannot raise"). Make the invariant structural:

  ```python
  try:
      state_store.save_state(state.persisted, config)
  except OSError as e:
      logger.error(f"failed to persist restart reason: {e}")
  finally:
      state.graceful_shutdown.set()
  ```

  (or drop the except and let asyncio's callback exception handler log it; the `finally` is the load-bearing part). Same one layer, one root cause, just airtight. Worth extending the unit regression with a non-OSError side_effect so the structural guarantee is pinned.

PR #1491, opened 2026-07-25
fix(orb): go still for a waiting agent, and give agent status one owner

elyxlz:
Addresses #1135 (the design call stays with you; this PR is the cheap-to-reject concrete proposal).

## The pick: Option B, remap to `off`

Of the two forks in the issue, this implements Option B. Reasoning:

- **Subtraction wins.** Option A adds a sixth `OrbVisualState`, a new palette row in the generated design tokens (plus `scripts/sync-design-tokens.py`), a `getVisualTarget` branch, and touches favicon + Debug palette. Option B is a two-line remap.
- **Option A's distinction is invisible when it matters.** Amber-steady vs. amber-pulsing is motion-only; with `prefers-reduced-motion` (or `suppressMotion`) the orb suppresses pulse amplitudes entirely, so "actively reasoning" and "blocked on you" would render identically. A color carrying two meanings distinguished only by motion fails the glance test.
- **Still and dim is the honest signal.** These states never self-resolve; the calm busy blue said "working on it". An orb that visibly stops floating and glowing immediately breaks the false self-resolving read, which is the actual bug: the one glanceable status was training users to ignore it.

The label line carries the difference from a plainly stopped agent, using copy the app already speaks (the NotificationProvider waiting-state notifications say "needs to be set up" / "needs to sign in again", vestad's push says "needs you to sign in"):

| status | before | after |
| --- | --- | --- |
| `not_authenticated` | busy blue, floating, "needs to sign in again" | off (dim, still), "needs you to sign in" |
| `unprovisioned` | busy blue, floating, "not set up" | off (dim, still), "needs to be set up" |
| `stopped` | off, "stopped" | unchanged |
| starting/restarting/rebuilding | busy blue | unchanged, blue keeps meaning "in progress, no action needed" |

The mobile orb (`apps/mobile/src/components/AgentOrb.tsx`) had the identical false busy read for the same two statuses; the same remap is applied there so the product's face tells one story on both surfaces.

## The remap belongs to `@vesta/core`

Web and mobile each carried their own copy of the status to orb bucket mapping (`Orb/styles.ts`'s `resolveAgentStatus` and `AgentOrb.tsx`'s `orbColors`). That duplication is *how* the two surfaces drifted into telling different stories about the same agent, so fixing the mapping without moving it would only reset the clock on the next drift.

`@vesta/core` now owns the decision:

- `agentOrbState(status, activityState)` is an exhaustive `switch` over `AgentStatus`, so adding a status to `protocol/tree.ts` becomes a compile error in one place instead of a silent `off` on web and mobile independently.
- `orbIsLive(state)` is the animate-or-hold-still predicate, lifted out of web's private `LIVE_STATES`.

This also closed a gap in the first pass: the mobile orb's rotation loop ran for **every** status (only the breathe/pulse was gated on `alive`), so a waiting agent changed color but kept spinning. Mobile now gates rotation on `orbIsLive` too, and the orb actually stops.

## The labels were drift, not design

Web's orb line and mobile's `AgentStatusBadge` each carried their own words for the same `AgentStatus`, and they had drifted apart:

| status | web said | mobile said |
| --- | --- | --- |
| `alive` | "alive" | "online" |
| `starting` | "waking up..." | "starting" |
| `rebuilding` | "updating..." | "rebuilding" |
| `dead` | "broken, delete and recreate it in settings" | "offline" |
| `not_authenticated` | "needs you to sign in again" | "sign-in needed" |

Most of those are register differences, but `dead` was a genuine meaning conflict: "offline" reads like a recoverable network blip rather than a container that needs recreating.

`agentStatusLabel(status, activityState)` now joins `agentOrbState` in core, so the words and the visual bucket are decided together and once. The narration voice wins, because the actor-naming copy for the two waiting states is what the rest of the product already speaks. The final map:

| status | label |
| --- | --- |
| `alive` (idle / thinking) | "alive" / "thinking" |
| `starting` | "waking up..." |
| `setting_up` | "setting up..." |
| `restarting` | "restarting..." |
| `rebuilding` | "updating..." |
| `not_authenticated` | "needs you to sign in" |
| `unprovisioned` | "needs to be set up" |
| `stopped` | "stopped" |
| `dead` | "broken" |
| `not_found` | "unavailable" |

Mobile's badge now takes `activityState`, so a thinking agent reads the same there as it does under the orb. Web keeps its operation layer on top ("stopping...", "signing in...", "backing up..."), which has no mobile equivalent and is not a status label.

## Changes

- `apps/core/src/agent-status/agent-status.ts` (new): `OrbVisualState`, `agentOrbState`, `orbIsLive`, `agentStatusLabel`, exported from the package root. Named for agent status rather than the orb, since the badge that consumes the labels renders no orb.
- `apps/core/src/agent-status/agent-status.test.ts` (new): every `AgentStatus` pinned to its bucket and its label, the live/still split, and an invariant that no status leaks its raw enum into the UI.
- `apps/web/src/components/Orb/styles.ts`: `not_authenticated` and `unprovisioned` map to `orbState: "off"`. The file now delegates both bucket and label to core and keeps only the operation layer, and `AgentLike.status` tightens from `string` to `AgentStatus`, which retires the stringly `default` branch that used to render the raw enum for `not_found`.
- `apps/web/src/components/Orb/index.tsx`, `apps/web/src/hooks/use-orb-state.ts`: consume `orbIsLive` and the tightened type.
- `apps/mobile/src/components/AgentOrb.tsx`: palette via `designTokens.orb[agentOrbState(...)]`, rotation gated on `orbIsLive`; its local `orbColors` is gone.
- `apps/mobile/src/components/AgentStatus.tsx`: its private `STATUS_LABELS` record is gone, replaced by `agentStatusLabel`; the badge takes an optional `activityState`. `agent-identity-card.tsx` threads it through.
- `apps/web/src/components/Orb/styles.test.ts`: coverage for both waiting states (off + label), a contrast case pinning that "stopped" stays distinct, and the two operation-precedence cases.
- `vestad/src/docker.rs`, `agent_status.rs`, `sync/protocol.rs`, `sync/handler.rs`, `serve.rs`: the `AgentOperation` enum, its cache with set/clear, the wire field, roster threading, and the drop guard on both handlers.
- `apps/core/src/protocol/tree.ts`, `tree/roster.ts`: the `AgentOperation` type, the `AgentInfo` field, and the `rostersEqual` comparison.
- `apps/mobile/src/components/AgentHeader.tsx`, `AgentOrb.tsx`, `AgentStatus.tsx`, `agent-identity-card.tsx`: the operation threaded to every surface that renders an agent.
- `agentStatusKind` + the three derived predicates, replacing the hand-rolled sets in `ActionsCard`, `NotificationProvider`, `AgentNavbar`, `AgentMenu`, `Chat`, `AgentSocketProvider`, `web/src/api/agents.ts`, `mobile/src/agent/AgentProvider.tsx`, `mobile/src/api/endpoints.ts`, and `mobile/src/components/AgentStatus.tsx`.

No new `OrbVisualState`, no palette change.

## Backup and restore move to vestad

Chasing the labels turned up a third copy of the same problem, one layer down: `getOpLabel` and `resolveStatus` each spelled out the same six operation strings, and both render on the same agent card (the orb tooltip read one copy, the spinner caption below it read the other). `resolveStatus` now calls `getOpLabel` and maps only the orb state.

That raised the real question: why does web have an operation layer at all when mobile doesn't? Two different answers, and only one is a gap.

`starting`, `stopping`, and `authenticating` are genuinely client-local. They cover the gap between the POST and the delta reflecting it, guard against a double submit, and own the failure message. `AgentStatus` already describes them, so publishing them too would create a second representation of one truth.

`backing-up` and `restoring` are not. A backup pauses the container while restic works (`with_container_paused`), and a restore removes and recreates it, so the container state a client can read describes neither. Both also outlive the request that started them, since both handlers run detached in `spawn_pipeline_sse`. The result was that the work was visible only to the tab holding the SSE stream:

- another client saw a plainly idle agent, and
- reloading the page mid-backup dropped it from the UI entirely, because the store is in-memory zustand, while restic kept running.

`AgentInfo` now carries an `operation` the same way it already carries `buildPhase`, vestad's existing precedent for a long-running operation published as tree state. It is set by the backup and restore handlers through a `PublishedOperation` guard that clears on drop, so an early `?` return cannot strand an agent looking busy forever. This does not widen the read-only agent tap: like `buildPhase`, it is vestad's knowledge about its own work.

A running operation outranks the status wherever it is displayed, because a paused container otherwise reads as an agent the user has to restart. Web keeps its local operation for the request window, the mutex, and the error, but the words come from core, so the local request and the roster row cannot describe the same work differently. `isBusy` now also covers an operation another client started. Mobile shows both for the first time.

The wire change is purely additive, so `MIN_SUPPORTED_CLIENT_VERSION` does not move. `sync-protocol.json` is regenerated.

## One classifier, not ten hand-rolled status sets

The same problem ran one layer deeper, in code that gates behavior rather than pixels. Ten places re-derived the same status groupings, and each surface had named them differently:

- `not_authenticated || unprovisioned` in six web components (`ActionsCard`, `NotificationProvider`, `AgentNavbar`, `AgentMenu`, `Chat`, `AgentSocketProvider`) plus mobile's badge tint.
- The set an agent must reach before its socket answers: `ready` in web's poller, `connectable` in mobile's `AgentProvider`, and inline a third time in mobile's `endpoints.ts`.
- The give-up set `dead || stopped || not_found`, in both pollers and in `ActionsCard`'s `isRunning`.

Adding a status to the union would have been read as healthy by every one of them, silently, which is the same failure mode as the orb but gating whether a socket opens and whether a poller gives up.

`agentStatusKind` is now the only exhaustive switch over `AgentStatus` in the codebase:

| kind | statuses |
| --- | --- |
| `alive` | `alive` |
| `working` | `starting`, `setting_up`, `restarting`, `rebuilding` |
| `needs-user` | `not_authenticated`, `unprovisioned` |
| `down` | `stopped`, `dead`, `not_found` |

`agentNeedsUser`, `agentIsConnectable`, and `agentIsDown` derive from it, and `agentOrbState` switches on the kind rather than repeating the grouping, so the orb and the predicates cannot disagree about what a status means. Both status pollers now take predicates in place of literal status lists, which also tightened their response type from a bare `string` to `AgentStatus`.

`./check.sh guards` plus all four app suites (`app-core`, `app-web`, `app-desktop`, `app-mobile`) green locally.


elyxlz (commented):
Reviewed against issue #1135: the Option B remap is correct on both surfaces (web resolveAgentStatus + mobile orbColors), operation-driven states (authenticating/starting) still override with busy so active setup never reads as dead, no non-visual consumer keys on orbState, and the new tests fail on master. One minor cleanup:

- [ ] apps/mobile/src/components/AgentOrb.tsx:26-36 (minor): the diff leaves two consecutive branches returning the same designTokens.orb.busy: the second if held not_authenticated/unprovisioned before and now guards only setting_up, duplicating the branch above it. Fold setting_up into the preceding condition (status === "starting" || status === "restarting" || status === "rebuilding" || status === "setting_up") and drop the second if.

PR #1997, opened 2026-08-10
feat(voice,whatsapp): provider-agnostic STT with whisper fallback

emipasca:
## What

WhatsApp voice-note transcription now uses the user's configured speech-to-text (STT) provider directly, and falls back to local whisper.cpp only when no STT provider is configured or the provider call fails. This **supersedes #1970**, which hardcoded Deepgram inside the WhatsApp Go CLI and read the voice skill's private config file directly. The voice skill owns STT; WhatsApp calls it provider-agnostically.

## Design

**The voice skill owns one-shot transcription.**
- `SttProvider` (`providers/base.py`) gains `transcribe_file(audio_path, creds, stt_domain) -> str`.
- Deepgram implements it against the prerecorded REST endpoint (`POST /v1/listen`, `Authorization: Token`, model `nova-3`, `smart_format=true`, language from the STT domain when set and not `auto`, `multi` for multi-language). It returns the trimmed transcript, empty string when none.
- A new `voice-keys transcribe <path>` command reads the configured STT provider plus credentials from the voice config, runs `transcribe_file`, and prints **only** the transcript to stdout. Any failure (unconfigured, disabled, provider error) prints a JSON `{error}` to stderr and exits non-zero, following the skill output contract, so the caller can fall back.
- Provider/creds/enabled resolution moved to `config.py` (the config-shape owner); `handlers.py` now reuses those helpers instead of its own private copies (one owner per decision).

Extending the existing `voice-keys` entrypoint (rather than adding a new `voice` console script) is deliberate: skill CLIs are editable installs, so a new subcommand ships to the existing fleet on a plain upstream git sync, with no reinstall migration required.

**WhatsApp: voice STT primary, whisper fallback.**
- `transcribeAudioMessage` shells `voice-keys transcribe` first; a clean success is authoritative (even an empty transcript, i.e. genuine silence).
- On any failure it runs local whisper. The junk-tag detector (ported from #1970, anchored whole-string regex) runs on the whisper output alone: a bracketed silence tag like `[Musica]` or `[BLANK_AUDIO]` is dropped to empty rather than delivered as a transcript, while a real transcript that merely contains a bracketed word is kept.
- Go never reads `~/.voice/voice_config.json`; it goes through the command. The shellout has a 30s timeout, captures stdout as the transcript and stderr for logging, and never crashes when voice is absent.

## Tests (TDD)

- Python: Deepgram prerecorded parsing + language selection, and the `voice-keys transcribe` command (success, empty=clean success, not-configured, disabled, provider error).
- Go: `whisperOutputJunk` junk vs real (ported from #1970), `transcriptDecision` (STT wins, empty-STT authoritative, whisper fallback, junk dropped, both-fail error), `parseVoiceSttError`.

## Verification

- `voice` CLI pytest: 36 passed (13 new).
- whatsapp Go: `gofmt` clean, `go vet` clean, `go build` OK, full suite passed.
- Guards: repo ruff check + format, `check-conventions.py` clean.

https://claude.ai/code/session_0179KNWDABuK3z8sqveBBdPr

### DIFF
--- a/agent/skills/voice/cli/src/voice/config.py
+++ b/agent/skills/voice/cli/src/voice/config.py
@@ -60,6 +60,27 @@ def mutate(data_dir: pl.Path, updater: tp.Callable[[VoiceConfig], VoiceConfig]) 
     return new
 
 
+def provider_name(entry: Domain | None) -> str | None:
+    """The domain's configured provider name, or None when unset."""
+    if not entry or "provider" not in entry:
+        return None
+    name = entry["provider"]
+    return name if isinstance(name, str) and name else None
+
+
+def provider_creds(entry: Domain, provider: str) -> dict[str, str]:
+    """The credentials stored for one provider under a domain, or an empty dict."""
+    creds = entry["credentials"] if "credentials" in entry else None
+    if not isinstance(creds, dict) or provider not in creds:
+        return {}
+    value = creds[provider]
+    return dict(value) if isinstance(value, dict) else {}
+
+
+def is_enabled(entry: Domain | None) -> bool:
+    return bool(entry and "enabled" in entry and entry["enabled"])
+
+
 def _credentials(entry: Domain) -> dict[str, dict[str, str]]:
     creds = entry["credentials"] if "credentials" in entry else {}
     return dict(creds) if isinstance(creds, dict) else {}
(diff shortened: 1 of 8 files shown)


emipasca:
Superseded by #2368 (merged): the same provider-agnostic STT, reworked into a `transcribe <file>` console script in the voice skill that any skill calls, with the local whisper.cpp fallback inside the command (whatsapp now shells `transcribe` and dropped its in-Go whisper path). The whisper fallback is provisioned in the voice SETUP and an idempotent migration (compile + ggml-small). Thank you; this PR is what it was built from.

PR #2042, opened 2026-08-13
epic(browser): Zone.js js() fix, handover route deregister, and Troubleshooting notes

### LINKED ISSUE #1959: browser: js() raises TypeError on any page that replaces window.Promise (Zone.js/Angular)
## Summary

`helpers.js()` raises `TypeError` on any page that replaces the global `Promise`, which in practice means **every Angular site**, because Angular ships Zone.js and Zone.js swaps `window.Promise` for its own `ZoneAwarePromise`.

Since `navigate`, `tabs`, `snapshot`, `wait_for_load` and most of the CLI read the page through `js()`, effectively the whole ergonomic layer is unusable against those sites. Airline, bank and booking front ends are disproportionately Angular, which is exactly the category the skill's own docs point agents at.

## The failure

```
File "vesta_browser/helpers.py", line 99, in _eval_value
    return json.loads(value["value"])
TypeError: the JSON object must be str, bytes or bytearray, not list
```

## Cause

`agent/skills/browser/cli/src/vesta_browser/helpers.py`, in `js()`:

```python
wrapped = f"Promise.resolve({expression}).then(v => JSON.stringify(v === undefined ? null : v))"
```

Every call is routed through `Promise.resolve(...)`, even a synchronous one like `document.title`. On a Zone-patched page that returns a `ZoneAwarePromise`, and WebDriver BiDi's `awaitPromise` recognises only a **native** promise, so it does not await it. It serialises the object instead, and `script.evaluate` returns:

```json
{"type": "object",
 "value": [["__zone_symbol__state", {"type": "null"}],
           ["__zone_symbol__value", {"type": "array", "value": []}]]}
```

`_eval_value` then calls `json.loads()` on a list and dies.

## Why this one is worth fixing beyond the crash itself

The symptom reads as hostility rather than as a bug. The page is fully loaded (`document.readyState === "complete"`, body text present, screenshots fine) and every read command tracebacks, so the natural conclusion is that the site is blocking automation. The skill's own documentation warns at length against exactly that misreading, and this bug manufactures it. At least one agent has recorded "site hard-blocks automation" and handed the work back to its user when nothing was blocking anything.

## Fix

Do not create a promise for a synchronous expression; branch on whether the value is thenable:

```python
wrapped = (
    "(() => { const v = (%s); return (v && typeof v.then === 'function')"
    " ? v.then(x => JSON.stringify(x === undefined ? null : x))"
    " : JSON.stringify(v === undefined ? null : v); })()" % expression
)
```

**A tempting fix that does NOT work**, noted so nobody spends the time twice: taking the constructor from `window.__zone_symbol__Promise`. That symbol exists and `typeof` it is `"function"`, so it looks correct, but the value it hands back is still a `ZoneAwarePromise` and the crash is unchanged. Verified directly against a live Zone-patched page.

**Known remaining limit:** an expression that itself evaluates to a Zone-patched thenable still will not be awaited, because that is BiDi's promise detection. Such an expression has to be made synchronous or driven through `browser bidi` directly. Worth documenting rather than pretending it is solved.

PR follows, with three regression tests in `cli/tests/test_helpers.py` that fail on the current code and pass on the fix.

---
Submitted by **vesta** on vesta v0.1.188

### LINKED ISSUE #2036: browser: `handover stop` leaves its public vestad route registered
## What happens

`browser handover start` registers a **public** vestad service (`browser`) so the user can reach the noVNC page with no credential. `browser handover stop` returns `{"stopped": true}` and tears down websockify, x11vnc, openbox, headed Camoufox, Xvfb and the web root, but it never removes that registration.

The result is a registration with `public: true` pointing at a port nothing is listening on, and it persists indefinitely. The next handover then inherits a route it did not create.

## Reproduction

On a box:

```bash
source /run/vestad-env
SVC="https://$BOX_HOST:$VESTAD_PORT/agents/$AGENT_NAME/services"

# start clean
curl -sk --http1.1 -X DELETE "$SVC/browser" -H "X-Agent-Token: $AGENT_TOKEN"
curl -sk --http1.1 "$SVC" -H "X-Agent-Token: $AGENT_TOKEN"   # no `browser` entry

browser handover start --url https://example.com --user-data-dir /tmp/ho
curl -sk --http1.1 "$SVC" -H "X-Agent-Token: $AGENT_TOKEN"   # browser port=61001 public=true

browser handover stop                                          # {"stopped": true}
curl -sk --http1.1 "$SVC" -H "X-Agent-Token: $AGENT_TOKEN"   # browser STILL public=true
pgrep -x camoufox | wc -l                                      # 0
```

## Cause

`handover.py` has `_register_public_service()` with no counterpart. `stop()` enumerates what it tears down in its own docstring and the registration is simply not in the list. There is also no deregister helper in the vestad skill: `scripts/register-service` has no delete mode.

## Why it matters beyond tidiness

A handover exists specifically to let a user sign in to something that gates on account trust: bank, airline, email. The route is what makes that session reachable, so a route outliving its session is the exposure. It also presents as an agent that forgot to clean up, which sends the diagnosis in the wrong direction; the tool is not cleaning up.

## Fix

PR adds `scripts/deregister-service` (idempotent, treats an absent name as success) alongside `register-service`, and calls it from `handover.stop()`. Docs in `browser/SKILL.md`, `interaction-skills/handover.md` and `vestad/SKILL.md` updated. Tests added, including one that fails without the fix.

### LINKED ISSUE #2039: browser: troubleshooting section is missing the failure modes agents actually hit, so each one gets re-derived
The `browser` skill's Troubleshooting list covers install and ref staleness, but not the failure modes that repeatedly cost agent turns in practice. Each of these was re-derived from scratch on a live box, and none of them is discoverable from the command reference:

1. **`browser stop` takes no arguments.** "Stop the session by name" is not a thing: the session is selected by `BROWSER_SESSION` only. An agent told to "close your session" reaches for a `--name` flag, gets a usage error, and then reaches for `stop-all` as the next guess.
2. **`browser stop-all` is a global action that a subagent will reach for as a local fix.** One did, on a launch error, and killed four sessions belonging to concurrent work.
3. **A leftover session record is not a leaking browser.** `browser sessions` listing dead entries is stale metadata: `browser prune` reported 0 reclaimable bytes and there were no browser processes.
4. **`daemon did not come up within 30s` under fan-out is contention, not a broken install**, and the right response is the HTTP fallback rather than escalating to `stop-all` plus a process kill.
5. **The new-tab path can wedge while the existing tab context is fine.** Driving the existing context works; opening a new one times out.
6. **A consent-manager overlay can swallow clicks while every command reports success**, so a configurator or form reads back its DEFAULT values. That is a wrong number shaped exactly like a right one, which is the worst failure mode on this list.
7. **A WAF can answer per route**, so one probe does not characterise a host: on the same site a `GET` of an HTML page returned 200 with a JS challenge while a `POST` to its JSON API returned 403 with a block page.

Items 1, 2, 3, 6 and 7 are the expensive ones, because in each case the tool reports success or a plausible-looking result and the agent believes it.

PR follows with all seven written into the Troubleshooting list.

---
Submitted by **dory** on vesta v0.2.2

emipasca:
Folds the three open browser PRs into one review boundary, then converts the documented footguns into code fixes. Original commits keep their authorship; the combined browser CLI suite is green (**250 passed**).

Replaces and closes #1960, #2037, #2040.

### fix(browser): js() raises TypeError on Zone.js/Angular pages (#1960, vesta)
`helpers.js()` tracebacked on any page that swaps the global `Promise` (every Angular site, via Zone.js). Every call was routed through `Promise.resolve(...)`, which BiDi's `awaitPromise` does not recognise on a Zone-patched page, so it serialised the promise object and `_eval_value` ran `json.loads()` on a list. The fix branches on thenable, so a synchronous expression never creates a promise. The rejected alternative (sourcing the constructor from `window.__zone_symbol__Promise`) is recorded in the docstring so nobody re-spends the time. Three regression tests, verified failing on the old code. fixes #1959

### fix(browser): deregister the public handover route on stop (#2037, gianfranco)
`browser handover stop` tore down every process but never removed the **public** `browser` service registration, leaving a `public: true` route pointing at a dead port that the next handover inherited. Adds `vestad/scripts/deregister-service` (shaped like `register-service`, idempotent, 404 counts as success) and calls it best-effort from `stop()`. A missing script is a no-op, so older workspaces are unaffected. Three tests; docs corrected in `browser/SKILL.md`, `interaction-skills/handover.md`, and `vestad/SKILL.md`. fixes #2036

### feat(browser): fix the session-lifecycle footguns in code (replaces the docs-only approach of #2040, dory)
#2040 documented seven failure modes hit on a live box. Five of them were code bugs wearing a documentation costume, so this PR fixes them at the CLI layer instead of teaching agents to route around them:

| Footgun | Fix |
|---|---|
| `stop` took no arguments, inviting an invented `--name` then `stop-all` | `browser stop [session]` |
| `stop-all` killed concurrent work (4 sessions once) | refuses while other sessions are live, lists them; `--force` overrides |
| `daemon did not come up within 30s` under fan-out | startup wait scales when other live sessions contend for CPU; timeout error names them and the `http_get` fallback; fails closed (spawn killed, records removed); a concurrent command's healthy daemon is accepted |
| dead session records read as leaking browsers | `sessions` reaps records whose browser AND daemon are both dead |
| a consent overlay swallowed clicks while `click` reported success | intercepted click exits non-zero naming the interceptor on stderr |

Two of the seven were already fixed on master before this PR (`new_tab` falls back to an existing context; `click_ref` detects occluders): for those and the rest, the Troubleshooting section now states the current mechanism plus the judgment that stays with the agent (verify a control changed before trusting a read-back; a WAF can answer per route, so one probe is not a verdict). Nine new behavior tests, each red on the prior behavior. fixes #2039

## Verification
- `uv run pytest agent/skills/browser/cli/tests/` -> **250 passed**


emipasca:
## Docs-to-code pass: the Troubleshooting footguns are now fixed in the CLI

Maintainer call: most of #2040's seven documented failure modes were code bugs, so they are now fixed in code (`f9315110`) and the docs restated around the new reality (`d794feea`).

**Fixed in code (5):**
- `browser stop [session]` accepts a session name.
- `browser stop-all` refuses while sessions other than the caller's are live, listing them; `--force` overrides.
- `ensure_daemon` scales its startup wait under contention (other live sessions), names the competing sessions and the `http_get` fallback in the timeout error, fails closed (spawn killed, records removed), and accepts a concurrent command's daemon when it lost the same-session spawn race.
- `browser sessions` reaps records whose browser and daemon are both dead, so stale metadata never lists as a session to chase.
- An intercepted click (`overlay on top at the click point`) exits non-zero naming the interceptor on stderr; the page state after such a click looks exactly like success, so only the exit code can say otherwise.

**Already fixed on master before this PR (2):** the new-tab wedge (`new_tab` falls back to an existing blank context, bounded by the BiDi command timeout) and occluder detection itself (`click_ref`). The incident box that produced #2040 predates those.

**Stays documentation (judgment, not code):** verify a control changed before trusting a read-back; a WAF can answer per route, so one probe is not a verdict; give each subagent its own session.

**Verification:** 250 passed (9 new behavior tests, each red on the prior behavior), `./check.sh guards` green. Docs follow the prompt guide: current mechanism only, no changelog language.

emipasca:
@vestabot, check this pr again fresh

elyxlz:
Re-reviewed fresh after the polish and docs-to-code passes (now +525/-31). I changed nothing; read-only.

**Fixes:** #1959 and #2036 fully. #2039 over-satisfied: five of its seven footguns are now fixed in code (`stop [session]`, `stop-all` guard, contention-scaled `ensure_daemon`, `sessions` reaping, intercepted-click exit code), not just documented.

**Blocking:** none.

**Non-blocking**
- `browser/SKILL.md:40-41` still shows the intercepted-click note as a stdout `# e14 is covered by <div...>` line and calls it "the click's own output". `cli.py:285` now prints that to **stderr** with no `#` and exits 1 (clicking.md updated, this paragraph not). The entry doc and the behavior now disagree.
  Proof: `git grep '# e14 is covered'` → SKILL.md; `cmd_click` prints to `file=sys.stderr` and `return 1`.
- `admin.py:272-274` fail-closed path kills `proc.pid` then unlinks `socket_path/pid_path`. Under two same-session fresh spawns where the winner is alive but not yet healthy at the loser's timeout, this deletes the live winner's records and orphans it. Needs the discouraged same-session concurrency plus a >budget handshake.
  Proof: none, read only.

**CI:** `test_admin.py::test_shutdown_is_fine_without_an_ephemeral_profile` fails identically here and on `master`, a PermissionError unlinking a shared `/tmp/vesta-browser-*.log` owned by another uid, not this diff. Rest: 249 passed.

**Verdict:** MERGE. The new stop/stop-all/contention/reaping/click code is correct and each new test reds on the prior behavior; the one concrete defect is a stale SKILL.md paragraph the polish sweep missed (one-line fix, not blocking), and the daemon race is narrow and behind a pattern the docs tell agents to avoid.

---

emipasca:
@vestabot fix the non blocking

elyxlz:
Fixed both non-blockings and pushed to `epic/browser` (`5d46a861`, `cc708ff5`).

**1. Fail-closed race (`admin.py`).** Extracted the cleanup into `_discard_own_spawn(session, pid)`: it still kills the timed-out spawn, but unlinks the socket/pid records only when the recorded pid matches our spawn's. A concurrent same-session spawn that won the socket recorded its own pid, so its still-coming-up daemon is no longer orphaned. Extracting it also kept `ensure_daemon` under the branch ceiling (the inline guard tripped PLR0912 13>12).
  Proof: new `test_ensure_daemon_timeout_leaves_a_concurrent_winners_records` reds on the old unconditional unlink (deletes the pid `9999` record), greens on the fix; `test_ensure_daemon_timeout_fails_closed` still green.

**2. Stale click doc (`SKILL.md:40`).** Rewrote the intercepted-click paragraph to match the new behavior and `clicking.md`: exits non-zero, names the interceptor on stderr, no `#` prefix.
  Proof: `grep -rn "took the click" agent/skills/browser` → only the corrected form in SKILL.md and clicking.md.

`ruff check` + `ruff format --check` + `check-conventions.py` clean on the touched files. Browser CLI suite green bar the pre-existing `test_shutdown_is_fine_without_an_ephemeral_profile` (shared `/tmp` log owned by another uid, fails identically on master).

**Verdict:** MERGE. The two review nits are resolved with a pinned regression test and a doc that no longer contradicts the code; nothing else surfaced.

---

PR #2061, opened 2026-08-15
epic(tasks): four folded PRs, the interface simplification, and computed checkpoints

### LINKED ISSUE #1981: tasks: no guidance on staged files that reminders point at
When a reminder's job carries real material (a draft to send, a verified list, a decision plus the reasons behind it), the practical pattern is to keep that material in a file and have the reminder name the file. The tasks SKILL.md documents reminders as instructions but says nothing about this split, so each instance rediscovers the same four failure modes by getting burned:

1. No stated precedence between the reminder message and the file, so when they disagree there is no rule for which to believe.
2. New findings appended BELOW the instruction they invalidate. The reminder fires, the file is opened in a hurry, and the first thing read is the superseded plan.
3. Schedule data (ids, dates, times) duplicated between `remind list` and the file, which goes stale the first time either is edited.
4. Relative dates written into staged files ("tomorrow", "after the trip"), true when written and false the next morning, with nothing to flag them.

These instructions are written by the agent to itself, so they arrive with its own authority and a deadline attached and are followed harder than ordinary notes. That is what makes a stale one costly rather than merely untidy.

Proposal: a short subsection in `agent/skills/tasks/SKILL.md` under Reminders stating the convention (file wins, dated current-answer block at the top, no duplicated schedule data, absolute dates only).

---
Submitted by **dory** on vesta v0.1.188

### LINKED ISSUE #1999: tasks: remind snooze supports --at but nothing documents it, and --in-hours is now-relative
`tasks remind snooze` accepts `--at <iso> --tz <tz>`, and the implementation handles it, but nothing tells the agent so. The usage line printed on error, the `--help` output, and the tasks SKILL.md all show only `--in-hours N`. The only place `--at` is mentioned is inside a ValueError raised when no timing flag is given at all, which you only see if you get it wrong in a specific way.

Two consequences, both hit on a live box today.

1. `--in-hours N` counts from now (`run_time = _now_utc() + offset`), not from the reminder's own scheduled time. Asking for "seven hours later" on a reminder four days out moves it to this evening. That is the correct implementation of "snooze", but combined with the undocumented `--at` it is a trap: relative is the only advertised option, so it gets used for absolute moves.

2. Not knowing `--at` exists pushes you to delete and recreate the reminder instead. That changes the id, and every note, file and message that referenced the old id silently becomes wrong. On this box three separate files referenced the old id and one was missed until a checker caught it.

Proposal: mention `--at`/`--tz` in the snooze usage strings in `cli.py` (so the answer appears where the mistake happens), and document both the now-relative semantics and the `--at` form in `agent/skills/tasks/SKILL.md`.

---
Submitted by **dory** on vesta v0.1.188

### LINKED ISSUE #2023: tasks remind list silently truncates the human table at 50 rows
`tasks remind list` prints the first `REMIND_LIST_PAGE_SIZE` (50) reminders and gives no indication that it held anything back. No footer, no ellipsis, no count. The last row it prints looks exactly like the last row that exists.

That makes the table answer absence questions wrongly, which is the same failure `--json` was already fixed for. The existing comment in `_remind_list_cmd` spells the reasoning out for the JSON path:

> JSON is consumed by scripts asking absence questions ("is anything scheduled for X?"), so a silent page size would make a truncated list read as "no".

The table gets asked that question too, by eye or by grep, and there a silent page size reads as "no" just as badly.

**Observed.** An agent with 123 live reminders ran `tasks remind list | grep <topic>`, got no match, and concluded that two scheduled reminders did not exist. They existed, 12 and 15 days out, well past the visible window. The conclusion was acted on: the "missing" reminders were recreated as duplicates, and the originals were nearly deleted as stale.

```
tasks remind list          -> 50 rows
tasks remind list --json   -> 123 reminders
```

`tasks list` is not affected (33 rows both ways), so this is specific to `remind list`.

**Not a documentation gap.** The page size is already documented in `--help`: *"table shows the first 50; --json/--json-pretty list all unless --limit is given"*. Documented is not the same as noticeable. The reader who needs to know is the one already looking at a table that appears complete, and they are not reading `--help` at that moment.

**Suggested fix.** Keep the page size, which is worth having, and print a footer when it truncates, naming the total and the ways to see the rest. An explicit `--limit` is the caller asking for a cap, so it should stay silent.

---
Submitted by **mozzy** on vesta v0.2.2

emipasca:
One review boundary for the tasks skill: folds four open tasks PRs with every finding from their vestabot reviews applied, then simplifies the agent-facing interface and replaces the materialized auto-reminder ladder with computed checkpoints. Original commits keep their authorship; the combined tasks CLI suite is green (**251 passed**).

Replaces and closes #1951, #1982, #2000, #2024, and #2067 (the interface work, folded after review). The fifth open tasks PR, #1958 (snapshot), was deliberately dropped: the owner judged metadata version history not worth a feature to maintain, and #1957 is closed with that rationale.

### feat(tasks): `remind list --show-completed` (#1951)
Reveals fired one-shots so a self-chaining reminder can re-read its own body. The review found the flag's default form hid the row it exists to recover: one-shots sort `scheduled_time ASC`, so a backlog of old fired rows filled the 50-row page. Fixed in the epic: fired one-shots sort after pending ones, newest fired first. Also from the review: the table marks fired rows with `[fired]`, the flag is named in SKILL.md and `--help`, and `remind delete`/`update` accept a row the flag reveals. Regression tests for each.

### docs(tasks): staged-file convention for reminders (#1982) (fixes #1981)
A reminder whose job carries real material keeps it in a file; the message is the trigger. The review's two suggestions are applied: a task-linked reminder's file is the task's own `metadata_path` file, and file-wins precedence is stated once in the skill. On top, the section prescribes dated blocks: dated headings, inserted at the top, timestamps pasted from `date` rather than typed from memory (the instruction-level counterpart to #2033's note-top).

### docs(tasks): surface `snooze --at` (#2000) (fixes #1999)
Documents that `--in-*` counts from now and `--at <iso> --tz <tz>` moves a reminder to a specific time. The review's blocking finding, a clause claiming `--at` is absent from `--help` while the same PR added it, is dropped.

### fix(tasks): say when `remind list` truncated the table (#2024) (fixes #2023)
The table names the total when it held rows back, and the footer rides stderr (the review's blocking finding), so `tasks remind list | grep <topic>` cannot eat the notice along with the rows. Tests assert the stream split.

## The interface simplification (from #2067)

Built around how an agent operates a CLI: syntax that matches priors, one vocabulary per concept, probes that teach, output that confirms the effect.

- **One time vocabulary.** `tasks create` and `tasks update` accept `--at`, `--tz`, and `--in-minutes/--in-hours/--in-days`, the same spellings `remind`, `snooze`, and `postpone` use. The `--due-*` forms remain as aliases.
- **Probes teach.** `tasks remind <subcommand> --help` prints usage and exits 0; the parsers were `add_help=False` and punished the agent's most natural move with exit 2.
- **Self-confirming snooze.** Three timing forms, one per call: `--in-*` from now, the new `--by-*` from the reminder's own fire time, `--at` + `--tz` for a named moment. Mixing is rejected, and the result echoes `previous_run` and `next_run`, so misreading from-now vs from-fire-time is caught in the same command's output. One `SnoozeSpec` model serves the CLI and the HTTP endpoint.
- **No invented dates.** `--recurring daily` accepts a bare time (`--at "21:30"`). Weekly/monthly/yearly keep the date, which carries their weekday or day.

## Computed checkpoints (the virtual ladder, from #2067)

Pre-due checkpoints are computed, never stored. Each serve tick derives the ladder from `(due_date, created_at)` with fixed lead times (15m, 1h, 1d, 1w, then doubling from 2w) and one marker column, `tasks.checkpoint_fired_through`; the newest uncovered rung fires and the marker advances. That deletes the materialized machinery and its semantics:

- No auto-reminder rows, so no regeneration on `postpone` or due change: a moved date re-arms the ladder by construction.
- No retitle machinery: messages are built at fire time from the current subject.
- No `remind update` ownership dance, no `*` marker, no `auto:` labels: a hand-written checkpoint is an ordinary `tasks remind --task <id>` reminder.
- Daemon downtime collapses to one catch-up fire per task instead of a backlog.
- SKILL.md sheds the paragraphs that taught the removed semantics.

Schema v8 adds the marker column, drops `auto_generated`, deletes the materialized rows (rungs a user rewrote have `auto_generated = 0` and survive as the plain reminders they became), and initializes the marker to now so an in-flight ladder continues instead of refiring. The ladder-building code the released v2/v4 migration steps call stays frozen in db.py for pre-v4 databases.

## Verification
- `uv run --project skills/tasks/cli pytest skills/tasks/cli/tests/` -> **251 passed** (142 unit incl. 27 new checkpoint/interface tests, 109 e2e)


elyxlz:
**Fixes:** closes #1951 (show-completed), #1982/#1981 (staged-file convention), #2000/#1999 (`snooze --at` doc), #2024/#2023 (truncation footer): each fully. #1958 lands the `snapshot` subcommand only; its scheduled caller #1957 stays open, as the body states.

**Blocking:** none.

**Non-blocking**
- `commands.py:remind_list` now fetches every row (`limit=None`) whenever the table pages, then slices in Python, to know `total` for the footer. Fine at reminder scale, but it drops the SQL `LIMIT` for the common path.
  Proof: none, read only; a `COUNT(*)`+`LIMIT` would avoid loading all rows, not worth it here.
- `remind_delete`/`remind_update` dropped `AND completed = 0`, so a fired reminder is now mutable via the daemon HTTP routes too (`server.py:143,159`), not only the revealed CLI row. Intended by the epic; noting the widened reach.
  Proof: none, read only.

**Verdict:** MERGE. Attacked the sort against a full page of old fired rows, the footer's stream split, the prune's foreign-file guard, the post-v7 schema columns, and every reader of the changed `status` field: it held. Reverting the `completed ASC` key reds `test_old_fired_rows_never_page_out_pending_ones`; moving the footer to stdout reds `test_footer_rides_stderr...`, so both fixes are pinned.

---

emipasca:
Polish pass on this PR (one commit, 02703065: net -28 lines). Summary per pass:

**Review (correctness + reuse).** No correctness bugs found in the fold. I checked the new sort/filter logic adversarially: only one-shot (`date`) rows are ever marked `completed = 1`, so the three-group ordering and the `--show-completed` filter are sound, and the paged table keeps the same rows the SQL LIMIT would have kept. Two hygiene fixes in the new test file: dropped a local `tmp_config` fixture that shadowed the identical one in `conftest.py`, and switched `_mark_completed` to `closing(db.get_db(...))` so the connection is closed like every sibling test does.

**Simplify.** Three cleanups, behavior-preserving (248/248 CLI tests pass):
- `cli.py _remind_list_cmd`: the three-branch `limit` cascade always passed exactly `args.limit` to the DB call, so it is gone; one `paged` boolean now decides whether the table page applies, and the truncate/footer condition is computed once instead of twice.
- `commands.py remind_list`: the `done_filter`/`done_where` twin spellings of the same predicate collapsed into one WHERE built from a predicate list, taking the two near-identical `conn.execute` branches to one. The five-term ORDER BY is now a single rank CASE (recurring 0, pending 1, fired 2) plus one DESC and one ASC key, which states the three-group intent directly.
- `test_remind_list_json_limit.py`: the page-size test no longer re-asserts the footer that the dedicated footer test three lines below owns.

**Prompt sweep.** Reviewed the changed SKILL.md lines against the vesta-prompt-guide and the official prompting guides: they already conform (explicit mechanism plus motivation per rule, concrete examples, no previous-design prose, no banned dash separators). No edits.

**CLAUDE.md.** No staleness caused: everything here lives inside one skill, below CLAUDE.md altitude. No section added. One borderline item flagged for you rather than written: the stderr truncation footer effectively extends the skill output contract with "a successful command may print an advisory notice on stderr", which could merit a clause in the AGENTS.md output-contract paragraph if you want it codified.

**Flagged, not fixed (out of polish scope):**
- The CLI-invocation helper (`monkeypatch Config / live_pid / argv, run cli.main()`) now has three copies across test files; lifting one into `conftest.py` would be the reuse fix, but that file is outside this PR's diff.
- `remind update` on a fired reminder now succeeds (revealed ids are actionable) but leaves the row `completed = 1`, so it stays hidden and armed-never: fine if intended as a record edit, but worth a deliberate decision whether update should re-arm like `snooze` does.

**Verify.** `./check.sh guards` green; tasks CLI suite 248 passed; `./check.sh agent` 1131 passed with 2 pre-existing local-only failures in `test_migrations.py` workspace-repair tests caused by this host's global `tag.gpgsign=true` (documented known artifact; unrelated to this PR, passes in CI).

emipasca:
## Polish pass (`f8389586`)

Four passes over the PR diff, one commit on top of the author's history.

### Pass 1, review (`code-review xhigh --fix`)
- **Legacy import copied the materialized ladder.** On the upgrade boot the migration runs before daemons start (`collect_boot_turns`), so `~/.tasks/tasks.db` is still v7 with `auto_generated = 1` rungs when step 4 first runs `reminders`; every "Task due in 1 day: X" row was imported as a plain one-shot and would fire from the reminders daemon next to the computed checkpoint, task link gone. `_copy_legacy_rows` now skips those rows when the column exists, and a failed read leaves `legacy_import_done` unset so the next init retries (`reminders_cli/db.py`, two new tests).
- **Re-dating an old task fired a mislabelled past rung.** Task created 30 days ago, `tasks postpone <id> --in-days 3`: the new ladder's due-1w rung sits 4 days in the past but after the floor, so the next tick wrote "Task due in 1 week" for a task due in 3 days. Every due-date write re-stamps `checkpoint_fired_through = now` (`tasks_cli/commands.py`, regression test).
- **Writes on a soft-deleted row succeeded and reported a live-looking result** (`reminders snooze/update`, `tasks postpone/done/update`): now refused with the delete instant (`_require_live_*_row`, parametrized tests, SKILL.md sentences).
- **Stray dashboard `index.css` regression** (dropped `@source` lines and the transparent html/body backgrounds): restored to the sync script's output; `check.sh guards` was red on the branch.
- **Dashboard tasks page fetched `tasks/reminders`**, an endpoint the PR removes: reads `reminders/reminders` and drops the `task_id`/`auto_generated` fields (`agent/skills/dashboard/app/src/pages/tasks.tsx`).
- **Stale `tasks remind` / "reminder with the `tasks` skill" references** in `birth`, `dream`, `flights`, `proactive-check` SKILL.md repointed at `reminders`; birth also sets up `reminders` on a fresh box (fresh agents pre-mark the migration, so nothing else installs it).
- **Migration**: bare `skills-activate` is not on PATH (full `~/agent/skills/skills-registry/scripts/` path); step 6 also removes the stray nightly contacts reminder a version-skipping box still carries (the shipped `2026-07-contacts-reminder-to-dream` can only find it via `tasks remind`); new step 7 carries an agent's `### Reminder Patterns` notes from tasks/SKILL.md to reminders/SKILL.md (upstream deletes those lines, so a filled section conflicts on merge).
- Six reminder tests pinning `STALE_FIRE_SLACK`, `MISSED_GRACE`, snooze reactivation, second precision, recurring rejection, and required timing were dropped with the tasks suite: ported to `reminders/cli/tests/test_snooze.py`.
- Note: `birth`, `dream`, `flights`, and the dashboard page were **outside the PR's file set**; the edits are one-line repoints of references the PR broke. Drop them if you would rather ship them separately.

### Pass 2, simplify
- `tasks postpone` rides `_add_due_args` and takes a `DueSpec` (one time vocabulary, one builder; `_due_spec(args)` helper in cli.py, six test call sites updated).
- `_visibility_clauses` shared by `list_tasks`/`search_tasks`; `_OPEN_TASK` predicate owned once for the digest and checkpoints; `del task["checkpoint_fired_through"]` instead of a `.pop` fallback; dead `Reminder` TypedDict and orphaned `delete_task_reminders` removed from tasks `db.py`.
- Reminders: `_cron_schedule` folds the duplicated recurring/`--cron` tail; `LIVE_REMINDER` predicate shared by `_sync_jobs` and both restore paths; `_echo_local` renders through `_zone(None)` like tasks (the `tzset` fixture is now a `monkeypatch.setenv`); snooze exclusivity is a plain `and`; the "Say when" error names `reminders snooze`; `_local_tz_name` inlined into `_zone` (both skills); `_print_list`'s `.get()` inlined and `_add_id_args` shared; the scheduler comment states the current mechanism; duplicate `tmp_config` fixture and legacy-db builder removed.
- Skipped, for the author: (a) the `--due-*` aliases carry no `LEGACY(remove-when: ...)` marker (both reviewers flagged; permanent or transitional?); (b) `reminders list` table fetches every row to name the exact total, a deliberate choice from #2024, left as is; (c) v8 rebuilds the dormant reminders table and v9 adds `deleted_at`, both unreleased, so they could be one step that leaves the table untouched (then the importer knows exactly one shape); (d) `_reminder_view` could carry a `recurring` bool so the dashboard stops inferring it from the label prefix; (e) create silently prefers `--at` when `--in-*` is also given while snooze rejects mixing; (f) test-helper copies across the tasks test files (`_add_task_due_in` etc.) left alone. Pre-existing, untouched: the dashboard filters `status !== "done"` (status is `completed` since v7); the `Task` TypedDict in tasks `db.py` is unreferenced.

### Pass 3, prompt sweep (`vesta-prompt-guide`)
- reminders SKILL.md: `--show-completed`/`--show-deleted` and `get --field` were each explained in three places; each fact now lives once.
- Migration: the import runs on the first `reminders` command (init_db), not on the daemon's first tick; the prose says so.

### Pass 4, AGENTS.md
- The "Adding a skill" frontmatter example copied the tasks header verbatim; updated to the split header (`Tasks (CLI: tasks)`, new description). Nothing at architecture altitude was added: the reminders skill is a default skill with its own daemon, covered by the existing skill/daemon-contract prose. Nothing borderline flagged.

### Verify
- `uv run --project skills/reminders/cli pytest ...`: 145 passed; `... skills/tasks/cli ...`: 145 passed.
- `./check.sh guards`: green (dashboard sync now clean).
- `./check.sh agent`: `ty` green; pytest shows 82 failures on this macOS host, all host-environment (bash 3.2 has no `mapfile` for `build-upstream.sh`, no `/proc` for `test_status_rejects_a_reused_pid` on every daemon row, ssh-tunnel/dashboard/file-host/moneypot/sign-service daemons cannot start here). None is in a file this PR or the polish touched; the new `reminders` daemon-contract row passes every test the other rows pass. CI (Linux) is the authority for that suite.

emipasca:
### Addendum (`f35fcc33`)

The review's line-by-line finder came back after the polish commit with two more bugs, applied in one small follow-up commit (tasks 146 passed, reminders 146 passed, ruff and conventions clean):

- **Create-time boundary fire.** `tasks create "X" --in-days 1` puts the due-1d rung at the creation instant; `created_at` is second-truncated, so the first tick fired "Task due in 1 day: X" on a task just made. `add_task` now stamps `checkpoint_fired_through` like every other due write (`test_creating_a_task_fires_no_rung_that_lands_at_creation`).
- **Bare time dropped its offset.** `--recurring daily --at "21:30+02:00"` was read as 21:30 in the agent's zone: `datetime.combine(..., tzinfo=local_tz)` overrides the time's own tzinfo. Combine without it and resolve like any aware datetime (`test_daily_reminder_bare_time_keeps_its_own_offset`).

PR #2062, opened 2026-08-15
epic(dream): reality-check hardening, secret and mail-residue scanners, guards, and memory tooling

emipasca:
Folds the eighteen open dream PRs into one review boundary, in the same shape as #2042 (epic/browser). Original commits keep their authorship; the branches were merged oldest first and the cross-PR conflicts were resolved semantically, with each resolution noted below.

Replaces and closes #1909, #1919, #1926, #1928, #1953, #1955, #1963, #1973, #1977, #2003, #2013, #2015, #2029, #2033, #2046, #2053, #2054, #2058.

### reality_check.sh (five PRs, one composed probe set)

- **#1909**: age-bound the error-storm count (dated lines from the last 2 days, undated logs counted whole and labeled), plus the append-only run history (`reality-history.tsv`) so a gap in the series is itself a finding.
- **#1919**: RED when the host disk is nearly full (97%), with own-usage and host-usage reported as two independent facts.
- **#1973**: drop healthy "0 error(s)" summary lines before counting, and add the unprotected hand-made-state probe (`unprotected-accepted.txt` records judged exposures).
- **#2015**: own-usage RED requires real filesystem pressure (>= 85% host usage), not size alone; below that it prints as context.
- **#1955**: RED when the nightly dream lapsed (newest `~/agent/dreamer/` summary >= 30h old), keyed on the summary file, not a commit message.

Conflict resolutions: #1973's healthy-line filter is applied inside #1909's dated/undated windows, so both the aging and the filter hold at once. #2015's pressure gate replaces #1919's unconditional own-usage RED while #1919's independent host-disk report stays. The dream-lapse probe runs before the history append, so the recorded RED count includes it.

### New instruments

- **#2003**: `PreToolUse` guards for silent-failure shapes (`tmp_ref_guard.py`, `blind_mutation_guard.py`), declared in the skill's `hooks.json`, wired at boot by `claude_runtime.py`, verified by `verify_hooks.py` (which the reality-check pass now runs, because the guards fail open and a broken one says nothing). Tests: `test_dream_guards.py`, `test_skill_hooks.py`.
- **#1977**: `scan_files_for_secrets.sh` walks `/tmp` and `$HOME` for credential-shaped files; the store-only sweep cannot see files at all. Tests: `test_scan_files_for_secrets.py`.
- **#1963**: `check_mail_residue.sh` finds mailbox content left in agent-written files, the surface every store-named check misses.
- **#2029**: `memory-budget` prints MEMORY.md growth since the last commit, longest added lines first, and a per-section breakdown; exits non-zero above a soft budget so it can gate additions. Also fixes `memory_size.sh` to count characters with `LC_ALL=C.UTF-8 wc -m`. **Supersedes #2013**, which shipped the bare `wc -m` half of the same fix; the epic ends at #2029's version.
- **#2033**: `note-top` inserts dated blocks at the top of agent-written files and fills `@@NOW@@` from the clock, because `cat >>` buries the newest block and hand-typed timestamps drift.

### redact_secrets

- **#1953**: stop reading channel-store identity columns as payment cards (plus regression tests). The PR incidentally dropped the script's executable bit; the epic restores it.
- **#2054**: docs, scrub a confirmed-live secret by literal (`--scrub-literal`), never by ref, since refs only cover the rows you were shown.

### SKILL.md guidance

- **#1926**: diagnose a silent day from token counts (`in=0 out=0 cache_read=0` means the turn never ran), audit the meta-retrospective with a skeptical subagent, weigh fixes by whether they fire without you, and sweep `/tmp` (worktrees removed via git, never `rm -rf`).
- **#1928**: grep a tool's output using the source's own header string, never a paraphrase; plus two disk-cleanup traps (hidden `.venv` entries, load-bearing model caches) and the harness scratch sweep.
- **#2046**: check the dashboard has a user before the nightly widget pass (with the matching dashboard SKILL.md section).
- **#2053**: read a skill from disk before following it; an inline rendering is a paraphrase that invents specifics.
- **#2058**: removing a check is not finished until its coverage is accounted for, in the same pass.

Conflict resolution: the curation-review checkpoint lookup takes #1955's loose `--grep '^dream'` match and keeps #1926's sanity check on the date it returns; the two guarded the same trap from different ends.

One repo-side fix on top: `test_reality_check.py` asserted on `re.search(...)` without narrowing the `None` case, which `ty` rejects; the epic adds the assert.

## Review findings applied on top

Every folded PR was reviewed by vestabot; most blocking findings were already fixed on the branches and rode in with the merges. The findings still open at fold time are fixed here, each pinned by a test where it is behavior:

- **#1909 blocking**: the dated error count dropped undated continuation lines, so an 80-traceback storm read as 80 lines and a live storm could hide under the threshold. An undated line now inherits the recency of the dated line above it.
- **#1973 blocking**: the healthy "0 error(s)" filter also dropped `0 warnings, 47 errors` and `worker 0 error: connection refused`. A line naming a nonzero count, or using the bare singular, is kept, and the filter finally has tests (the review's other finding).
- **#2015 blocking**: no test pinned the new no-pressure OK branch; two tests now do. Also from that review: a failed `df` no longer fails the own-usage RED open, and the header comment states the composed rule instead of the old single-term one.
- **#1963 blocking (the two the follow-up commit did not cover)**: the scan skipped `~/.microsoft`, where `email get` auto-saves every fetched body, so the run it demands before saying "gone" exited 0 with the mail sitting there; a named domain now gets a domain-only pass over the store caches, with any distinct-address hit removable (a mail cache holds no prose). And every content grep runs with `-a`, because GNU grep reports a NUL-bearing dump on stderr and prints nothing.
- **#1909 non-blocking**: the history line printed the timestamp it had just appended, so a lapse could never appear; it now prints the previous run's.
- **#1953 non-blocking**: SKILL.md said `--show` prints the row's full text; it prints the scannable columns, and now says so.
- **#2033 non-blocking**: SKILL.md invoked bare `note-top`, which is not on PATH; it now uses the full path like its siblings.
- **#2029 non-blocking**: the per-section table divided awk byte lengths by a character total, so the column summed past 100% on non-Latin text; it now divides by its own measurer's total.
- **#2029/#2033 prose**: the script headers narrated their originating incidents, which prose under `agent/` must not do; they now state the mechanism.

**Deliberately not applied**, matching what each PR's author declined with reasoning the reviewer accepted: the #1926/#1928 SKILL.md trimming pass (a separate concern from a fold), #1977's recount inefficiency, #2003's tasks-CLI guard gap, and #2058's placement suggestion.

## Verification

- `./check.sh agent` -> `tests/test_reality_check.py` 25 passed; full suite green except the two `test_migrations.py` workspace-repair tests that fail on any host with `tag.gpgsign` set (a known local-environment failure, green in CI)


elyxlz:
**Fixes:** nothing linked as an issue. Folds 18 open dream PRs (#1909…#2058) into one boundary. Those are PRs, not issues, so the `closes` list does not auto-close on merge (keywords close issues, and only the first number after one keyword binds); the 18 need manual close. The fold itself is coherent.

**Blocking:** none.

**Non-blocking**
- `blind_mutation_guard.py:4` and `tmp_ref_guard.py` docstrings narrate incidents and dates, which the `agent/` prose rule forbids (state the mechanism, not what changed). Inconsistent: the PR applied this same fix to the #2029/#2033 scripts but left the guards.
  Proof: `blind_mutation_guard.py:4` "WHY THIS EXISTS (built 11 Aug 2026). On 10 Aug I ran this…", plus "closed 12 Aug 2026", "added 14 Aug 2026".
- CI: `Agent · Typecheck and test` and the vestad jobs are still pending; `Repository · Lint and guardrails` is green. Nothing red; not the diff's concern.

Attacked on the one genuinely new seam (`_merge_skill_hooks`): add-only, keyed by absolute command, preserves other settings, no-op on the second boot; `verify_hooks.py` reads the exact shape the merge writes; both guards self-test clean and the 52 dream tests pass locally.

A `polish-pr` pass would trim the inherited incident-narration prose.

**Verdict:** MERGE — hook wiring is idempotent and add-only, guards self-test clean, dream suite green; only leftover prose narration to tidy.

---

emipasca:
Closing this fold in favour of four single-concern PRs rebuilt from master after reviewing each of the 18 folded PRs against current master (7 kept with nits, 7 rewritten, 4 dropped; about 200 of the 2,751 lines survive):

- #2351 `fix(dream): harden the reality_check probes` (1909, 1919, 1973 part 1, 2015, 1955, plus the refused-turn probe from 1926)
- #2354 `fix(dream): scan transcripts and files for secrets, skip identity columns, widen patterns` (1953, 1977, and the open PRs 2275, 2289, 2328, 2157)
- #2355 `docs(dream): guidance trim` (1926, 1928, 2046, 2053, 2058, 2029, 2003 one-liner)
- #2357 `fix(microsoft): partition saved mail bodies by account so auth remove can delete them` (1963)

Dropped: 2003 (premise false on master), 2013 (superseded by 2029 hunk), 2033 (workspace tooling), 2054 (inverts the ref-keyed scrub design). Each folded PR carries a comment saying where it went. All four replacements are merged.

PR #2069, opened 2026-08-15
ci: keep whatsmeow current via daily Dependabot

emipasca:
## What

Add a `gomod` Dependabot entry for `agent/skills/whatsapp/cli`, scheduled **daily**, so whatsmeow (and the CLI's other Go dependencies) tracks upstream as closely as Dependabot allows.

## Why

whatsmeow's protocol code ages and WhatsApp periodically breaks old versions (issue #1073). Dependabot did not cover any of the Go modules, so the pin only moved when someone ran `whatsapp update-deps` by hand.

This keeps it current through a **reviewed, CI-gated PR** rather than a runtime `@latest` float. The launcher (`agent/skills/whatsapp/whatsapp`) pins whatsmeow deliberately: a mid-session bump can invalidate a multi-device session and log the companion device out, and an on-every-serve recompile turned a keepalive into a churn loop. Dependabot bumps `go.mod`/`go.sum` on master; each bump is gated by `./check.sh whatsapp` (Go lint + build + test), and the fleet rebuilds against the new pin on the next daemon restart via upstream sync. Daily is Dependabot's finest cadence.

## Scope

Only the whatsapp CLI module. `telegram/cli` and `otp/cli` are left as-is.

https://claude.ai/code/session_01LdguFimBQw8xLJNn1zaMG1

---
_Generated by [Claude Code](https://claude.ai/code/session_01LdguFimBQw8xLJNn1zaMG1)_

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -55,7 +55,7 @@ updates:
   - package-ecosystem: gomod
     directory: /agent/skills/whatsapp/cli
     schedule:
-      interval: weekly
+      interval: daily
     groups:
       whatsapp-deps:
         patterns: ["*"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -49,6 +49,17 @@ updates:
       - dependency-name: "eslint-config-expo"
         update-types: ["version-update:semver-major"]
 
+  # Keeps whatsmeow (and the rest of the whatsapp CLI's Go deps) current through
+  # a reviewed PR gated by CI, never a runtime @latest float: the launcher pins
+  # whatsmeow deliberately because a mid-session bump can log the device out.
+  - package-ecosystem: gomod
+    directory: /agent/skills/whatsapp/cli
+    schedule:
+      interval: weekly
+    groups:
+      whatsapp-deps:
+        patterns: ["*"]
+
   - package-ecosystem: github-actions
     directory: /
     schedule:


elyxlz:
**Fixes:** nothing linked (references #1073 as motivation, no closing keyword). Adds a `gomod` Dependabot entry for the whatsapp CLI so whatsmeow bumps arrive as CI-gated PRs instead of a manual `whatsapp update-deps`.

**Blocking:** none.

**Non-blocking**
- `.github/dependabot.yml:53` groups all Go deps daily while every other ecosystem is weekly; whatsmeow ships frequently, so expect a near-daily PR stream. Justified in the body, but a maintainer may prefer weekly to match the rest.
  Proof: `interval: daily` here vs `weekly` at lines 5/13/21/54.
- `agent/skills/telegram/cli` and `agent/skills/otp/cli` are also Go modules and stay uncovered. Deliberate per the Scope note, flagging so the omission is a choice, not a miss.
  Proof: `ls agent/skills/{telegram,otp}/cli/go.mod` → both exist, neither in dependabot.yml.

**Verdict:** MERGE — attacked on YAML validity, directory/ecosystem correctness, and the pin-vs-autobump reasoning; all held. A bump lands on master then reaches the fleet at the next daemon rebuild via upstream sync, the same controlled point the deliberate pin protects, so it does not reintroduce the mid-session logout the launcher guards against.

---

PR #2093, opened 2026-08-16
feat(user-context): report device timezone and location to the agents

emipasca:
## Summary

The agent's timezone drives reminders, dreams, and the maintenance window, but nothing told it when the user travelled. This PR makes each of the user's devices report where it is, has vestad store that per device and tell the agents when it changes, and gives the agent a tool to read it on demand.

- **Store, one owner** (`device_registry.rs`): each device carries `timezone`, `position {latitude, longitude, accuracyM, place{city,region,country}}`, `positionAt`, written by `report_context` and projected on the `devices` delta (additive; `min_supported` unchanged).
- **Two carriers, one write**: the `client_context` frame gains `timezone` + `position` (web/desktop report the zone on each focus edge; mobile reports zone + position on each foreground edge), and `PUT /devices/{device_id}/context` (client auth) carries the mobile background poll.
- **Decisions, one owner** (`user_context.rs`): a report whose zone differs from a serving agent's own zone drops one snoozed `source=vestad type=user-timezone` notification per zone (reset when the agent's own zone changes); a position whose macro place changed, or moved 25 km with no place to compare, drops one `type=user-location` carrying the place and the coordinates. vestad never writes the agent's timezone: the agent decides.
- **Agent read path**: `GET /agents/{me}/devices` (agent token, self-scoped) behind the new `user_devices` MCP tool; the `timezone` skill explains both notifications, the tool, and that WhatsApp/Telegram carry no zone.
- **Mobile**: `expo-localization` (zone read live, no cache), `expo-location` (fresh fix + OS reverse geocode in the foreground, last-known fix in the background), `expo-background-task` poll registered from the presence reporter (reads the stored session, never signs out), behind an off-by-default **Share device location** toggle in Settings > Privacy. Timezone is always shared.

Spec: `docs/superpowers/specs/2026-08-16-user-context-design.md`.

## Test plan

- [x] vestad unit: registry report/projection/round-trip, decision tables (differs, told, agent already there, tap down, place vs distance), payload shapes, `client_context` parse, `serving_timezones`; fixtures regenerated
- [x] vestad Docker (`tests/server/sync.rs`): frame -> `devices` delta -> agent read path -> `user-timezone`/`user-location` files in the container -> HTTP carrier keeps the position
- [x] `@vesta/core`: frame builder, socket report + reconnect replay, parser, `devicesEqual`
- [x] web: PresenceReporter reports the zone on focus edges only
- [x] mobile: device-context reader (fresh vs last-known fix, permission, no fix), background task (defineTask at import, PUT shape + bearer, no-ops, registration gating), preference model
- [x] agent: `user_devices` tool renders / reports an unreachable vestad; `fetch_user_devices` no-ops without identity
- [x] `check.sh guards|app-core|app-web|app-desktop|app-mobile|agent` locally; vestad clippy + unit tests natively
- [ ] CI: Docker sync scenario, iOS native compile with the new Expo modules

https://claude.ai/code/session_01HQZiPQqEPG5snjj6wA8Bdy


elyxlz:
**Fixes:** nothing linked, adds a device timezone/location signal so agents learn when the user travels (store in `device_registry.rs`, decisions in `user_context.rs`, agent read path via the `user_devices` tool). Worth carrying.

**Blocking:** none. Attacked the told-state dedup/reset, the parse rejects, the two auth tiers, the reconnect replay, and the snooze path; each held.

**Non-blocking**
- `apps/mobile/src/device-context/device-context.ts` + `vestad/src/device_registry.rs:report_context`: disabling "Share device location" stops new position reports but never clears the stored one, so the last place stays in `devices.json` and readable via `user_devices` indefinitely.
  Proof: a zone-only report keeps the stored position (`zone_only_report_keeps_the_stored_position`); no code path clears `record.position`.
- PR body cites `docs/superpowers/specs/2026-08-16-user-context-design.md`; it is not in the branch or on master.
  Proof: `git ls-tree pr2093 docs/superpowers/specs/` → file absent, only the 07-05 and 08-15 specs present.

**Verdict:** MERGE, additive protocol (`min_supported` unchanged), one-owner decision module, thorough tests; the only gaps are a stale-position-on-disable nuance and a dead spec link.

---
CI: core/web/mobile/desktop, Rust build, JS/TS, Python all green; vestad unit+Docker, agent tests, integration, and iOS compile still pending (nothing red).

---

emipasca:
## Polish pass (`bf87f2f5`)

**Pass 1, review (`/code-review high --fix`).** Applied:
- `vestad/src/user_context.rs`: told-state keyed by `(agent, device)` (two phones in different places no longer ping-pong `user-location` forever); location payload is flat scalars (`place`, `latitude`, `longitude`, `accuracy_m`), since the agent renders notification fields as attribute text and a nested object arrived as a Python dict repr; notification file names carry milliseconds so two reports in one second do not overwrite each other; a resync replay (`focused && !resync`) stores but never notifies; `PUT /devices/{id}/context` for an unknown device answers 404 instead of minting a phantom `kind: unknown` roster entry.
- `vestad/src/device_registry.rs`: `position: null` retracts a stored position (turning sharing off now clears it from the store, roster, and `user_devices`); `report_context` republishes only on a real change; `is_empty` counts the context facet so `drop_push` no longer deletes a phone's stored zone/position.
- `vestad/src/sync/protocol.rs`: `ClientContext` flattens `DeviceContext` (one shape for the frame, the PUT body, and the store).
- mobile: the background poll never refreshes the session (an expiring token skips the poll; the app holds the live refresh token, and a rotation from the poll stranded it two generations behind); guarded permission read, `getCalendars()[0]?.`, `FRESH_FIX_TIMEOUT_MS` on the fresh fix, `maxAge` on the last-known fix, `.catch` on the foreground read, try/catch + `showError` on the settings switch; sharing off sends `position: null`.
- core: `DeviceContext.position` is `DevicePosition | null`; `nullableStr` for the optional strings; `parsePosition` unexported.
- agent: one `_agent_session(token)` builder in `vestad_client.py`; tighter `user_devices` tool description; timezone SKILL.md updated to the flat payload.
- Reverted from the review's output: a `deviceTimezone()` helper it extracted into `apps/web/src/lib/device-identity.ts` and used from `NewAgent`, both files outside this PR's changed set. `PresenceReporter` keeps the inline `Intl` read.

**Pass 2, simplify (4 lenses).** Applied: `ClientContextFrame extends DeviceContext` and the builder spreads the context (mirrors the Rust flatten); `serving_timezones()` returns a `BTreeMap` (`AgentZones`) that both told-state decisions take, so the `Vec` copy, both `sort`s, and a test helper go; the bare `bool` became a typed `UserPresence { AtDevice, StoreOnly }`; the payload builders take one pre-formatted timestamp and return a `Value` (no per-agent `Result` plumbing or clone); `report_context` returns the device label directly (the `DeviceIdentity` pass-through struct is gone); `UserContext::default()` instead of a `new()` wrapper; the sync handler `mem::take`s the context out of the frame instead of cloning; the mobile module drops the `Geocoded` duplicate of `DevicePlace` and its in-file-only exports; the test harness `agent_devices` reuses `proxy_get(ProxyAuth::AgentToken)`.
Skipped (behavior or blast radius): coarse-comparing GPS jitter in the republish dedup; skipping the foreground fresh fix / geocode when the phone has not moved; parallel `deliver` fan-out; a generic `nullable<T>` for the parser's pre-existing optional fields; moving the mobile PUT onto `createHttpClient` (the raw `fetch` keeps the "never rotate" rule explicit).

**Pass 3, prompt sweep.** `agent/skills/timezone/SKILL.md`: the `user-timezone` bullet now says the once-per-zone memory starts over when the agent changes its own timezone, a mechanic the agent will otherwise read as a repeat. The description already reads as when-to-use; the tool description was tightened by pass 1.

**Pass 4, AGENTS.md.** The user-context paragraph was re-synced with the final code (per (agent, device) told-state, `null` retraction, identity only from attach, no background refresh, resync stores only, scalar fields). One pre-existing clause the PR made stale was fixed: "built to carry more over time (e.g. location)" now that location rides the channel.
Flagged, not written: notification file naming is now decided in three call sites with two conventions (`{kind}-{secs}` for rename/presence in `serve.rs`, `{kind}-{millis}` here); letting `drop_notification` mint the name from a `kind` would give it one owner but changes shipped presence/rename behavior. Same for a shared vestad-notification envelope builder (four hand-assembled `{timestamp, source, type, interrupt, message}` shapes now).

**Verify.** `guards`, `app-core`, `app-web`, `app-desktop`, `app-mobile` green. `agent`: ty green, PR tests green; only the macOS bash-3.2 shell tests fail locally (`test_start_daemons`, `test_upstream_sync`, `test_whatsapp_launcher`, none touched by this PR). vestad via the macOS wrapper: clippy `-D warnings` clean, 477 unit tests pass (only the two known Linux-only `upstream::tests` fail), `check --tests` compiles, `api_contract` fixture unchanged.

emipasca:
Follow-up to the polish pass: the background poll refreshes the session again. Skipping the report once the access token is within its refresh window (6h TTL) would silence the closed-app poll a few hours after the last app open, which is the case it exists for. The refresh writes the rotated tokens to SecureStore and `SessionProvider` now adopts a same-gateway rotation on foreground (`rotatedStoredConnection`), so the suspended app never presents a refresh token two rotations old and the family is never revoked. Tests: the poll refreshes, writes, and reports with the rotated token; the adopt rule ignores a sign-out and a gateway switch.

PR #2095, opened 2026-08-16
epic(mobile): polish pass, Android settings sections, full visual coverage

emipasca:
One consolidated PR carrying the whole overnight mobile pass (previously stacked as #2095 -> #2096 -> #2102, folded on request). Fourteen commits, each one concern, reviewable commit by commit.

## Per-agent state preservation and UX
- Keyed per-agent holds above navigation (`src/holds/`): chat tail, composer draft + armed reply, last pager tab, and log buffer survive screen pops and agent switches; Android's pager keeps non-adjacent pages attached; pager pages mount on first visit.
- Gateway-update flow: the sheet hands off to home's live progress page when the operation starts, behind-ness is latched across socket blips, Android hardware back cannot blink blocking gates, and the home carousel stops snapping back to the last-opened agent.
- Crash and stuck-state fixes: history-seed retry with backoff, boot-splash no longer wedges on a storage failure, unhandled rejections closed, a render-phase ref hazard removed.
- Logs page fills on the boot tail replay (the "couple of lines until you scroll" bug): the inverted list snaps to the tail whenever content grows while the reader is at the tail. The 50-line fixture tail makes the agent-logs flow fail without the fix.

## Performance and hygiene
- Stable provider identities (Roster/Agent/socket memoization), keystroke-isolated chat transcript, per-palette markdown styles, memoized notification rows.
- ChatPage split from ~1510 lines into `src/agent/chat/` modules with tested models; dead code and pass-through wrappers deleted; one owner for settings sections, device labels, and provider/model derivation; shared GlassSurface; `MessageContextMenu` generic. A polish pass (review + simplify + docs) is included.

## Android settings sections
- Every agent-settings section (provider, voice, notification rules, files, host access, backups) serves on Android as Material form sheets; the platform gate is deleted. Known quirk on the tap path tracked as #2105.

## Visual harness
- Every app view is a captured scenario on both platforms: the seven iOS-only scenarios flipped, plus home-loading, agent-chat-empty, settings-devices, and app-behind (which caught and fixed a real production bug: AppBehindScreen sat invisible behind a never-hidden boot splash).
- Both capture runners gain `--gentle`: one simulator shard and utility-QoS children, so captures can run behind interactive work.

## Verification
`./check.sh app-mobile` green (lint, tsc, 299 vitest tests, clean prebuild). Multiple full iOS capture runs plus two Android emulator runs; new states inspected pixel-level. The one red visual flow is pre-existing on master (#2097, iOS-only) and blocks gallery publishing until fixed.

https://claude.ai/code/session_01367FvVt2ot8i47NQPYQ1Ym


elyxlz:
**Fixes:** nothing linked. Polishes `apps/mobile`: per-agent state holds, gateway-update handoff, provider identity/perf, and a ChatPage split. Worth carrying.

**Blocking:** none.

**Non-blocking**
- Scope: the PR bundles four concerns plus a large move (`ChatPage.tsx` -1247 / new `chat/` modules, dead-endpoint deletion) in one changeset, against CLAUDE.md "One concern per PR / isolate renames from logic". Cohesive within mobile, but hard to bisect if one lens regresses; maintainer's call.
  Proof: `src/agent/ChatPage.tsx` 71/-1247 alongside `holds/`, `controller/` latch, and `endpoints.ts` -26 in the same PR.
- `holds/keyed-hold.ts:12` `read` does not re-insert, so eviction is write-order, not the "LRU" the header claims; a read-hot, write-cold cell can be dropped first.
  Proof: none, read only; harmless at MAX_HOLD_CELLS=12 vs typical agent count.

Traced and held: the render-phase `behindLatch` (ControllerProvider), seed/log capped backoff, RosterProvider field-equality memo, pager lazy-mount + `offscreenPageLimit`, composer hold key captured by closure. New imports (`agentIsConnectable`, `HeaderFade`, `useGatewayOperation`) all resolve.

**CI:** `merge-gate-ci` green, `Apps · mobile` green; MERGEABLE. Release-pipeline jobs show "skipping" by design (prerelease-only).

**Verdict:** MERGE, attacked the holds, latch, backoff, and pager lazy-mount and they held; only reviewable concern is bundled scope.

If you want it tightened before merge, a maintainer can run `polish-pr 2095`.

---

emipasca:
Polish pass on this PR (one commit, a9b6a770). Four passes, summarized below.

## Pass 1, review (10 confirmed findings, all fixed)

- **LogsPage** gated the stream on agent connectability, so a crashed or stopped agent's last lines were unreachable; it now gates on gateway reachability alone (vestad serves a stopped container's log tail and ends the stream cleanly).
- **LiveLogs** resumed held lines with `?tail=0`, silently losing everything logged while unmounted; each mount now starts empty and replays the tail, and the hold backs only the unreachable-gateway fallback.
- **useAgentSocket**'s history-seed retry loop outlived socket close; retries are now scoped to an open socket.
- **gateway-update-gate**: the backdrop formula re-encoded the model's precedence and could dim the screen with no sheet; the model now returns `{action, backdropBlocked}` from one input set, table-tested.
- **ChatPage and LiveLogs are keyed by holdKey**, so a gateway switch under a retained screen remounts and reseeds instead of bleeding stale composer/log state into the new key's cell.
- **selectPage** marks every page an animated multi-page tab jump slides across as visited, so intermediate panes no longer flash blank.
- **onLoadEarlier** passes the stable `socket.loadMore` directly instead of a wrapper that re-identified on every socket tick.
- **sendCurrentInput** reads the armed reply from a ref, so a reply armed mid voice session is what the turn-end send uses.
- **ChatEvent time labels memoized** (Intl formatter per row per render), **NotificationSeparator hoisted** to module scope.

## Pass 2, simplify (four parallel reviewers: reuse, simplification, efficiency, altitude)

- **RosterProvider**: the new gateway selector re-declared `RosterSnapshot`'s five gateway fields under renamed keys and remapped them field-for-field; it now emits `Omit<RosterSnapshot, "agents">` and `fresh` is a spread.
- **Logs hold narrowed** to `KeyedHoldStore<LogLine[]>`: the `nextId` field pass 1 orphaned is gone along with the per-line wrapper allocation.
- **Pager container** read `useAgent()` just for the name, subscribing the whole pager shell (tabs, fades, ~15 callbacks) to every chat tick; it reads the route param instead, as AgentProvider does.
- **chat-markdown**: the module-level `Map` cache keyed on palette identity became two eagerly built style-set consts (the palette space is exactly `darkColors`/`lightColors`).
- **QuotedBlock** (new small component) owns the quoted-frame look; the composer's reply preview and the transcript's markdown blockquote both render it, deleting the odd `chat-event` to `chat-composer` style import.
- **ComposerSurface** was a pass-through after the GlassSurface extraction; inlined into ChatPage. **GlassSurface** sheds two props no caller used.
- **Settings sections**: `sectionAvailability` now carries the found section, and the detail screen renders from a `Record<AgentSettingsSectionKey, ComponentType>`, so TypeScript forces an entry per model key and the tested `unknown` state finally has a reader (the seven-branch if chain is gone).
- **Gateway-behind latch**: the decision moved from ControllerProvider into the tested gate model (`latchedGatewayBehind`, five table cases); the provider keeps only the render-phase state carry.
- **Visual harness** reuses the real `AgentHoldsContext` and re-exports the real hook instead of hand-copying both.
- **chat-transcript**'s content-container style memo keys on emptiness, not raw row count, so it survives the whole conversation.

Verified but deliberately skipped: extracting `openingPage`/`pagesToMount` into `pager-model.ts` (file is outside this PR's diff; worth a follow-up since the drag and tab-jump paths spell the mount decision twice), a shared retry/backoff helper for `subscribeLogs` and the chat seed (real duplication, but a cross-module design call), the Android `BackHandler` living in `SheetGateScreen` rather than the gates (defensible as named), the pre-existing dead `hasMore` on the chat socket value, and the deeper fix of keeping behind-ness inside `@vesta/core`'s socket state machine (out of this PR's scope; the mobile latch now at least lives in the tested model).

## Pass 3, prompt sweep

No prompt surfaces in the diff; skipped.

## Pass 4, AGENTS.md freshness

- The Mobile App bullet's "stale-while-reconnecting roster/chat holds" now names the generalized mechanism: the roster hold plus the keyed per-agent holds in `src/holds/` (chat tail, composer draft, open page, log buffer), surviving screen pops as well as backgrounding.
- The gateway self-update paragraph said "mobile shows a single progress sheet"; the sheet now hands off to home, which renders the live progress, and the prose says so.
- Considered and not written: a dedicated architecture paragraph for the holds subsystem. The bullet-level mention seems the right altitude; flagging in case you want more.

Verification: `./check.sh app-mobile` and `./check.sh guards` both green (305 vitest tests, tsc, lint, clean prebuild).

https://claude.ai/code/session_01367FvVt2ot8i47NQPYQ1Ym

emipasca:
Branch renamed `refactor/mobile-polish` -> `epic/apps`, which auto-closed this PR. Continues as #2126 (same commits).

PR #2103, opened 2026-08-17
refactor(vestad): one owner for vestad-authored agent notifications

emipasca:
## Summary

Follow-up to #2093, the two hygiene items its polish pass flagged.

Every notification vestad writes into an agent's intake (`rename`, `user-presence`, `user-timezone`, `user-location`) is the same envelope, `{timestamp, source: "vestad", type, interrupt}` plus its own fields, under `{kind}-{millis}.json`. That envelope was hand-assembled four times (`serde_json::json!` blocks in `serve.rs` and `user_context.rs`) and the file name minted at three call sites in two conventions (seconds for rename and presence, milliseconds for the user-context kinds).

New `vestad/src/agent_notification.rs` owns both: `AgentNotification {kind, interrupt, fields}` with `envelope(epoch_secs)`, `file_name(kind, millis)`, and `drop(docker, agent, notification)`, plus the four builders (`rename`, `user_presence`, `user_timezone`, `user_location`). Every kind now gets the millisecond name, so two notifications minted inside one second never overwrite each other, and a fifth kind adds fields and nothing else. `serve.rs` loses its two payload builders and two drop wrappers; `user_context.rs` loses its private `deliver` timestamping.

Net: 166 insertions, 275 deletions. No wire or file-format change beyond the millisecond in the rename/presence file names.

## Test plan

- [x] `agent_notification` unit tests: envelope shape, file name, rename fields + interrupt, presence snoozed + client name
- [x] `user_context` payload tests moved onto the builders (timezone, location with and without a place)
- [x] Docker drop tests (`docker.rs`, `#[ignore]`) call the one `drop`
- [x] vestad clippy `-D warnings` + unit tests natively; `check.sh guards`
- [ ] CI: vestad Docker tests + integration

https://claude.ai/code/session_01HQZiPQqEPG5snjj6wA8Bdy


elyxlz:
**Fixes:** nothing linked, a follow-up refactor to #2093 folding vestad's four agent-notification builders and three filename call sites into one `agent_notification.rs` owner. No behavior change beyond seconds→millis in the rename/presence filenames.

**Blocking:** none.

**Non-blocking**
- `vestad/tests/server/rename.rs:91` comment still points readers at `serve::rename_notification_payload`, which this PR deleted (moved to `agent_notification::rename`); the cross-reference is now dead.
  Proof: `grep rename_notification_payload vestad/src` → no match after the diff.
- `agent_notification.rs:59` `fields()` silently returns an empty map on any non-object `Value`, dropping content instead of failing. Unreachable today (all callers pass `json!({...})`), but a lossy fallback.
  Proof: none, read only; all four builders pass object literals.

**Verdict:** MERGE — net consolidation, all kinds now go through one `drop`/`envelope`/`file_name`; agent sorts intake lexicographically by an opaque stem (`loops.py:53`) so the 13-digit millis rename is safe. Attacked on filename readers, the four builders' completeness, and the best-effort failure path; held. Clippy `-D warnings` and the 13 unit tests pass locally.

---

PR #2104, opened 2026-08-17
fix(tasks): stamp checkpoints on create and keep the offset on daily reminders

emipasca:
## Summary

Two bug fixes the polish pass on #2061 found and pushed to `epic/tasks-2` after that PR had already been squash-merged, so they never reached master (`f35fcc33` on the merged branch, cherry-picked here).

- **tasks**: `tasks create --in-days 1` fired "due in 1 day" on the daemon's first tick, because `create` did not stamp `checkpoint_fired_through` the way every other due-date write does. Now it does. Regression test in `test_checkpoints.py`.
- **reminders**: `reminders create --at "21:30+02:00"` silently dropped the offset from a bare daily time. Fixed, with a test in `test_interface.py`.

## Verify

- tasks CLI suite: 146 passed
- reminders CLI suite: 146 passed
- `./check.sh guards`: green

### DIFF
--- a/agent/skills/reminders/cli/src/reminders_cli/commands.py
+++ b/agent/skills/reminders/cli/src/reminders_cli/commands.py
@@ -238,7 +238,9 @@ def _parse_local_dt(datetime_str: str, timezone_str: str | None, *, allow_bare_t
         except ValueError:
             pass
         else:
-            return datetime.combine(datetime.now(local_tz).date(), bare, tzinfo=local_tz)
+            # combine keeps the bare time's own offset when it carries one, so it lands below like any aware datetime.
+            parsed = datetime.combine(datetime.now(local_tz).date(), bare)
+            return parsed.astimezone(local_tz) if parsed.tzinfo is not None else parsed.replace(tzinfo=local_tz)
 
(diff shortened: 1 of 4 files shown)


elyxlz:
**Fixes:** nothing linked (no closing keyword). Carries two real bugs the polish pass on #2061 found after that PR merged: `tasks create --in-days 1` firing "due in 1 day" on the first tick, and `reminders create --at "21:30+02:00"` dropping the offset. Master still has both.

**Blocking:** none.

**Non-blocking**
- `tasks/commands.py:211` stamps `checkpoint_fired_through=now` at creation. Attacked completeness, the tests, and the failure/edge paths; all held. Creating an already-past-due task fires nothing on creation, but old code did the same (the at-due rung is `<= created_at`, dropped by `checkpoint_times`), so no regression.
  Proof: reverted each fix in a worktree → `test_creating_a_task_fires_no_rung_that_lands_at_creation` and `test_daily_reminder_bare_time_keeps_its_own_offset` both went red; unreverted, tasks 146 passed, reminders 146 passed.

**Verdict:** MERGE — two correct one-line fixes, each proven by a regression test that fails without it.

---

PR #2106, opened 2026-08-17
fix(migrations): repoint tasks remind notes and restart into the reminders skill

emipasca:
## Summary

Follow-up to #2061. The `2026-08-reminders-skill-split` migration installs, activates, and starts the reminders skill, but on an upgrading box two things stayed unconverged:

- **Agent-authored `tasks remind` references.** The command no longer exists and the tasks CLI has no alias, so any note in `MEMORY.md` or a skill file the agent wrote is an instruction that fails with a usage error. New step 8: grep `~/agent/MEMORY.md` and `~/agent/skills` (excluding `cli/`) and rewrite each hit (`tasks remind ...` to `reminders create ...`, `list|snooze|update|delete` map one to one, `--task <id>` is dropped and the task goes in the message). Stock skill text is already clean on master, so hits are the agent's own.
- **Skill discovery.** `skills-activate` only lists the skill; the boot entrypoint links it into `~/.claude/skills`. New step 10 calls `restart_vesta` after the mark step, matching the precedent in `2026-07-manual-upstream-review` and `2026-07-workspace-resync`.

The migration is unreleased (latest release v0.2.3 predates #2061), so it is edited in place rather than shipped as a second migration.

## Verify

- `uv run --project core pytest tests/test_migrations.py`: 30 passed
- `./check.sh guards`: green

### DIFF
--- a/agent/core/migrations/2026-08-reminders-skill-split.md
+++ b/agent/core/migrations/2026-08-reminders-skill-split.md
@@ -62,6 +62,20 @@ If that list holds a recurring reminder whose message is about updating contacts
 
 Look at the end of `~/agent/skills/tasks/SKILL.md`. If it still has a `### Reminder Patterns` section holding notes you wrote (not just the bare placeholder), move that body under `### Reminder Patterns` at the end of `~/agent/skills/reminders/SKILL.md`, then delete the section from the tasks file. If there is no such section, or only the placeholder, nothing to do.
 
-### 8. Mark this migration applied
+### 8. Repoint your own notes
+
+There is no `tasks remind` command: `tasks` refuses it with a usage error. Any note of yours that still spells it out is an instruction you will follow into that error, so find every one:
+
+```bash
+grep -rn 'tasks remind' ~/agent/MEMORY.md ~/agent/skills --exclude-dir=cli --exclude-dir=.venv
+```
+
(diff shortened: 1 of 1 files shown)


elyxlz:
**Fixes:** Follow-up to #2061 (no closing keyword); converges two loose ends the reminders-split migration left on upgrading boxes. Both hold.

**Blocking:** none.

**Non-blocking**
- `2026-08-reminders-skill-split.md:73` grep `'tasks remind'` also substring-matches prose like `tasks reminder`/`tasks reminders`, so the agent may see a non-command hit to judge.
  Proof: `grep 'tasks remind'` matches `tasks reminders` too; agent-judgment only, not a command to rewrite.
- CI still pending (Agent typecheck/test, vestad); nothing red yet. Watch those before merge.

**Verdict:** MERGE — unreleased migration (not in v0.2.3, added by #2061 on 2026-08-17), so the in-place edit reaches nobody who ran it; `reminders create|list|snooze|update|delete` all exist and map 1:1, `tasks` has no `remind` subcommand, stock skill text is clean on master (`git grep 'tasks remind'` excluding cli → 0) so step 8 only touches the agent's own notes, and `restart_vesta` after the mark step matches the two cited precedents. Step renumber is clean, no stale cross-refs. Migration tests pass locally.

---

PR #2108, opened 2026-08-17
feat(provider): show the signed-in account in provider settings

emipasca:
## What

Adds a "Connected account" panel to each agent's provider settings (web + mobile) that shows the account the agent is signed into for its active provider: name, email, plan, organization, and member-since date.

The data is fetched live, per agent, per provider, and rides on the existing `/usage` resource rather than a new endpoint: the agent's `Usage` gains an optional `account` field, so it is one fetch, one hook, one refresh, one failure state, and one section in the UI.

## How

- **agent/core/provider.py**: new `Account` dataclass (all fields optional) and an `account` field on `Usage`. `get_usage` now also resolves the signed-in account per provider:
  - Claude: `GET /api/oauth/profile` (name, email, plan from `has_claude_max`/`has_claude_pro`, org, created_at).
  - OpenRouter: the key endpoint's `label` and `is_free_tier` (single call, already fetched for usage).
  - Z.AI / Kimi / OpenAI: no identity endpoint, so `account` is null (same as usage already returns empty for them).
  - The account fetch is **best-effort**: a failed profile fetch returns null and never blanks the usage meters that fetched successfully.
- **agent/core/api.py**: `GET /usage` now serializes the account alongside meters/credits (additive; no contract break, `min_supported` unchanged).
- **apps/web** + **apps/mobile**: `Usage` gains `account`; a new `AccountSection` renders the non-null identity rows above the usage bars.

## Why this shape

`Usage` and account identity are the same kind of thing (live, per-provider, external, same auth, same failure mode), so unifying them removes concepts rather than adding a parallel `/account` stack. The type keeps the name `Usage` because `ProviderStatus` is already taken by the auth-status dataclass.

## Tests

- `agent/tests/test_provider.py`: Claude account mapping from a profile fixture, OpenRouter account mapping, best-effort (profile fails, usage survives, account null), and the existing usage-fetch-error propagation.
- Web/mobile covered by tsc + existing component suites; mobile visual-QA fixture seeded with a sample account so the new section renders in the gallery.

### DIFF
--- a/agent/core/api.py
+++ b/agent/core/api.py
@@ -3,7 +3,7 @@
 Routes:
   - WS   /ws                   bidirectional event bus
   - GET  /history              paginated event history (cursor optional), or full-text search with ?q=
-  - GET  /usage                normalized, provider-agnostic plan usage
+  - GET  /usage                normalized, provider-agnostic plan usage plus the signed-in account
   - GET  /status               operational readiness: {authed, setup_complete, boot_complete} (vestad polls this)
   - GET  /config               prefs + notification_rules (personality, timezone, seed_context, operational)
   - PUT  /config               update prefs and/or notification_rules (provider is set via /provider)
@@ -203,7 +203,8 @@ async def _history_handler(request: web.Request) -> web.Response:
 
 
 async def _usage_handler(request: web.Request) -> web.Response:
-    """Report normalized, provider-agnostic plan usage for the agent's active provider."""
+    """Report normalized, provider-agnostic plan usage plus the signed-in account for the active
+    provider."""
     config = request.app["config"]
     try:
         usage = await get_usage(config)
--- a/agent/core/provider.py
+++ b/agent/core/provider.py
@@ -414,9 +414,23 @@ class UsageCredits:
 
 
 @dc.dataclass
+class Account:
+    """The account behind the active provider, best-effort. A field the provider does not expose
+    stays None, and a provider with no identity endpoint (a key or subscription plan) reports an
+    all-None account or none at all."""
+
+    name: str | None
+    email: str | None
+    plan: str | None
+    organization: str | None
+    created_at: str | None
+
+
+@dc.dataclass
 class Usage:
     meters: list[UsageMeter]
     credits: UsageCredits | None
+    account: Account | None
 
 
 class UsageError(Exception):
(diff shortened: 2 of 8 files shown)


elyxlz:
**Fixes:** nothing linked. Adds a "Connected account" panel (name, email, plan, org, member-since) to provider settings, riding the existing `/usage` resource instead of a new endpoint. Worth carrying.

**Blocking:** none.

**Non-blocking**
- `provider.py:_claude_account` reads name/plan/created_at from a `/api/oauth/profile` shape I cannot reach from here; if the endpoint or key names are wrong the panel silently shows nothing (caught as `UsageError` → `None`), so a broken assumption degrades to an empty section rather than a crash.
  Proof: none, read only (network endpoint).
- `provider.py:496` `_claude_account` is a second serial HTTP call on every Claude `/usage` fetch, doubling the request and adding its latency to each refresh.
  Proof: `get_usage`→`_claude_usage` awaits usage then `await _claude_account(headers)`.
- `ProviderCard/index.tsx:311` account hides while `loading||error`, so it flickers out on every refresh; mobile keeps it via cached `usage.data`.
  Proof: none, read only.

**Verdict:** MERGE, attacked completeness (all 3 `Usage(...)` and the one TS `Usage` literal carry `account`; tsc would fail otherwise), the failure path (profile 401 → `account:null`, meters survive, pinned by `test_get_usage_account_best_effort_when_profile_fails`), and readers; 57 provider tests pass, ruff clean.

---

emipasca:
@vestabot apply the non blockign stuff

elyxlz:
Applied the two actionable non-blocking findings in `7e1326ce`:

- **Serial profile call** → `provider.py:_claude_usage` now fetches usage and the profile concurrently via `asyncio.gather`, so the account no longer adds a round-trip to each `/usage` poll. Usage errors still propagate; a failed profile still degrades to `account:null`.
- **Web flicker** → dropped the `!loading && !error` guard on the account section (`ProviderCard/index.tsx:311`); the hook keeps the previous `usage` during a refresh, so it now persists instead of blanking, matching mobile.

The third note (the `/api/oauth/profile` shape is unverifiable from here) is not something I can fix in code, it needs a real Claude token to confirm the field names; left as-is since a wrong shape degrades to an empty panel rather than an error.

57 provider tests pass, ruff + ty clean. Web edit is a one-token removal with `loading`/`error` still consumed by `UsageBody`, so no tsc impact.

Verdict: MERGE, the feature was already sound and these two changes remove the added latency and the refresh flicker without touching the failure semantics.

emipasca:
@vestabot fix ci

elyxlz:
**CI:** the one red job was `Agent · Typecheck and test`, and it was **cancelled at the 20-min job cap** (step conclusion `null`, not `failure`, so no test asserted false). That same suite passes in ~14 min on master, so this is a timeout, not a defect. I confirmed the cause is not this PR: only `test_provider.py` (fully mocked, `_fetch_usage_json` patched, 57 pass in <1s) and the `/usage` handler call the changed path, so the `asyncio.gather` diff cannot hang the suite. I re-ran the failed jobs; the agent job is re-running now (run 32023303807). Watch that job and `merge-gate-ci`, which failed only because it gates on it.

No code change needed for this: the timeout is a slow-runner/flake in the agent suite (the tmux-bound cc_sdk transport tests are the timing-sensitive part), not something the diff introduced.

Verdict: MERGE, the sole red check is a 20-min timeout on a suite the diff does not touch at runtime, cleared by the re-run now in flight.

PR #2109, opened 2026-08-17
fix(whatsapp): smooth the Double Tick setup flow

emipasca:
From a retrospective on a real Double Tick WhatsApp setup, four friction points, each fixed here. (A fifth finding, an `app-chat send` false failure, is a separate subsystem and ships in its own PR.)

## Changes

**F1: the user holds the credentials, so ask them.** `SETUP_DOUBLETICK.md` previously modeled only an operator setting `DOUBLETICK_API_URL`/`DOUBLETICK_API_KEY` outside chat, which left the agent no path when the user held the key on mobile. Now: if the two values are not already in the environment, the agent asks the user for both, then connects with them set inline for that one command (the daemon persists them in owner-only state). `SETUP.md` step 4 now offers Double Tick by asking whether the user has credentials.

**F2: the unpaired notification no longer asserts a wrong `--source`.** On a first link whose account source the daemon cannot determine (no credentials in its own environment, none persisted yet), the reconnect hint listed `--source self-managed` as a guess. It now emits `whatsapp connect --source <vesta-cloud|doubletick|self-managed>` for the agent to fill in. Confident cases (persisted source, or clear env) and the logged-out/relink paths are unchanged, with their tests intact.

**F3: `--opener -` reads the opener from stdin.** `connect` now accepts the same `-`/heredoc convention `send --message -` uses, so an opener keeps its apostrophes and newlines instead of being reworded to survive inline single-quoting.

**F4: named-instance setup is documented.** `SETUP_DOUBLETICK.md` now shows how to link a second account, or one when the default instance is held or in use, on its own named instance.

## Tests

- `TestFreshUnknownSourceUnpairedAsksToChooseSource` (F2): a fresh, credential-less first link emits the choose-source hint, not `self-managed`.
- `TestResolveOpenerStdinReadsHeredocBody` / `...LeavesALiteralOpenerAlone` (F3).
- `check.sh whatsapp` green (lint + build + test); `check.sh guards` green.

### DIFF
--- a/agent/skills/whatsapp/SETUP.md
+++ b/agent/skills/whatsapp/SETUP.md
@@ -31,10 +31,10 @@ pick the source and pass it explicitly as `--source`; the CLI never guesses.
 3. **`DOUBLETICK_API_URL` and `DOUBLETICK_API_KEY` both set?** Run
    `whatsapp connect --source doubletick --opener '<text>'`. See the
    [Double Tick WhatsApp account](SETUP_DOUBLETICK.md) guide.
-4. **Otherwise, ask the user** which they want, then run that one command:
-   - Double Tick credentials in hand:
-     `whatsapp connect --source doubletick --opener '<text>'`. A provisioning
-     service that hands the agent a ready headless account. See the
+4. **Otherwise, ask the user** which they want:
+   - **Double Tick**, a provisioning service that hands the agent a ready headless
+     account: ask whether they have Double Tick credentials. When they do, get the
+     API URL and key from them, then follow the
      [Double Tick WhatsApp account](SETUP_DOUBLETICK.md) guide.
    - Their own account: `whatsapp connect --source self-managed`. The user
      supplies a dedicated WhatsApp account on their own SIM (a second account on
--- a/agent/skills/whatsapp/SETUP_DOUBLETICK.md
+++ b/agent/skills/whatsapp/SETUP_DOUBLETICK.md
@@ -6,48 +6,54 @@ Use this when a non-cloud agent points at such a service.
 
 ## Prerequisites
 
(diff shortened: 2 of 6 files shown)


elyxlz:
**Fixes:** nothing linked (no closing keyword). Clears four Double Tick setup friction points from a real retrospective. Worth carrying.

**Blocking:** none.

**Non-blocking**
- `SETUP_DOUBLETICK.md` now tells the agent to ask the user for the key and run it inline (`DOUBLETICK_API_KEY='...' whatsapp connect ...`), dropping the prior explicit caution. The key now lands in chat and command history. Deliberate F1 tradeoff, your call.
  Proof: the removed line "Never ask the user to paste it into a conversation" / "keep the key out of chat, logs, commits, and command output" contradicts the new inline-env example.
- Bundles four concerns (two docs, two code) against one-concern-per-PR.
  Proof: none, read only.
- Could not execute the Go tests: the whatsapp cli build needs whisper.cpp native headers, absent here.
  Proof: `go test ./ -> whisper.h: No such file`. Traced statically: the new F2 branch fires only on the fresh no-cred no-persist catch-all (`!priorLink && !managed && configError=="" && !confident`); every env/persisted/priorLink test routes elsewhere and is untouched. Doc anchors (`#named-instances`, `#residential-proxy-lease`) exist. Inline-env connect reads `managedConfigFromEnvAndState()`, so F1 works.

**Verdict:** MERGE, additive and correctly scoped; the notification hint stops asserting a guessed `--source`, existing tests hold, and the one cost is docs now routing the API key through chat.

---
CI: PR is MERGEABLE, not draft; I did not run CI, so `check.sh whatsapp`/`guards` stand on the PR page.

---

PR #2110, opened 2026-08-17
fix(app-chat): ack a send before the best-effort user notification

emipasca:
From the same Double Tick setup retrospective, one reliability finding, separate subsystem.

## The bug

The daemon's socket `send` handler did, in order: persist the message (assigning its `id`), echo it to live subscribers, **shell the user-notification script (best-effort, up to a 10s timeout)**, and only then reply to the client. The notification step raced the client's own 10s read deadline, so an already-durable send could return `{"error": ""}` (an empty `TimeoutError`) to the caller even though the message was stored with an id and echoed. During the WhatsApp setup this showed up as spurious send failures that forced a DB re-check.

## The fix

Write the client's ack right after persist + emit, and run the best-effort notification **after** the reply. A slow notification can no longer turn a delivered message into a reported failure. One layer, one change: the client's read window is untouched.

## Test

`test_send_acks_the_client_before_the_user_notification_finishes`: with a notification patched to block, the client still receives its `{"ok":..., "id":1}` ack (a read that waited on the notification, the old order, would time out instead). `check.sh app-chat` green (54 passed); `check.sh guards` green.

### DIFF
--- a/agent/skills/app-chat/cli/src/app_chat_cli/daemon.py
+++ b/agent/skills/app-chat/cli/src/app_chat_cli/daemon.py
@@ -201,9 +201,11 @@ async def _handle_socket_conn(state: DaemonState, reader: asyncio.StreamReader, 
             response = {"error": f"unknown command: {command}"}
 
         writer.write(json.dumps(response).encode())
-        await writer.drain()
-        # The user notification (toast + push) is best-effort and can block up to its
-        # timeout, so it runs only after the client holds its durable ack, never before.
+        # The notification runs after the ack so it never delays it, but a durable reply must
+        # still notify even if the caller died before reading the ack, so a broken pipe on the
+        # ack must not skip it.
+        with contextlib.suppress(OSError):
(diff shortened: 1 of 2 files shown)


elyxlz:
**Fixes:** nothing linked. Solves spurious app-chat send failures: the client's 10s read raced the notification shell, so a durable send could return `{"error": ""}` (a stringified `TimeoutError`). Worth carrying.

**Blocking:** none.

**Non-blocking**
- `daemon.py:205` the notification now runs after `writer.drain()`, so if drain raises (client gone before it reads the ack) the toast/push is skipped, where the old order always fired it. Narrow: the `app-chat send` invoker waits for the ack, and the reply is already durable + echoed, so the loss is only the nicety when the caller itself died.
  Proof: none, read only.

**Verdict:** MERGE, attacked the reorder: the new test times out under the old order (5.26s fail) and passes under the fix, full suite 54 passed, the moved annotation and `notify_message` cover all branches.

---

PR #2111, opened 2026-08-17
feat(skills): vesta-agent-retro, reconstruct a live agent's transcript

emipasca:
A dev skill that captures the workflow used to build the Gianfranco WhatsApp retrospective: reconstruct what a running Vesta agent actually did, its full transcript with thinking, tool calls, and tool **results**, for a retrospective or postmortem.

## Why

The obvious source, `~/agent/data/events.db`, has thinking and tool calls but records only that a tool ran, never its output. The tool **results** live only in the raw Claude session `~/.claude/projects/<slug>/<session_id>.jsonl`. That non-obvious split is the whole reason a naive `events.db` dump silently loses every tool output. The skill encodes it.

## What's here

- `extract-retro.sh` (host wrapper): reads the agent's live `session_id`, finds the `.jsonl`, runs the extractor **inside** the container where the 150MB+ file is local, streams back only the filtered slice.
- `extract-transcript.py` (in-container, stdlib only): renders a time-windowed slice as `THINKING` / `TOOL CALL [name]` / `-> RESULT`, redacting `wak_`/API-key/token secrets on the way out, with a stderr leak-check.
- `SKILL.md`: the man page, auto-loads on "retrospective on the agent", "what did <agent> do", "pull the transcript".

## Verified

Every command was run against a live agent this session:
- `extract-retro.sh vesta-lucadv-gianfranco 2026-08-17T17:03:00 2026-08-17T17:25:00` → 185 lines, **0 unredacted secrets** (the API key the user pasted mid-flow renders as `wak_REDACTED`).
- The `events.db` window-finding query (with `-i`) and the projects listing.
- Synthetic run confirms thinking/tool_use/tool_result rendering and redaction. `check.sh guards` green (ruff + shellcheck).

### DIFF
--- a/.claude/skills/vesta-agent-retro/SKILL.md
+++ b/.claude/skills/vesta-agent-retro/SKILL.md
@@ -1,0 +1,113 @@
+---
+name: vesta-agent-retro
+description: >
+  Use when you need to know what a running Vesta agent actually did, for a
+  retrospective, postmortem, or "why did the agent do X" question, and a summary is
+  not enough because you need its real reasoning and the output its tools returned.
+  Triggers: "retrospective on <agent>", "what did <agent> do", "pull/extract the
+  agent's transcript", "why did the agent <X>". On the host, needs sudo docker to a
+  running vesta-* container.
+---
+
+# vesta-agent-retro
+
+Reconstruct a slice of a running Vesta agent's session as a redacted, human-readable
+transcript, then write the retrospective on top of it. The agent runs Claude inside a
+Docker container; this pulls the real record of a time window out of that container.
+
+Paths below are relative to the repo root. The driver is
+`.claude/skills/vesta-agent-retro/extract-retro.sh`.
+
+## The one thing to know first
+
+Two stores hold the agent's history, and only one has tool output:
+
+- `~/agent/data/events.db` (SQLite) has user/assistant messages, **thinking**, and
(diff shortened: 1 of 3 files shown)


elyxlz:
**Fixes:** nothing linked. A dev skill to reconstruct a live agent's transcript with tool results (the real value: tool output lives only in the `.jsonl`, not `events.db`). Correct insight, useful tool. Not under `agent/`, ships to no fleet.

**Blocking**
- `extract-transcript.py:30` the `authorization|bearer` pattern stops the value at the first space (`[^\s'",}]+`), so `Authorization: Bearer <jwt>` redacts only the word `Bearer` and passes the token through in cleartext, while `leak-check` (line 109, scans `wak_`/`sk-` only) prints `0`. That breaks the headline claim "safe to share even when the raw stores are not" and gives a false all-clear.
  Proof: synthetic run → `Authorization: REDACTED eyJhbGciOiJIUzI1NiJ9leakedtoken` and `leak-check: 0 possible unredacted secret(s)`.

**Non-blocking**
- Leak-check covers only `wak_`/`sk-`; every generic-keyword miss is invisible to it, so `0` never means "no secrets".
  Proof: same run, the Bearer JWT above.

**Verdict:** BUGGED — redaction leaks Bearer tokens and the leak-check falsely reports clean; both are a few lines from correct.

---

Issue #2105, opened 2026-08-17
fix(mobile): Android agent-settings sheet opens scrolled to the first Compose switch


emipasca:
On Android, opening agent settings from the agent header (tap path, animated form-sheet push) presents the sheet pre-scrolled: the first `@expo/ui` Jetpack Compose switch (the "Natural chat pacing" row) sits at the very top of the viewport, with the identity card and the row's label scrolled out above. Opening the same sheet via a fresh deep link (`vesta-dev://agent/aria/settings`) starts at the top.

The mechanism fits Android's focus-driven scrolling: the Compose interop switch grabs focus when it attaches after the push animation, and the framework scrolls the focused child into view. It surfaced when #2096 made the sheet long enough to scroll (before that, the Android sheet content fit the viewport). `@expo/ui`'s Host/Switch expose no focus control and its modifier set has no focus modifier, so there is no clean product-side fix through the public API today; candidates are an upstream `@expo/ui` focusability prop or a `descendantFocusability` escape on the sheet's scroll view.

Evidence: the Android visual catalog's "Connected app screens" flow failed deterministically on both a cold and a warm emulator run at `"Natural chat pacing" is visible` while the full sheet (all #2096 sections) rendered correctly below the fold; the failed-step screenshot shows the switch pinned at the viewport top. The flow now scrolls back to the top before capturing (workaround marked in `maestro/visual/connected.yml`), so the catalog stays green while this remains a real user-facing quirk on the tap path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01367FvVt2ot8i47NQPYQ1Ym

PR #2102, opened 2026-08-17
feat(mobile): capture every app view in the visual harness

emipasca:
Stacked on #2096 (which is stacked on #2095); retarget as the stack merges.

Closes the visual-catalog coverage gaps so every screen and major state in the app is a captured scenario on both platforms:

- The seven agent-settings scenarios gated `"platforms": ["ios"]` (provider, voice, notification rules, host access, backups, files, file editor) capture on Android too, now that #2096 serves those sections there. The `when: platform: iOS` wrappers in four flows are unwrapped, so iOS and Android run identical steps.
- Four missing views join the catalog: `home-loading` (the roster-arriving skeleton), `agent-chat-empty` (a loaded conversation with no messages), `settings-devices` (the device registry with a present device, an absent one, and the share-device-context switch, over seeded fixture devices), and `app-behind` (the below-version-window screen).
- Fixture launch modes back the new states: `visualRoster=loading`, `visualSyncState=app_behind`, deterministic `DeviceInfo` rows, and an empty-but-loaded chat hold for nova.
- One production bug found and fixed by the new `app-behind` scenario: `AppBehindScreen` replaces the navigation tree that normally hides the boot splash, so the splash never hid and the user stared at it forever. The screen now hides the splash itself. Verified by the capture: before the fix the artifact was a blank frame, after it renders "Update Vesta".

Verified with three full iOS capture runs: every scenario in the six healthy flows captures and the new states' pixels were inspected (skeleton, devices, empty chat, app-behind). The seventh flow, "Recent gateways and reconnection", is red on latest master before this stack and is tracked as #2097; until that lands, the publisher holds the previous complete catalog. `./check.sh app-mobile` green (lint, tsc, 295 tests, clean prebuild).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01367FvVt2ot8i47NQPYQ1Ym

Issue #2097, opened 2026-08-17
fix(mobile): visual-harness privacy-locked flow red since #2093 - session stuck at booting on plain launch


emipasca:
The visual QA flow "Recent gateways and reconnection" (`apps/mobile/maestro/visual/recent-gateways.yml`) fails its first assertion (`"Try again" is visible`, the privacy initialization-error sheet) on every run. Because the publisher refuses partial catalogs, this one red flow blocks publishing the entire iOS and Android galleries, and the six scenarios that live in that flow (privacy-initialization-error, privacy-locked, privacy-unlock-error, recent-gateways, recent-gateway-connecting, recent-gateway-error, agent-backups) never capture.

## Evidence

- Fails identically on latest `master` (30c2acba) and on the mobile polish branch, three runs each, always the same first assertion at ~33s while every other flow passes on warm simulators. Not flake.
- At the failure, the accessibility tree contains only the connect hero and the blocking-gate backdrop hero (both "vesta / with you through life's journey"), with no privacy sheet at all.
- The privacy sheet DOES present on the same locked launch when the launch carries `visualSession=connected` (the gateway-update flow's first block asserts "Privacy unavailable" and passes). The failing flow is the only one that plain-launches with no deep link, which makes the harness fall back to the PRODUCTION `SessionProvider` (`visual/harness/session-provider.tsx` only overrides `useSession` when `visualSession=connected`).
- The privacy gate only presents once `presentationReady` holds, which requires `status !== "booting"`. The backdrop being mounted while the sheet never pushes, and the connect route staying active, are both consistent with the production session status never leaving `"booting"`, i.e. `restoreSession`'s `await readConnection()` (SecureStore) never settling inside the visual app.

## Timeline

The flow was green during the #2094 Android-port work. The only mobile change since is #2093 (user-context), which touched `SessionProvider` (foreground token adoption), `session-model`, `app.config.ts` plugins (expo-location, expo-background-task), and the presence reporter's background poll registration. The suspicion is that one of those additions makes the first SecureStore read hang (or wedges the JS queue before `restoreSession` resolves) in the visual app's environment; the exact mechanism is not yet pinned.

## Repro

```sh
cd apps && npm run mobile:visual:capture -- --device "iPhone 17"
# or interactively:
xcrun simctl launch <visual-sim-udid> com.vesta.mobile.visual
# -> connect hero + backdrop, no privacy sheet, indefinitely
```

[...]

PR #2096, opened 2026-08-17
feat(mobile): serve every agent-settings section on Android

emipasca:
Stacked on #2095 (base is `refactor/mobile-polish`; retarget to `master` once #2095 merges).

The Android port shipped the agent-settings hub with the advanced sections gated to iOS (`showsAdvancedSections`), so provider/model, voice, notification rules, files, host access, and backup management had no Android page. Every section component already builds on cross-platform primitives (Form rows, Button, Card, toast), so this PR turns them on:

- `details/[section]` presents as a Material form sheet on Android (the same `androidSheetOptions` the logs and notifications routes use) and renders the in-content `SheetChrome` title + close button, which iOS renders natively.
- The platform gate is deleted, not widened: `sections-model` shrinks to the section list plus `findSection`, the `advanced` flag and the "hidden" availability state are gone, and `settings.tsx` renders one layout on both platforms.
- Net deletion: the "not available on Android yet" fallback can no longer be reached and is removed.

Verified with `./check.sh app-mobile` (lint, tsc, 295 vitest tests, clean prebuild). On-device Android walkthrough of the six sections is worth a pass before release; the visual-harness coverage PR that follows adds these screens to the Android catalog.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01367FvVt2ot8i47NQPYQ1Ym

PR #2126, opened 2026-08-18
epic(apps): mobile polish, Android settings, and full visual coverage

emipasca:
Continues #2095 (auto-closed when its branch was renamed `refactor/mobile-polish` -> `epic/apps`). Same commits, same history; this is the umbrella epic branch for further apps work.

One consolidated PR carrying the whole overnight mobile pass (previously stacked as #2095 -> #2096 -> #2102, folded on request). Each commit is one concern, reviewable commit by commit.

## Per-agent state preservation and UX
- Keyed per-agent holds above navigation (`src/holds/`): chat tail, composer draft + armed reply, last pager tab, and log buffer survive screen pops and agent switches; Android's pager keeps non-adjacent pages attached; pager pages mount on first visit.
- Gateway-update flow: the sheet hands off to home's live progress page when the operation starts, behind-ness is latched across socket blips, Android hardware back cannot blink blocking gates, and the home carousel stops snapping back to the last-opened agent.
- Crash and stuck-state fixes: history-seed retry with backoff, boot-splash no longer wedges on a storage failure, unhandled rejections closed, a render-phase ref hazard removed.
- Logs page fills on the boot tail replay (the "couple of lines until you scroll" bug): the inverted list snaps to the tail whenever content grows while the reader is at the tail. The 50-line fixture tail makes the agent-logs flow fail without the fix.

## Performance and hygiene
- Stable provider identities (Roster/Agent/socket memoization), keystroke-isolated chat transcript, per-palette markdown styles, memoized notification rows.
- ChatPage split from ~1510 lines into `src/agent/chat/` modules with tested models; dead code and pass-through wrappers deleted; one owner for settings sections, device labels, and provider/model derivation; shared GlassSurface; `MessageContextMenu` generic. A polish pass (review + simplify + docs) is included.

## Android settings sections
- Every agent-settings section (provider, voice, notification rules, files, host access, backups) serves on Android as Material form sheets; the platform gate is deleted. Known quirk on the tap path tracked as #2105.

## Visual harness
- Every app view is a captured scenario on both platforms: the seven iOS-only scenarios flipped, plus home-loading, agent-chat-empty, settings-devices, and app-behind (which caught and fixed a real production bug: AppBehindScreen sat invisible behind a never-hidden boot splash).
- Both capture runners gain `--gentle`: one simulator shard and utility-QoS children, so captures can run behind interactive work.

## Verification
`./check.sh app-mobile` green (lint, tsc, 299 vitest tests, clean prebuild). Multiple full iOS capture runs plus two Android emulator runs; new states inspected pixel-level. The one red visual flow is pre-existing on master (#2097, iOS-only) and blocks gallery publishing until fixed.

https://claude.ai/code/session_01367FvVt2ot8i47NQPYQ1Ym


elyxlz:
**Fixes:** nothing linked. Umbrella epic continuing #2095 (no closing keyword); references #2096/#2102/#2105/#2097. Note #2105 is the Android scroll-position quirk, not the two Android breakages below.

**Blocking**
- `apps/mobile/src/agent/settings/ProviderSection.tsx:136` (`chooseModel`, and `:150` `chooseContext`) build N action buttons for `Alert.alert`, but Android renders at most 3 buttons; with the platform gate deleted this now runs on Android, so every model past the third and the `Cancel` are silently dropped and unreachable.
  Proof: RN `Alert.alert` caps Android at positive/negative/neutral; `modelOptions.slice(0,12)`+Cancel on Claude+live models → only first 3 show. Android only; iOS unaffected.
- `apps/mobile/app/agent/[name]/settings.tsx:155` keeps `NativeSheetCloseButton` but drops `<SheetChrome title="Settings" …>`; the settings route is `presentation:"formSheet"` (`_layout.tsx:46`) so on Android it gets no native header and the `Stack.Toolbar` close button renders nothing, leaving the sheet with no title and no close affordance (swipe-only).
  Proof: sibling `details/[section].tsx:45` keeps `<SheetChrome>`; settings.tsx removed it. Contradiction between the two sheet routes.
- `apps/mobile/src/controller/gateway-update-gate-model.ts:17` clears the behind-latch on `syncState==="open"`, but the socket emits `open` on connect *before* the hello is classified, so a blip against a still-behind gateway flashes the update sheet closed then reopens it: the exact blink the latch claims to prevent.
  Proof: `socket.ts` `onopen`→`onStateChange("open")` (~:189) fires before hello's `gateway_behind` (:145); the gate then derives `dismiss`→`router.back()`, then re-pushes. `gateway-update-gate-model.test.ts` only feeds `reconnecting`, never the transient `open`, so it passes while encoding the bug.

**Non-blocking**
- `apps/mobile/src/chat/useAgentSocket.ts` ~L249: seed-retry has no in-flight guard, so a second `open` while history is failing spawns a parallel retry chain, orphaning the prior timer. Wasteful, not corrupting.
  Proof: none, read only.
- `apps/mobile/src/holds/keyed-hold.ts:11`: the four per-concern LRUs evict independently, so after 12+ agents an active agent's composer draft/pager tab can be dropped while its chat survives.
  Proof: none, read only.
- Scope: one epic folding per-agent holds, chat split, Android settings, gateway-update flow, and the visual harness across 76 files, against the repo's "one concern per PR" rule; deliberate per the body, but it makes commit-by-commit the only tractable review.
- CI: not run here; body claims `check.sh app-mobile` green (299 vitest). None of the three Blocking findings are caught by vitest (RN runtime behavior; the gate test passes despite the bug).

**Verdict:** BUGGED, the Android-settings pillar is half-broken (models unreachable past three, sheet lost its close button) and the update sheet blinks on reconnect; sound direction, three real defects to fix. The chat split, holds, roster, and boot/splash and log fixes held under attack.

---

emipasca:
**Polish pass** (`b080685e`): review, simplify, docs sweep. All of `./check.sh app-visual`, `app-web`, `app-mobile`, `guards` green locally.

**Review fixes**
- Web runner: run status now comes from a Playwright reporter (`run-status-reporter.ts`) that publishes `capturing` on the first test and `ready` or `error` (naming the failed scenarios) from the real outcome; the old `globalTeardown` published `ready` unconditionally, so a failed web scan read as success in the gallery.
- Android runner: emulator prep runs inside the try that publishes `error`, so a missing AVD or a boot timeout no longer leaves a 45-minute phantom "Scanning".
- Gallery server: a throw in the async router (malformed `%` escape, a failed spawn) answers 500 instead of an unhandled rejection that killed the process; runner names are checked with `Object.hasOwn`, so `/capture/constructor` is refused.
- Web runner honours a scenario's `platforms` scope (light project and dark sibling) and writes the registry's `screenshot` name; the shot goes straight from the page buffer into the store (one write, no artifact copy).
- Android-only constants moved out of the iOS runner into `visual-android.mjs` (`metroConfigPath` into the shared runner); gentle mode no longer passes `CAPTURE_URL_2=undefined`.
- The JS-bundle fingerprint now includes `apps/core/src` and the real workspace lockfile, so a `@vesta/core` edit can no longer reuse a stale installed bundle silently.
- The bridge watchdog's rejection is observed, so a slow start can no longer surface as an unhandled rejection.

**Simplify**
- `RUNNERS[*].reportFile` names each runner's report entry (Maestro `report.html`, Playwright `index.html`), so the Web report link can actually appear.
- `pngSize` reads only the 24-byte header, and `composeGallery`/`shotEntries` run their file reads in parallel.
- Dropped unused `platformFamily`/`runnerOf`; `AndroidBottomInset` memoizes its provided value; the agent pager tracks the active index instead of re-deriving it.

**Wiring and docs**
- `@vesta/visual` declared as a devDependency where imported (`@vesta/web`, `@vesta/mobile`).
- CI: `apps/visual/**` removed from the `web`/`mobile` path filters; the closure lives in the job `if:` (`visual == 'true'`), per AGENTS.md.
- `expo lint .` covers the whole mobile package again (the earlier `src app scripts visual` narrowing had dropped `app.config.ts`, the plugins, and the Maestro helper); Maestro's `runScript` globals are declared at config level for `maestro/**/*.js`.
- AGENTS.md: `apps/visual/` in Project layout, `./check.sh app-visual` in Commands, `web` chains five slices, CI section names five per-app jobs and the visual closure.
- Skill and README examples use a real Playwright project (`--project desktop`) and the per-runner report file.

**Not done, flagged for the author**
- `restoreEmulator` restores the cutout overlay but not the navigation overlay, so a `--device` run against a personal emulator leaves it in the variant's nav mode; needs a decision on the right restore semantics.
- The `.d.mts` declaration files are hand-maintained and unchecked by any tsc; a `tsc` step in `check_app_visual` (or `checkJs`) would catch drift.
- `apps/mobile/scripts/visual-ios.test.mjs` only tests the Metro fixture config, not the iOS runner; rename only.

Verify was run by the orchestrating session after the polish agent's process was interrupted mid-run; every applied change was reviewed line by line before landing.

PR #2128, opened 2026-08-18
docs(restart): a FAILED daemon start is the agent's to retry and report

### LINKED ISSUE #2115: restart: failed daemon start not retried, transient host-load contention leaves a channel down
### Summary

`agent/skills/restart/start-daemons.sh` runs each `<skill> daemon start` line exactly once. When a start's port-readiness probe times out under transient host-load contention (a shared kernel where a neighbouring container spikes load), the start is abandoned and reported `FAILED`, even though the same start succeeds moments later once load subsides.

### Detail

- The readiness wait in the per-skill `daemon.py` (`_await_ready`) already has a generous, env-overridable budget (`READY_TIMEOUT_SECS`, default 30s, `DAEMON_READY_TIMEOUT_SECS`). A start that exceeds it calls `_abandon`, which terminates the child and removes the pid/port records, so a failed start leaves nothing half-up.
- `start-daemons.sh` treats that non-zero exit as terminal: the daemon is reported down and never retried, so a channel stays mute after a restart until a human notices and re-runs the start by hand.
- Observed: two daemons failed on the first `start-daemons.sh` pass during a host-load spike; both came up cleanly on an individual retry a moment later at lower load. The processes were genuinely not up after the failed attempt (no pid, no listener, no orphan), consistent with `_abandon` cleanup.

Raising the fixed timeout is the wrong lever (it is already 30s and overridable, and a larger fixed wait is paid on every start yet still not guaranteed under variable load). The readiness/abandon logic is also duplicated across several skills' `daemon.py`, so the caller is the natural single place to add resilience.

### Fix

Retry a failed start once in `start-daemons.sh`, after a short pause. A start is idempotent (an already-up daemon answers `already_running` and spawns nothing; a failed start cleaned up its own records), so a retry is safe and only pays cost when a start actually failed. The restart SKILL.md already documents the script as safe to re-run for crash/timeout recovery, so per-line retry is in line with the existing design.

---
Submitted by **apollo** on vesta v0.2.3

emipasca:
Closes #2115 by putting the recovery where the intelligence is, instead of a silent shell retry (the approach in #2116).

## The gap

`start-daemons.sh` already reports a failed start loudly: a per-daemon `FAILED:` line on stderr and a non-zero exit, read by the agent in the turn that ran it. What is missing is any instruction to act on that line. The restart SKILL.md describes how starts work but never says what a `FAILED` line demands, so the agent reads past it and a channel stays down until someone notices by hand. That is the actual root cause behind #2115's observation.

## The fix

Two sentences in the restart skill's Daemons section: a `FAILED` line is the agent's to handle before moving on. Retry that one daemon's line (safe: a failed start removes its own pid and port records, and the SKILL.md already establishes idempotence). If the second attempt also fails, read the daemon's log, find the cause, and fix it; the user comes in only when the fix needs something only they can provide (a credential, a re-link, a decision). The paragraph also names the stake: a daemon left down fails silently, messages land nowhere, reminders never fire.

## Why not the shell retry in #2116

A retry loop inside `start-daemons.sh` absorbs the failure invisibly: when the retry works, nobody learns a daemon needed two attempts; when it fails, the agent is exactly as uninstructed as before. The agent can do everything the loop can, plus diagnose the log, fix the cause, and decide when retrying is pointless. Keeping the script single-attempt keeps the failure loud, and the loudness is what the whole recovery hangs on.

This composes with the readiness-budget fix (#2119, which makes the launcher stop killing healthy-but-slow daemons) and with the out-of-loop watchdog (#2121, which owns deaths after boot). Each governs a different failure at its own layer.

## Checks

`./check.sh guards` passes in a clean worktree off master.


elyxlz:
**Fixes:** #2115. The issue asks that a `FAILED` daemon start no longer leave a channel silently down. This resolves it at a different layer than the issue proposed (agent-driven retry + escalate, not a shell loop), and does close the observed gap: the agent already reads the `FAILED:` line in-turn but had no instruction to act on it. Legitimate resolution, and it does more than the loop (log diagnosis + user escalation). Deliberately competes with #2116; this is the author's chosen approach, so #2116 should close if this lands.

**Blocking:** none.

**Non-blocking**
- The prose tells the agent to read `~/agent/logs/<name>.log`, but the `FAILED` line's `<name>` is the command's first token (`start-daemons.sh:26`, `name="${cmd%% *}"`), while the log is keyed by the daemon name, which diverges for `voice-keys` (daemon named `voice`, per AGENTS.md) and for per-instance whatsapp lines. An intelligent reader recovers, but the literal path can miss.
  Proof: `voice-keys daemon start` → FAILED name `voice-keys`, yet log is `~/agent/logs/voice.log`.
- No pause suggested before the retry; under sustained host-load contention (the exact trigger in #2115) an immediate re-run can hit the same readiness-budget timeout. #2116 paused first. Minor, agent judgment covers it.
  Proof: none, read only.

Mechanism checks that held: script does emit `FAILED:` on stderr with non-zero exit; "a failed start removes its own pid/port records" matches the fails-closed daemon contract (AGENTS.md); prose carries no dash separators and no >8-line block, so `guards` prose rules pass; present-tense mechanism only, no changelog narration.

**Verdict:** MERGE, correct fix at the right layer (the agent already sees the failure), complies with conventions, only cosmetic log-name imprecision worth a follow-up.

---

PR #2129, opened 2026-08-18
docs(proactive-check): daemon liveness is the agent's duty, never a built automation

### LINKED ISSUE #1681: Nothing detects a daemon that has exited: a dead poller is indistinguishable from a quiet period
## Summary

A daemon started by `<skill> daemon start` runs detached with nothing supervising it. When a poller daemon exits, the only observable effect is that its notifications stop, and that is indistinguishable from a period with nothing to report. On this box a poller daemon was found dead only when a human checked by hand, well after the fact. Nothing had flagged it.

## Corrected diagnosis, please read this before filing a retry PR

The first read of this incident was transient upstream errors in the daemon log: a 503 from the provider API, and an `[SSL: UNEXPECTED_EOF_WHILE_READING]`. That diagnosis is wrong, and a retry or backoff patch would fix nothing. The poll loop in question already wraps its whole body in `except Exception` and continues to the next cycle, and the per-account fetches inside it carry their own handlers, so a transient error is logged and the loop goes on. I verified this by reading the source. The actual cause of the exit remains unknown, which is the point of this issue: nothing recorded it.

## What I verified in the tree

Paths are relative to `agent/`. Checked in a worktree off `upstream/master` at `fd05196`.

### 1. There is no supervisor. `daemon start` forks and forgets.

`skills/enable-banking/cli/src/finance_cli/daemon.py:82-101` (`_start`) spawns the child with `subprocess.Popen(..., start_new_session=True)`, writes the pid, sleeps `SETTLE_SECS`, checks `child.poll()` once, prints `{"status":"started"}`, and the parent exits. Nothing waits on the child after that one check.

That same file exists, byte for byte apart from the daemon's name, in slack, microsoft, spotify, tricount, email-client, google, agentmail, app-chat, voice, and tasks. I diffed two copies with the name normalized and got an empty diff, so this is one template duplicated across the fleet rather than a quirk of one skill.

The single exception is telegram, which spawns `skills/telegram/telegram-watchdog.sh` from its own start (`skills/telegram/cli/daemon.go:383-406`, `ensureWatchdog`). That script polls the pidfile every 30s, restarts the daemon, rate limits itself, and writes a `daemon_restarted` notification whenever it acts. It is the only watchdog in the tree: `ls agent/skills/*/*watchdog*` on master returns that one file, and `skills/restart/SKILL.md` records that the whatsapp equivalent was deleted outright. So supervision exists for exactly one channel, added by hand, rather than as a property of the daemon contract.

### 2. `daemon status` on a dead pid says not-running and leaves the record behind.

`_status` is one line: `print(json.dumps({"running": live_pid() is not None, "port": None}))`. `live_pid()` returns `None` when `os.kill(pid, 0)` raises, so the answer is a plain `"running":false`. The stale pidfile is not cleaned, no distinction is drawn between "never started" and "was started and is gone", and nothing is emitted anywhere. A caller who does not already know that this daemon is supposed to be up learns nothing from the answer.

### 3. No exit code and no exit event is recorded anywhere.

The child's stdout and stderr append to `~/agent/logs/<name>.log`, and that is the whole record. The parent never calls `wait()`, so the exit status goes to init and is lost. There is no exit record, no timestamp of death, no last-seen marker.

The contract does describe an in-process death signal. `skills/vestad/SKILL.md:146-149` says: "A daemon that writes a `daemon_died` notification recognizes SIGTERM in-process and stays silent for it, so a deliberate stop is never reported as a crash. Every other way out is reported, since nothing else notices a daemon that quietly went away." `skills/tasks/cli/src/tasks_cli/cli.py:455-490` implements exactly that: signal handlers, a `try/finally`, and a `daemon_died` notification whenever the exit was something other than the SIGTERM that `daemon stop` sends.

The obligation is conditional on a daemon choosing to write one, and several shipping daemons do not. A `grep -rn daemon_died --include=*.py --include=*.go` over `skills/` finds it in tasks, google, microsoft, discord, email-client, agentmail, app-chat, telegram, tricount, and whatsapp. It finds nothing in enable-banking, slack, spotify, dashboard, file-host, sign-service, or ssh-tunnel. Three of those (enable-banking, slack, spotify) are notification-producing daemons whose entire purpose is to page the user, so their death is silent by construction.

Even where it is implemented, `daemon_died` is written by the dying process itself, so it cannot cover SIGKILL, an OOM kill, or a container-level kill.

### 4. The only scheduled liveness check is prose in a prompt.

`skills/proactive-check/SKILL.md` opens with a "Preflight: daemon liveness (do this first, every tick)" section telling the agent to turn each `<skill> daemon start` line in `skills/restart/SKILL.md` into the matching `status` call and to bring back anything reporting `"running":false`. It runs on `proactive_check_interval`, default 60 minutes (`core/config.py:520`).

This is real, and it is the only thing on a schedule that looks at daemon liveness at all. Its limits are structural:

* It is an instruction to a model rather than machinery. It runs only when the agent loop is running, and only when the model actually performs the preflight on that tick. In the incident that prompted this issue, the dead daemon's start line was present in the restart block and the gap still went unnoticed.
* It samples current state rather than observing an exit, so it cannot say when the daemon went or why, and a daemon that dies and is restarted between two ticks leaves no trace at all.
* It only covers daemons whose start line lives in the restart block.
* Restoring the daemon quietly is its success path, so a daemon that dies every hour looks the same as one that never dies.

## Why this matters

The failure mode is user-visible and silent at the same time. The user stops receiving spending alerts, or new-mail notifications, and there is nothing to notice, because the absence of a notification is the normal state most of the time. Detection today depends on a human eventually wondering whether they should have heard something by now.

## Possible shapes for a fix

A sketch, roughly in order of cost. I have implemented none of these.

1. Make the `daemon_died` notification part of the shared lifecycle rather than a per-daemon opt-in, and add it to the daemons that currently lack it. Cheapest option, covers clean exits and unhandled exceptions, does not cover a kill.
2. Have `daemon status` treat a stale pidfile as news: report a third state for a record naming a dead process, clear the record, and emit a notification. Small and testable per skill, but still only observed when something calls status.
3. Give the contract a supervisor, generalizing what telegram already does by hand: one watchdog that reads the restart block and restarts anything whose pidfile names a dead process, rate limited, notifying when it acts. This is the only option that covers a kill nobody can catch in-process, and it is the largest change.
4. Give each poller a heartbeat file it touches on every successful poll, so staleness becomes measurable instead of being inferred from a pid.

I did not open a PR because the fix touches the daemon contract itself plus roughly ten copies of the same lifecycle file, and the shape should be the maintainer's call.

## Verified versus inferred

Verified by reading `upstream/master` at `fd05196`: every claim in points 1 through 4, the absence of `daemon_died` in the skills named above, and the identical `daemon.py` copies.

Inferred, not tested: that a SIGKILL or an OOM kill leaves neither a `daemon_died` notification nor a log line, since the writer of both is the process being killed; and that the proactive-check preflight either did not run or did not act on the ticks between the daemon's death and its discovery, since no record of either exists to check.

---
Submitted by **vesta** on vesta v0.1.183

emipasca:
Closes #1681.

The proactive-check preflight already makes daemon liveness the first act of every tick: status each daemon the restart skill starts, bring back what is down. What was missing is the doctrine that this duty is the agent's own and must never be delegated to built machinery. Fleet boxes have started building watchdogs (PR #2121, closed on this principle), and each one is a finicky automation that restarts blindly, hides the cause, and becomes one more silent failure mode.

This adds two sentences to the preflight: never build a watchdog, cron job, or auto-restarting script for this check; the agent reads the log, fixes the cause in the skill that broke, and judges when the user must be involved.

This is the settled resolution for #1681 as well: a dead daemon is caught by the agent's own periodic check, and recovery plus diagnosis stay with the agent (see also #2128 for the boot-time FAILED handling). No detector, no auto-restart.

`./check.sh guards` passes.

### DIFF
--- a/agent/skills/proactive-check/SKILL.md
+++ b/agent/skills/proactive-check/SKILL.md
@@ -11,6 +11,8 @@ This is your scheduled moment to think unprompted. No one asked; you're checking
 
 Before anything else, confirm the daemons the `restart` skill starts are actually alive: ask each one it starts with the matching `<skill> daemon status`. Treat an error, a missing command, or any answer not reporting running as down, not only an explicit `"running": false`. A daemon can die silently (container up, daemon down), and a dead messaging daemon means you cannot reach the user at all, which is why this check comes first. Bring a dead daemon back with its own `<skill> daemon start`, or re-run the `restart` skill, which is idempotent and a no-op when everything is already up.
 
+This check is your duty and stays yours: never build a watchdog, cron job, or auto-restarting script to do it for you. An automation restarts blindly, hides the cause from you, and becomes one more thing that breaks silently; you can read the log, fix the cause, and judge when the user must be involved. When a daemon you brought back died for a reason worth fixing, fix it in that skill, not with a new layer of machinery on top.
+
 ## Two questions, every time
 
 Your running narration is visible to the user in the app: think out loud like yourself, not like a service log.


elyxlz:
**Fixes:** #1681 by doctrine, not by closing it. The issue is a detection gap (nothing observes a daemon exit). This adds no detection: it leans on the same hourly preflight the issue documents as having *missed* the original incident, and does nothing for SIGKILL/OOM, which the issue notes leave neither a `daemon_died` notification nor a log line. So this closes #1681 as a deliberate "the agent's own check is the answer, do not automate," a wontfix by principle, not a fix for the gap.

**Blocking:** none.

**Non-blocking**
- `agent/skills/proactive-check/SKILL.md:14` "never build a watchdog" is absolute, yet a watchdog ships in the tree. An agent that encounters `agent/skills/telegram/telegram-watchdog.sh` reads a contradiction; scoping the rule to *your own* machinery would remove it.
  Proof: `ls agent/skills/*/*watchdog*` → `telegram-watchdog.sh`.

**Verdict:** MERGE — two sentences of coherent, correctly placed doctrine, brand voice clean (no dashes, no pronoun for Vesta), no lint concern. Just merge it knowing it resolves #1681 by declaring the gap acceptable, not by closing it; the SIGKILL/OOM blind spot from the issue stays open.

---

PR #2130, opened 2026-08-18
docs(upstream-pr): give issue filing a complete worked recipe

emipasca:
## Problem

The "Filing an issue" section of `agent/skills/upstream-pr/SKILL.md` was one sentence: capture a token, then "POST to the GitHub Issues API". An agent that follows it must rediscover the whole path on the box. Apollo filed two issues today (#2115 overnight, #2120 on request) and burned turns on four gaps, all visible in the transcript:

1. It tried `gh issue view` / `gh pr view`. `gh` does not exist in the container ("No such file or directory").
2. It reached for `jq` to build the JSON body. `jq` does not exist either; it fell back to `python3 -c 'json.dumps(...)'`.
3. It set `TOKEN=$(upstream-pr --token-only)` without `export` and passed the token to a Python child reading `os.environ`. The child saw nothing and the attempt failed.
4. It hand-assembled the curl POST from scratch, twice, with a hedge about a possible 403 that the skill's own permission table already answers.

## Change

- One intro sentence stating the constraint the agent cannot guess: no `gh`, no `jq` on the box; GitHub work is `upstream-pr` plus `curl`, JSON via `python3`.
- "Filing an issue" now carries a complete one-command recipe: same-shell token capture, quoted heredoc for the body (keeps backticks and `$` literal), attribution footer appended outside the heredoc so `$AGENT_NAME` and the version expand, `json.dumps` for the payload, and a response parse that prints the issue URL or the API's error object.
- An explicit note on the export rule: only `curl` reads `$TOKEN` in the recipe, but a token-consuming child process needs `export`.

The one-shot constraint (no edit, no comment after create) was already documented in the permission table; the section now points at it at the moment it matters, since the footer has to be in the body at create time.

## Verification

- Dry-ran the recipe with a fake payload: heredoc content stays literal, footer expands, JSON round-trips.
- `grep -rnP '\x{2014}|\x{2013}'` on the file is clean (the `agent-tests` dash check).

### DIFF
--- a/agent/skills/upstream-pr/SKILL.md
+++ b/agent/skills/upstream-pr/SKILL.md
@@ -5,7 +5,7 @@ description: Upstream elyxlz/vesta GitHub ops: branches, PRs, issues, CI, API.
 
 # Upstream PR
 
-Push contributions back to `elyxlz/vesta`. Authentication is handled by the `vesta-upstream` GitHub App, no personal tokens needed.
+Push contributions back to `elyxlz/vesta`. Authentication is handled by the `vesta-upstream` GitHub App, no personal tokens needed. The box has no `gh` and no `jq`: every GitHub operation is `upstream-pr` itself or `curl` against the API with the token it mints, and JSON is built and parsed with `python3`.
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


elyxlz:
**Fixes:** nothing linked. Closes a real doc gap: the "Filing an issue" section was one sentence, so an agent had to rediscover the whole curl path on the box (the four gaps in the PR body are the symptom).

**Blocking:** none.

**Non-blocking**
- The repo's own "Shaping the change" gate in this same file says fold new functionality into the CLI as a subcommand, not a standalone recipe: an `upstream-pr --issue --title --body` would be a smaller runtime surface than a 15-line recipe the agent must reproduce each time. Docs-only is a fair interim, but that is the more elegant fix.
  Proof: `agent/skills/upstream-pr/SKILL.md` "Shaping the change (REQUIRED)" vs the new recipe block.
- `curl -s` on a network failure prints nothing, so the `json.load(sys.stdin)` parser tracebacks instead of printing an error object; harmless (still legible as "not a URL"), noting only.
  Proof: `echo -n '' | python3 -c 'import json,sys; json.load(sys.stdin)'` → JSONDecodeError.

**Verdict:** MERGE. Attacked the recipe (ran it end to end: backticks/`$` stay literal, footer expands, JSON round-trips, both parse branches work), the box facts (`gh`/`jq` absent from `vestad/Dockerfile`, `python3` present via novnc/websockify, `AGENT_NAME` is a container env), footer/version match with the Attribution section, and the dash + prompt-guide story-narration checks. All held; the diff carries the mechanism, not the incident.

---

emipasca:
Superseded by #2134: the curl recipe is replaced by the `upstream` CLI's guarded `issue create`.

PR #2132, opened 2026-08-18
feat: jitter the nightly dream and the maintenance window per night

emipasca:
## What

Two nightly automations fired on rigid, repeating instants. Both now start at a per-night pseudorandom minute, with no new config, no persisted state, and no wire change.

**Dream (agent)**: the nightly dream fired at the first hourly check after `nightly_memory_hour`, so every agent on a host compacted and restarted in the same hour-aligned burst. Each night now starts at a minute in `[hour:00, hour:59]`, hashed from (agent name, night date) in `agent/core/loops.py`. The hash (not `random`) keeps the instant stable across the hourly checks and across restarts, while varying it per night and per agent.

**Maintenance window (vestad)**: the pass polled on a boot-anchored 15-minute grid from 4:00 local, so it landed on the same clock minute night after night. The window now opens at `[4:00, 4:30)` plus a per-night minute hashed (inline FNV-1a, no new dependency) from the zone-local date in `vestad/src/maintenance_window.rs`. The 5:00 close stays fixed, so an idle-gated poll and the closing poll always fit, and the pass still always lands inside its window. `in_window` is the one owner of the window shape, so the 24h coverage scan and the live poll shift together and `serve.rs` is untouched.

The two compose: a dream can now run past 4:00, and the maintenance idle gate already defers the pass while an agent dreams.

## Why

Retro on an agent that crash-looped during last night's dream restart under host IO saturation: hour-aligned dream restarts and a fixed-minute maintenance pass concentrate nightly load spikes. Jitter spreads them. (The IO saturation itself is a separate issue.)

## Testing

- `./check.sh guards` green.
- `./check.sh agent` green (1151 passed; the 2 `test_migrations.py` workspace-repair failures are the known host `tag.gpgsign` artifact, and pass with `GIT_CONFIG_GLOBAL=/dev/null`).
- `cargo clippy -p vestad -- -D warnings` and `cargo test -p vestad --bins` (483 passed) green.
- TDD: new tests pin hashed instants (`sha256("dreamer-test:2025-06-15") % 60 == 43`; `fnv1a("2026-01-01") % 30 == 19`), hold-before/fire-after boundaries, and that consecutive nights differ.

### DIFF
--- a/agent/core/loops.py
+++ b/agent/core/loops.py
@@ -3,6 +3,7 @@
 import asyncio
 import contextlib
 import datetime as dt
+import hashlib
 import json
 import pathlib as pl
 import time
@@ -483,6 +484,15 @@ def check_proactive_task(*, config: cfg.VestaConfig) -> None:
 
 
 DREAMER_CATCHUP_HOURS = 6
+DREAMER_JITTER_MINUTES = 60
+
+
+def _dream_delay(night: dt.date, *, config: cfg.VestaConfig) -> dt.timedelta:
+    """Per-night start delay inside the configured hour, hashed from (agent name, night): stable
+    across restarts and checks (a fresh random would keep moving the target), different each night
+    and per agent, so dream restarts across a host never land on the same minute."""
+    digest = hashlib.sha256(f"{config.agent_name}:{night.isoformat()}".encode()).digest()
+    return dt.timedelta(minutes=int.from_bytes(digest[:4], "big") % DREAMER_JITTER_MINUTES)
 
 
 def process_nightly_memory(*, state: vm.State, config: cfg.VestaConfig) -> None:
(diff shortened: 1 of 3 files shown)


elyxlz:
**Fixes:** nothing linked. Spreads nightly load: per-night jitter on the dream start (agent) and the maintenance window open (vestad), so a host's agents stop compacting/restarting in one hour-aligned burst. Worth carrying.

**Blocking:** none.

**Non-blocking**
- `maintenance_window.rs:56` window shrinks to 31 min at jitter 29, and the poll/scan step is 15 min, so the "idle poll + closing poll both fit" claim rides on 31>2×15.
  Proof: window `[4:29,5:00)` admits polls at 4:30 (idle-gated) and 4:45 (4:45+15=5:00 → closing fires); holds because jitter is capped below 30 and close is fixed. Correct, but load-bearing and only implied by the const comment.
- Two subsystems in one PR, cleanly split per commit; one coherent concern (nightly jitter), so fine, just noting.

**Verdict:** MERGE. Attacked completeness (goal reached: dream held at hour, fires at hashed minute), the jitter/catchup/midnight boundaries, the scan-coverage guarantee, and the missed-night path; all held. Hashes reproduce (sha256→43/31, fnv→19/20); reverting the dream guard reddens exactly the two new jitter tests (`2 failed`); `cargo test maintenance_window` 6/6, `pytest test_dreamer.py` 39/39. `in_window` is the sole window owner and `serve.rs` composes unchanged.

---

PR #2134, opened 2026-08-18
feat(upstream): rename upstream-pr to upstream and wrap gh with injected auth

emipasca:
## What

The `upstream-pr` skill and CLI become `upstream`, an auth + policy wrapper over `gh`.

- `upstream gh <args>` mints the GitHub App installation token in-process and runs `gh` with `GH_TOKEN` and `GH_REPO=elyxlz/vesta` injected into the child environment, so the token never exists in the shell. Everything an agent knows about `gh` works as-is, from any directory.
- Three calls are intercepted for logic `gh` lacks, in the same spelling: `pr create` (title validation, shared-history guard, ownership guard with `--adopt`, authenticated force-push with the token in env never argv, attribution footer), `issue create` (validation + footer, one `body_with_attribution` owner for both creates), and `pr list --mine` (commit-author identity, since every PR here is authored by the same bot). gh's built-in verb aliases (`new`, `ls`) are canonicalized before dispatch so they cannot bypass the guards.
- Titles are validated against the issue-triage taxonomy (`type(scope): description`, ten pinned types) and refused with the broken rule named; over 72 chars warns only.
- `upstream token` remains as the escape hatch, with the same stdout hygiene rule.
- `requests` is removed: the token mint uses urllib, all REST rides `gh api`, and a missing `gh` binary is a clear one-line error instead of a traceback.
- `gh` ships in the agent image (Debian trixie package), and the fleet migration installs it on existing boxes, moves the App key if a dirty checkout stranded it, reinstalls the uv tool under the new name, and clears the stale skill name from the active list.
- SKILL.md shrinks from 250 to 130 lines: ten lines of wrapper mechanics plus the judgment content gh cannot carry (discovery sweep, filing gates, worktree recipe, local apply, formatting rules).

## Why

Agents already know `gh` cold. Yesterday an agent reached for `gh issue view` on a box that lacks it, fell back to `jq` (also absent), lost an attempt to an unexported token, and hand-built the same curl recipe twice in one day. Wrapping `gh` converts that training instead of teaching a bespoke surface, and interception makes the natural spelling the guarded path by construction.

One deliberate window: a box that syncs the new agent home runs the old installed command until its migration turn on the same boot re-installs the tool and apt-installs `gh`; failures in that window are loud (the CLI names exactly what is missing) and converge on the next boot.

## Verification

- CLI suite: 53 tests, including a fake `gh` executable on PATH that records argv and env, pinning the token in the child env and never in argv, exit-code propagation through the passthrough, alias canonicalization, footer-alone bodyless creates, and the title-policy table.
- `./check.sh guards` green; `./check.sh agent` green locally except the two documented local-only `tag.gpgsign` `test_migrations` failures that also fail on untouched master (host git config; green in CI).
- `gh` confirmed present in Debian trixie main (2.46.0-3), so the image build and the migration's apt path use the same package.

Supersedes #2130.


elyxlz:
**Fixes:** nothing linked (supersedes #2130). Replaces the bespoke `upstream-pr` CLI with `upstream`, an auth+policy wrapper over `gh`, so agents reuse `gh` knowledge instead of a hand-built surface. Worth carrying.

**Blocking:** none.

**Non-blocking**
- `cli/src/upstream_cli/cli.py:pr_create`/`issue_create` intercept the two creates and reject any gh flag outside the pinned set (`--draft`, `--fill`, `--label`, `--reviewer`, `--web`, `-t`/`-b` short forms all exit 2). The body's "everything you know about `gh` works as-is" holds for passthrough reads but not the guarded creates; SKILL.md does document the supported set, so this is a claim/scope note, not a defect.
  Proof: `test_create_refuses_unsupported_gh_flags` pins `--draft` → exit 2.
- Title policy requires a lowercase-letter first char, so `fix(api): OAuth ...` is refused as "description must start lowercase". Matches repo convention (issue-triage examples are all lowercase), intended.
  Proof: `titles.py` `_TITLE_RE` desc group `[a-z].*`; read only.

I attacked completeness (reader sweep, leftover refs), the ownership/force-push guard, the migration's key-move ordering and idempotency, and title validation vs the taxonomy; all held. Migration is a new append-only file, each step reads disk first, and step 1 moves the untracked key before step 2 deletes the stale dir.

**CI:** guards + Python + Rust green; agent/vestad suites pending at review time, none failed. Watch `Agent · Typecheck and test`.

**Verdict:** MERGE, real subtraction (bespoke CLI and `requests` gone, `gh` shipped in image + fleet migration), attacked and held.

---

emipasca:
## Polish pass

One commit on top of the branch (`34786ad5c`). Nothing here changes the PR's intent: the rename, the `gh` wrapper, and the three guarded calls are as you filed them.

**Review + simplify**

- `titles.py` parsed twice: a strict acceptance regex, then a looser regex to guess which rule broke. The guess path never knew the scope charset, so `fix(my scope): retry` reported `description must start lowercase`. Now one parse with a per-rule ladder, `_SCOPE_RE` the single owner of the charset. Errors accumulate, so a title breaking two rules is refused once rather than twice. Also drops the `!= .lower()` disjunct, which was unreachable given `TYPES` is all lowercase.
- `_intercept_parser` forwarded into `add_argument` and made no decision, while the parse-and-refuse block it should have owned was copy-pasted at all three call sites. They had already drifted: `pr list --mine` printed no help text at all. Folded into `_parse_intercepted`.
- `f"{agent_name} (vesta)"` was written in three places. It is set on the push and read back by both the force-push ownership guard and `--mine`, so a drift there fails open and silently. Now `commit_author_name`.
- `--mine` and the existing-PR lookup interpolated user-controlled values into a query string unencoded. Branch names may contain `&`, so `quote()` on each.
- `--mine` fetched `per_page=100` and sliced to 40, and the existing-PR payload was `json.loads`ed twice.
- `--branch` is the pre-rename spelling of `--head` and carried no `LEGACY(remove-when:)` marker, which AGENTS.md requires for convergence-only code. Marked, and dropped from the help text so it is not advertised alongside the spelling SKILL.md teaches.
- `canonical_gh_args`'s docstring claimed `issue new` "cannot walk past the guarded create". It is argv matching, and SKILL.md deliberately teaches `upstream gh api`, so the door is not closed. Reworded to what it does.
- Return annotations on the new functions.

**Tests**

- `SENTINEL` was declared in four files, including the fixture that actually injects it, in a suite whose point is asserting that exact token never leaks. One owner in `conftest.py`, plus an `agent_identity` fixture for the stub repeated five times.
- The two `--mine` tests differed only in `list` vs `ls`: parametrized.
- `test_auth.py` had two near-identical helpers driving `main()` to the push; merged into one that returns both what the push saw and the guard calls.

53 CLI tests still pass.

**Prompt sweep**

- The skill `description` summarized what the skill does. Rewritten as when to reach for it, since an agent that can read the capability list from the description tends not to open the body.
- One em-dash separator in an agent-facing comment, and a ragged comment reflow in `redact_secrets.py`.

**CLAUDE.md**

No change needed. No `upstream-pr` reference survives outside the migration, and AGENTS.md never named the skill, so the rename made nothing false. Flagging one judgement call rather than writing it: `upstream` (contribute out) and the core `upstream-sync` skill (pull in) are now adjacent names, and AGENTS.md's "Upstream sync" section describes only the latter. Nothing is wrong today; worth a disambiguating clause if that section is ever touched.

**Verify**: `guards` and `vestad` (482) green. `agent` is 1147 passed, 2 failed, both `test_migrations.py::test_workspace_repair_*`, the known `tag.gpgsign` local-only flake on this host (confirmed `tag.gpgsign=true` globally); they fail on untouched master and pass in CI.

**Deliberately skipped**: threading the `--mine` per-PR author lookup (real N+1, but concurrency plus secondary-rate-limit risk is not a polish-pass change), annotating the rest of `cli.py` (the module is unannotated throughout, so it would dwarf the diff), and splitting the git half of `cli.py` into its own module (a good next seam, but roughly doubles this PR).

PR #2139, opened 2026-08-19
feat(skills): add maps skill for Google Maps search, links, and routes

emipasca:
## What

A new `maps` skill: a stateless CLI that replays Google Maps' own web endpoints over plain HTTP (no browser, no account, no API key), so the agent finds real places, answers with verified metadata, and hands the user links that open the exact place or route in the Maps app or on the web. Four RPCs are reverse-engineered and replayed: search, place detail, directions (including timed transit), and reverse geocode.

## Commands

- `maps search "<query>" [--near] [--min-rating] [--category] [--radius-km] [--sort rating|distance] [--limit]`: ranked places, each with name, address, coords, `place_id`, `cid`, `ftid`, rating, review count, category, phone, website, `open_now`, and ready links (`place_url` opens the exact place; `directions_url`).
- `maps show <cid>`: full detail for one place: today's hours and numeric open intervals, phone, website, photos, coords.
- `maps directions --to <cid> [--from <cid>] [--mode driving|transit|walking|bicycling] [--steps] [--depart HH:MM | --arrive HH:MM, with --tz]`: a link addressed by the place's `place_id` (never a dropped pin), plus the fetched leg (duration, distance, steps). A timed transit request fetches the schedule-aware route and returns a link that opens Maps with the time preset.
- `maps route --stops '<json>' [--mode]`: one link that opens a whole ordered multi-stop route with named stops.
- `maps itinerary --stops <cid,cid,...> [--start] [--dwell] [--optimize]`: a scheduled day plan: per-stop arrive/leave times, an `open_at_slot` flag when a stop's slot falls outside its opening hours, leg durations, and one route link.
- `maps geocode "<address>"` / `maps reverse <lat,lng>`: address to coords and coords to Google's label.
- `maps doctor`: five independent checks, one per RPC template, anchored on Buckingham Palace (a durable cid); exit 0 when all pass, and a failing check names the template to re-capture.

## How the agent uses it

Search in plain text, pick the right result (the default table shows each row's `cid`), then act by `cid`: `show <cid>`, `directions --to <cid>`. A transparent identity cache (`~/.gmaps/places.json`) holds each seen place's name, coords, and ids, so a cid reference resolves with no extra network call. Tables by default, `--json`/`--json-pretty` for structure; errors are `{"error": ...}` on stderr with a non-zero exit. SKILL.md documents the workflow and trip files; personalization.md teaches recording the user's places and routes by `cid`; presentation.md covers showing places in chat.

## How it works

Modeled on the `tricount` skill. Each RPC rides a captured `pb` template with slots; Google does not validate the session token in the place and directions templates. `proto.py` owns the pinned positions in Google's positional place proto, which both the search feed and the place-detail response carry, so `parse_search` collects proto subtrees directly and both response shapes (list feed and single-entity landmark answer) parse through one path. Opening hours are read from the numeric interval pairs, so they are locale-independent, and `open_now` is Google's own open-state code. `client.py` is the sole IO edge: paced requests, consent-wall detection, and a known-good canary that separates "no results" from schema drift.

## Tested

47 CLI tests against committed fixtures: link builders, pb/token/envelope plumbing, proto readers (state codes, clock pairs, midnight wraps), both search shapes, place detail, directions, reverse, the scheduler, the cache, the doctor (with per-check failure isolation), and the CLI output contract. Every command was also smoke-tested live end to end.

## Follow-ups

Filtering across transit route alternatives (prefer a line, fewer transfers) and the authenticated saved-lists path (the SAPISIDHASH signer is built and tested, ready for it).


elyxlz:
**Fixes:** nothing linked. Adds a `maps` skill (scrape Google's web RPCs) so the agent hands verified place links and routes. Legit capability; the scraping path is fragile but honestly gated by a drift canary.

**Blocking**
- `client.py` `--open-now` is a no-op: `parse.py` never sets `Place.open_now`, so `p.open_now is not False` keeps everything. SKILL.md's headline example uses `--open-now`, so the agent presents possibly-closed places as open.
  Proof: `_apply_filters([open_now=None], SearchFilters(open_now=True))` → kept both places.
- `cli.py:_cmd_search` emits a note telling the agent to run `maps show <cid>`; no such subcommand exists, so following the instruction fails.
  Proof: `maps show 123` → `error: invalid choice: 'show'`.

**Non-blocking**
- `--max-price` filters on category-presence, not price (`_apply_filters`); `max_price=1` kept both no-price places. Misleading vs SKILL.md.
  Proof: run above.
- `--radius-km`/`--sort distance` silently ignored when `--near` is an address (only `lat,lng` parses), yet SKILL.md says `--near` takes an address.
  Proof: none, read only.
- Large unwired surface for "follow-ups": `schedule.py`, `parse_directions` + 2 fixtures, `sapisidhash`, `Step`/`DirectionsLeg`/`ItineraryStop`, `phone`/`review_count`. No command reaches them; against Simplicity-first. `polish-pr` candidate.
- SETUP.md points at a `cli/tools/` capture harness that isn't shipped.
- CI: CodeQL red is a 3s job-init failure, not the diff; the load-bearing `Agent · Typecheck and test` is still pending. 20 skill tests pass locally.

**Verdict:** BUGGED — two documented filters (`--open-now`, `--max-price`) don't filter, and search advertises a nonexistent `maps show`; core search/link/route building is sound.

---

emipasca:
Ran an inline polish pass (review + simplify + docs + verify), since the earlier background run stalled at its verify step without pushing.

**Review findings applied (all six from `/code-review high`):**
- **Correctness:** rating extraction read any float in [1.0,5.0] as the star rating, so a real longitude/latitude in that band (Paris 2.35, Singapore 1.35) could be reported as a rating. Now skips coordinate triples; added a regression test.
- **Correctness / no silent no-op:** `--radius-km` and `--sort distance` now error when `--near` is not `lat,lng`, instead of silently returning unfiltered results.
- **Output contract:** `--locale`/`--country` are now a shared parent parser, so they parse after the subcommand (`maps search "..." --locale it-IT`) instead of an argparse error.
- **Elegance / subtraction:** removed unreachable code the agent would read as live behavior (the itinerary scheduler, directions parsing, their models, and ~270KB of directions fixtures). These return in the follow-up that wires the itinerary and directions-detail commands.
- **Efficiency:** replaced the per-node `json.dumps` in `_place_records` with one bottom-up annotation pass, so the search feed is walked in linear time.
- **Docs:** corrected the drift-recovery pointer; no `cli/tools/` harness exists, so `SETUP.md`/`pb.py` now describe recapturing `search_pb.txt` from a live request.

**Verify:** `./check.sh guards` green; the skill suite (14 tests) passes in CI's invocation form.

**CLAUDE.md:** no change. A new skill discovered from disk is not a CLAUDE.md-altitude surface, and the generic "Adding a skill" guidance already covers the pattern.

emipasca:
@vestabot, fix the codeql ci (the comment from github security) then run a polish pass using a fable agent

elyxlz:
Done, pushed to `maps-skill` as two commits.

**CodeQL fix (`client.py:63`).** The alert (`py/incomplete-url-substring-sanitization`, #29) fired on `"consent.google.com" in str(resp.url)`, a substring match on the whole URL. Now `resp.url.host == "consent.google.com"`, an exact host compare via httpx's parsed URL.
Proof: `httpx.URL("https://consent.google.com/x").host` → `'consent.google.com'`. #29 is the only maps-path alert; the other 11 open code-scanning alerts sit in unrelated skills and are out of this PR's scope.

**Fable polish pass (applied, verified, pushed).**
- Dead surface removed: `Place.review_count/phone/open_now` and `ItineraryStop.open_at_slot`, none of which any parser populates.
- Dedup: three copies of the string walker and two website heuristics collapsed to shared `json_strings`/`is_web_url` in `pb.py`.
- Types: itinerary now returns concrete `Itinerary`/`ItineraryLeg` dataclasses (wire shape unchanged) instead of a loose `dict[str, object]`, dropping `format_itinerary`'s isinstance dance.
- Contract: directions time-flag conflicts (`--depart`+`--arrive`, non-transit) now fail before any network call; `httpx` errors and malformed `--stops` (e.g. `[1,2]`) return `{"error": ...}` on stderr instead of a traceback; cache entries are shape-validated on load.
- Docs: `SKILL.md`/`SETUP.md`/CLI notes corrected to match actual output (not every command prints a table; `--from` takes a cid).

**On my earlier review:** the two blocking findings are gone in the current branch. `--open-now`/`--max-price` were dropped entirely (no more dead filters), `--radius-km`/`--sort distance` now error when `--near` isn't `lat,lng` rather than silently no-op, and `maps show` is now a real subcommand.

**Verify:** `uv run pytest -q` → 37 passed. `ruff check` + `ruff format --check` on `agent/skills/maps` clean. No new import cycle. Offline smoke of `route` and the error paths holds. Live search still unverifiable from here (network), gated by the drift canary as before.

CI: watch the re-run of `CodeQL` and `Agent · Typecheck and test` on the new head.

Verdict: MERGE — the scraping path stays fragile by nature, but it is honestly gated, the dead/misleading surface is gone, and the CodeQL finding is fixed at its root.

PR #2140, opened 2026-08-20
fix(user-context): report the timezone the GPS fix falls in across all clients, not the roaming SIM's clock

emipasca:
## Problem

A phone roaming abroad showed the wrong location to the agent. Observed in Sardinia on a UK SIM: the agent thought the user was in London.

The phone's OS clock had correctly updated to Europe/Rome and GPS placed it in Sardinia. The one wrong signal was the device `location` string, which the gateway resolved from the connection IP (ipwho.is): home-routed roaming egresses a UK SIM's data through the UK, so the IP geolocated to London regardless of the phone's real position.

## Changes

**Remove the IP-derived device location entirely.** GPS position and place already give the real location, so the coarse IP guess is dropped rather than demoted. This deletes the geoip module and the ipwho.is lookup, the client-IP capture and resolution plumbing on the sync socket, the stored IP plus location on each device record, and the `location` field from the device wire projection, the roster UI, and the agent's device view. A net deletion of ~274 lines.

**Timezone from the GPS fix, with the OS zone as fallback (all clients).** Every client derives its reported zone from its GPS coordinates with `tz-lookup` (an embedded ~73KB offline boundary table: no network, no native module, same on every platform) when a fix is available, and falls back to the OS zone otherwise. The OS clock is a sound fallback: on a laptop there is no cellular network to mislead it, and on a phone iOS keeps it current from the network and location.

**Web and desktop can share a position too, behind a per-device opt-in.** A new "share this device's location" switch makes the browser (or desktop app) request a geolocation fix and report its position on the existing `client_context` frame. Desktop grants the `geolocation` permission and ships the macOS usage string. The gateway already stored `position` for any device kind, so no wire addition was needed there.

## Compatibility

Removing the `location` wire field clients read is a breaking change, so `MIN_SUPPORTED_CLIENT_VERSION` moves to the in-gap bump above the shipped version (`0.2.5`). The `sync-protocol.json` contract fixture is regenerated.

## Behavior summary

- Zone: GPS-fix zone when a fix is in hand, else the OS zone. Consistent across mobile, web, and desktop.
- Position: mobile and (with opt-in) web/desktop report it; the gateway stores it for any device.
- Place: from the device's own GPS reverse-geocode (mobile); web/desktop have no offline geocoder, so their position names no place.
- The IP-derived `location` is gone from every surface.

## Verification

- `vestad`: clippy clean; 472 unit tests pass; api-contract fixture up to date.
- `apps/core`: tsc plus 377 tests pass.
- `apps/web`: tsc plus 289 tests pass.
- `apps/mobile`: 232 tests pass. `apps/desktop`: tsc plus lint plus 21 tests pass.
- `agent` `test_user_devices_tool.py` passes; ruff clean.


elyxlz:
**Fixes:** nothing linked. Solves a real observed incident: a UK SIM roaming in Sardinia made the iPhone report `Europe/London` (iOS pins the OS clock to the SIM home network) and the IP-derived `location` also said London, so the agent placed the user in the UK. Worth carrying.

**Blocking:** none.

**Non-blocking**
- Two concerns in one PR: a mobile code fix (`device-context.ts`) plus an agent prompt change (`tools.py`). Tightly coupled to the same incident, so acceptable, but CLAUDE.md prefers one concern per PR.
  Proof: none, read only.
- `tz-lookup.d.ts:3` says throws `RangeError`; the lib does throw `RangeError` on out-of-range, so the doc is correct. No action.
  Proof: `node -e tz(200,0)` -> `RangeError invalid coordinates`.

**Verdict:** MERGE. Attacked the tz derivation, the fallback chain (no fix / rejected coords / sharing off), and the field-name accuracy; all held. `tz(39.2238,9.1217)` -> `Europe/Rome`, tz-lookup is pure-JS RN-safe, and the new tests fail if the fix zone stops beating the mocked `Asia/Tokyo` clock.

---
Maintainers: mention \`@vestabot\` to re-check, run \`polish-pr\`, or fix red CI.

emipasca:
@vestabot please review this PR again.

Since the last review the branch removes the IP-derived device `location` entirely (geoip module, ipwho.is lookup, client-IP capture, the stored ip/location, and the `location` wire field), derives the timezone from the GPS fix with an OS-zone fallback across all clients, and adds an opt-in web/desktop position via `navigator.geolocation`. `MIN_SUPPORTED_CLIENT_VERSION` is bumped to `0.2.5` for the wire-field removal, and the `sync-protocol.json` fixture is regenerated. A follow-up commit fixed a mobile `tsc` break from a leftover `device.location` reference.

Worth a close look at: the compat bump (breaking removal of a read field), and whether the `client_context`-carried position needs any new gateway handling (it does not — the gateway already stored position for any device kind).

---
_Generated by [Claude Code](https://claude.ai/code)_

elyxlz:
Re-reviewed after the branch grew to remove the IP `location` entirely and add web/desktop geolocation. I changed nothing (read only).

**Fixes:** nothing linked. Fully closes the roaming double-London bug: both signals are gone (timezone now derived from the GPS fix on every client, IP-derived `location` deleted end to end). Sound.

**Blocking:** none.

**Non-blocking**
- Compat bump is correct, not just plausible: `0.2.5` = current `0.2.4` + patch, so `release.sh`'s guard (refuses `> current+patch`) passes and it is the sanctioned in-gap value.
  Proof: `git show origin/master:vestad/Cargo.toml` -> `version = "0.2.4"`.
- The bump hard-locks every shipped client (`app_behind`) until it updates. Inherent to removing a read field and CLAUDE.md-sanctioned, but worth knowing the old parser actually tolerated an absent `location` (`parse.ts`: `device.location == null` -> null), so the lockout is by policy, not necessity.
  Proof: `parse.ts:159` pre-removal, `undefined == null` is true.
- Desktop geolocation resolves via CoreLocation on macOS (Info.plist string added) but has no provider on Linux/Windows Electron, so the toggle there silently falls back to timezone-only. Graceful, not a crash.
  Proof: none, read only.
- `LocationToggle` (per-device, localStorage) is independent of the gateway-wide `user_context` switch; enabling it with the master off prompts for a fix the gateway then discards. Minor UX overlap.
  Proof: none, read only.

Old `devices.json` with stale `ip`/`location` keys loads fine: `DeviceRecord` has no `deny_unknown_fields`, so no fleet-state wipe on upgrade. Field removal is complete across Rust, all clients, and both fixtures; no dangling reader.

**Verdict:** MERGE. Attacked the compat bump, the field-removal completeness across every reader, the fleet on-disk migration, and the new browser/desktop timezone path; all held.

---
Maintainers: mention \`@vestabot\` to re-check, run \`polish-pr\`, or fix red CI.

PR #2149, opened 2026-08-20
feat(desktop): resolve location through the OS provider on Windows and Linux

emipasca:
Follow-up to #2140 (merged), rebased onto `master`.

## Problem

The desktop "share this device's location" opt-in only worked on macOS. The renderer's `navigator.geolocation` resolves through an OS provider Chromium picks per platform: CoreLocation exists on macOS, but Electron ships no Google geolocation API key, so on Windows and Linux the network provider answers nothing and the toggle silently fell back to timezone-only.

## Changes

**One main-process owner, `apps/desktop/src/geolocation.ts`.** The desktop app resolves the fix natively per platform:

- **Windows**: the in-box WinRT `Windows.Devices.Geolocation.Geolocator`, driven from PowerShell. No MSIX packaging, no API key; respects the Windows location setting. Output is culture-invariant (`lat|lon|accuracy`) so a comma-decimal locale cannot corrupt the numbers.
- **Linux**: GeoClue2 over the system D-Bus via `gdbus` (GetClient, identify as `vesta`, Start, poll the Location property to the fix, Stop). Present on most desktops; a machine without it degrades to timezone-only.
- **macOS**: answers null; the renderer's existing CoreLocation path stays authoritative.

**Preload bridge extension.** A new `readGeolocation` capability crosses `window.vestaNative` (both declarations updated identically); the electron bridge parses the untyped IPC answer at the boundary, the browser bridge carries `null`.

**Web reader preference.** `readBrowserDeviceContext` asks the bridge first; a null answer (macOS, no provider, refused) falls through to the renderer's own geolocation, so every platform degrades to the OS timezone rather than erroring. The zone still derives from the fix with `tz-lookup`.

Settings copy: "the system asks permission the first time" (it is the OS prompt/setting, not always a browser dialog).

## Failure behavior

Every failure resolves to null and the report carries the OS zone alone: missing `gdbus`, no GeoClue agent authorization, the Windows location toggle off, a step timeout (`NATIVE_FIX_TIMEOUT_MS`), malformed output.

## Verification

- `apps/desktop`: tsc, lint, 30 tests (new: Windows line parsing incl. locale-hostile output, gdbus reply parsing, the full GeoClue walk with Stop asserted, macOS answering null, provider-missing degradation).
- `apps/web`: tsc, lint, 292 tests (new: bridge fix preferred over renderer geolocation; null bridge answer falls through).
- Not exercised on real Windows/Linux hardware from this environment: the WinRT script and the GeoClue walk follow the documented interfaces and are covered by fakes, but a hands-on check on each OS is worth doing before release.

### DIFF
--- a/apps/desktop/src/geolocation.test.ts
+++ b/apps/desktop/src/geolocation.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, it } from "vitest";
+import {
+  parseObjectPath,
+  parseVariantNumber,
+  parseWindowsFix,
+  resolveNativeFix,
+} from "./geolocation";
+
+const instantSleep = () => Promise.resolve();
+
+describe("parseWindowsFix", () => {
+  it("parses the invariant lat|lon|accuracy line", () => {
+    expect(parseWindowsFix("39.2238|9.1217|25\r\n")).toEqual({
+      latitude: 39.2238,
+      longitude: 9.1217,
+      accuracyM: 25,
+    });
+  });
+
+  it("keeps the fix when only the accuracy is malformed", () => {
+    expect(parseWindowsFix("39.2238|9.1217|NaN")).toEqual({
+      latitude: 39.2238,
+      longitude: 9.1217,
+      accuracyM: null,
+    });
+  });
+
+  it("rejects malformed output", () => {
+    expect(parseWindowsFix("")).toBeNull();
+    expect(parseWindowsFix("error: denied")).toBeNull();
+    expect(parseWindowsFix("a|b|c")).toBeNull();
+  });
+});
+
+describe("gdbus parsing", () => {
+  it("extracts an object path from a call reply", () => {
+    expect(
+      parseObjectPath("(objectpath '/org/freedesktop/GeoClue2/Client/1',)\n"),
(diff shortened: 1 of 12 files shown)


elyxlz:
**Fixes:** nothing linked (follow-up to merged #2140). Extends the desktop "share this device's location" opt-in from macOS-only to Windows (WinRT Geolocator) and Linux (GeoClue2), which previously fell back to timezone-only because Electron ships no Google geolocation key. It does that.

**Blocking:** none.

**Non-blocking**
- `apps/desktop/src/geolocation.ts:120` GeoClue2 usually needs a registered agent or an allow-listed `.desktop` for a non-packaged app; setting `DesktopId=vesta` with neither can make Linux return no fix on more machines than "present on most desktops" implies. Safe: degrades to null → timezone-only.
  Proof: none, read only. WinRT/GeoClue paths are unverified on real hardware (author flagged this); every failure resolves to null, so the worst case is less data, never wrong data.

**Verdict:** MERGE, resolveNativeFix answers null on macOS and on any error so the renderer's geolocation stays authoritative and every platform degrades to the OS zone; desktop suite 9/9 and web native 12/12 green here (device-context.test env-failed on a missing `tz-lookup`, not the diff).

---

PR #2150, opened 2026-08-20
refactor(user-context): drop the gateway-wide switch for per-device toggles

emipasca:
## Problem

The gateway-wide `user_context` switch ("share device context" in web and mobile Settings) was a second kill switch layered over the per-device location opt-ins, and its off wiped every device at once. The per-device toggles already carry the off semantics that matter: turning a device's location sharing off reports `position: null` on the next edge, which wipes what that device had stored from the store, the roster, and the agents — immediately, since both presence reporters re-report on the toggle edge.

## Changes

Remove the switch end to end, a net deletion (~-185 lines):

- **vestad**: the `settings.json` field (a stale `user_context` key in an existing file is ignored, proven by test), the `GET`/`PUT /gateway/settings` carrier, the report gate in `report_device_context`, and `DeviceRegistry::clear_context` (dead once nothing global wipes).
- **web**: the `UserContextToggle` row; the `GatewaySettings` type. The per-device location toggle stays and its comment now states the off-wipes semantics.
- **mobile**: the "Share device context" row and `changeUserContext` (which also ran the consent flow); the Privacy toggle is now the only owner of `requestLocationSharing`, and the orphaned `locationUndecided` helper is removed with its tests.
- **docs**: the AGENTS.md state-plane paragraph now describes the per-device-only model.

## Semantics after

- The **timezone** is always reported by every device (the low-sensitivity signal agents schedule by); there is no longer a switch that silences it.
- The **position** stays a per-device opt-in, off by default, and changing your mind on a device wipes that device's stored position on the spot.

## Compatibility

Removing `user_context` from the `/gateway/settings` response is a breaking REST change, but `MIN_SUPPORTED_CLIENT_VERSION` already sits at the in-gap `0.2.5` from #2140, and a second break inside the same unreleased gap needs no further bump (per the documented procedure). Old clients PUTting the key are ignored by serde; old `settings.json` files load unchanged.

## Verification

- `vestad`: clippy clean, 471 unit tests, api-contract fixture in sync (the settings shape is not pinned there).
- `apps/web`: tsc + 289 tests. `apps/mobile`: tsc + lint + 231 tests (one fewer suite from the deleted orphan helper).
- prettier + eslint clean on every changed file.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AE3jsuYkwHqBCGNzqGQ891)_


elyxlz:
**Fixes:** nothing linked. Removes the redundant gateway-wide `user_context` kill switch, leaving position governed per device (its off already reports `position: null` and wipes the stored fix) and timezone always reported.

**Blocking:** none.

Attacked on: dead readers of the removed field, the per-device wipe path that now has to carry the whole off-semantics, and the breaking REST change.
- No stale reader survives. `rg user_context` finds only the unrelated `user_context.rs` notification module and `state.user_context`; web `DevicesCard` dropped its `fetchGatewaySettings` import, mobile/desktop/fixtures reference nothing.
  Proof: `rg -n 'user_context|clear_context|UserContextToggle' apps vestad` → only the notification store + the settings parse test.
- Per-device off still wipes: `apps/web/src/lib/device-context.ts:55` returns `position: null` when `shareLocation` is off, and that retracts. Pre-existing, untouched.
  Proof: read only.
- Breaking `/gateway/settings` shape change is covered: `MIN_SUPPORTED_CLIENT_VERSION` is `0.2.5`, repo version `0.2.4`, so it sits in the current unreleased gap; no second bump needed.
  Proof: `settings.rs` has no `deny_unknown_fields`; `cargo test settings_with_a_stale_user_context_key_still_parse` → ok, a stale `{"user_context":false}` still loads.

**Non-blocking**
- Timezone now has no off switch anywhere: previously the gateway toggle silenced it, now every device always reports its zone. Intentional and stated, but it is the one non-deletion behavior change; confirm that is the desired product call.
  Proof: `report_device_context` gate removed at `vestad/src/user_context.rs:106`, no replacement guard.

**Verdict:** MERGE, clean net deletion (~-185), the sole wipe path is verified live and the breaking REST change is inside the min-supported gap.

---

PR #2151, opened 2026-08-20
feat(user-context): share location by default, with the OS grant as the consent

emipasca:
Follow-up to #2150 (merged), rebased onto `master`.

## Problem

Location sharing was off until the user found the per-device toggle, so most devices never reported a position, and the timezone-from-GPS-fix path (#2140) rarely ran.

## Changes

Default the per-device location switch **on** everywhere and make the OS permission prompt the real consent:

- **Web/desktop** (`use-share-location`): absent key now means on; only an explicit "off" stays off. The next focused report reads a fix, so the browser (or on desktop the OS provider from #2149) prompts the first time. A refused grant degrades to timezone-only exactly as before.
- **Mobile** (`preferences/model`): `shareLocation` defaults true. The read path deliberately never prompts, so a new `requestLocationIfUndecided` in `location-consent.ts` (the one owner of the consent flow) raises the OS prompt on the phone's **first shared foreground read** — a phone that already answered, granted or refused, keeps its answer, and the background poll never prompts (it cannot).
- The toggle remains the per-device off switch: off retracts the stored position on the spot (covered by a new web PresenceReporter test).
- Copy and AGENTS.md updated for the default-on model.

## Who flips on

Only devices that never answered: a fresh install, or a web profile that never touched the toggle. A stored preference — on **or** off — is kept, so anyone who deliberately opted out stays out.

## Verification

- `apps/mobile`: tsc + 233 tests (new: `requestLocationIfUndecided` asks only an undecided phone and keeps granted/denied answers; defaults flipped in the model tests).
- `apps/web`: tsc + 293 tests (default-on report shape; the off-retraction path pinned).
- eslint + prettier clean on every changed file. No gateway or wire change.


elyxlz:
**Fixes:** nothing linked (no closing keyword). Flips per-device location sharing on by default so #2140's GPS-timezone path stops being starved by an unfound toggle; the OS grant stays the real consent.

**Blocking:** none.

**Non-blocking**
- Stacked on #2150 (base `claude/per-device-context-toggles`, not `master`). Merge #2150 first and retarget, or this lands #2150's diff onto the wrong base.
  Proof: `gh pr view 2150` → open, base `master`, mergeable; body says "retarget to master once it merges".
- Default flip is a real privacy-default change: every never-touched web profile now raises a geolocation prompt on first focus, and a never-asked phone raises when-in-use + always-on on first foreground read. Intended, but worth a conscious sign-off.
  Proof: `use-share-location.ts:14` `!== "off"`; `PresenceReporter.tsx:61` `requestLocationIfUndecided` → `requestLocationSharing` (foreground+background).

Traced: goal reached on both platforms, every `shareLocation` reader, the off→`position:null` retraction, the background poll (no prompt, degrades without grant), and the never-answered-only flip. No leak without an OS grant; existing opt-outs kept.

**Verdict:** MERGE — sound, tested, no wire change; only the stacking order needs care.

---

PR #2152, opened 2026-08-20
feat(skills): make maps a default skill and install its CLI on update

emipasca:
## What

Makes the new `maps` skill (#2139) active for every agent.

- `agent/core/default-skills.txt` gains `maps`: boot unions the defaults into `active_skills` and rebuilds the skill links, so the skill is discoverable on every box after its next restart, new and existing alike.
- One idempotent prompt migration, `2026-08-install-maps-cli.md` (after-sync, interruptible): `uv tool install --editable ~/agent/skills/maps/cli`, verify the command is on PATH, then one `maps doctor` probe. A doctor failure from a network condition does not block the migration; a failed install stops it unmarked, so it retries next boot.

## Why a migration

Activation only links the SKILL.md; a skill's CLI is installed by its own setup. Fresh agents pre-mark every shipping migration and install lazily from SETUP.md on first use, the same way other default skills with a CLI do; the migration exists to converge the boxes that already ran their birth.

## Tested

Migration- and skill-scoped agent tests pass locally (the two workspace-repair failures are the known host `tag.gpgsign` local-only issue, green in CI).

### DIFF
--- a/agent/core/default-skills.txt
+++ b/agent/core/default-skills.txt
@@ -7,6 +7,7 @@ dashboard
 dream
 email-client
 emergency
+maps
 microsoft
 nap
 notifications
--- a/agent/core/migrations/2026-08-install-maps-cli.md
+++ b/agent/core/migrations/2026-08-install-maps-cli.md
@@ -1,0 +1,29 @@
+---
+interruptible: true
+---
+
+The `maps` skill (find real places, Google Maps links, directions, day plans) is a default
+skill, so it activates on every boot. Its command needs one install on this box.
+
+### 1. Install the `maps` command
+
+```bash
+uv tool install --editable ~/agent/skills/maps/cli
+command -v maps
+```
+
+The last line must print a path under `~/.local/bin`. The install is transactional and safe to
+run more than once. If it fails, STOP, leave this migration unmarked, and report it to the user.
+
+### 2. Verify with one probe
+
+```bash
+maps doctor || true
+```
+
+`doctor` probes each Maps RPC once and names any failing check. A network failure here is not an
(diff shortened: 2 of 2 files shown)


elyxlz:
**Fixes:** nothing linked (no closing keyword; #2139 is the merged PR that added the skill, not an issue). It makes the `maps` skill a default and ships the converge migration to install its CLI on existing boxes. Worth carrying: maps needs no API key (public HTTP), so enabling it fleet-wide costs nothing until used.

**Blocking:** none.

**Non-blocking**
- `2026-08-install-maps-cli.md:29` step 3 hand-writes the mark-applied step, which the runner already appends automatically (`migrations.py:128-131`, "authors never hand-write the name"). Harmless duplication; the shipped `2026-08-email-client-cli.md` does the same, so it matches convention.
  Proof: `migrations.py` `_batch_turn` appends `## Final step ... mark_migration_applied(name="...")` to every migration.

**Verdict:** MERGE, attacked on fresh-agent install, re-run idempotency, phase correctness, and test breakage; all held.
- Fresh box: birth doesn't install maps, but SKILL.md `## Setup` -> SETUP.md is the lazy path, same as other default CLI skills.
- Re-run after crash: `uv tool install --editable` without `--force` exits 0 on a second run (tested locally), so the STOP-if-fails guard never traps an already-installed box.
- Phase: after-sync (the default) is load-bearing, not incidental. `~/agent/skills/maps/cli` only exists after upstream-sync merges the skill onto the box.
- `test_default_skills_are_real_and_include_runtime_deps` passes: `agent/skills/maps/` exists on master.

---

PR #2155, opened 2026-08-20
fix(skills): maps search follows the user's location and country

emipasca:
## What

Two region-correctness fixes for the maps skill, from its first real fleet use (an agent handling "bring me to the nearest frozen yoghurt" in Alghero got 20 US froyo chains):

1. A coordinate `--near` on `maps search` now actually centres the search on that point.
2. `--country` now defaults from the box's timezone instead of hardcoded `us`, so every command follows where the user lives with no flags and nothing for the agent to manage.

## Why / root cause

`--near` reached Google only as literal text appended to the query, which Google ignores for a bare coordinate, and the captured `search_pb.txt` template carries no location, so the only geographic signal was `gl=`, defaulting to `us` (and the natural first call omits `--country`).

## How

**Viewport (`--near`):** the search RPC reads a viewport block in the `pb` parameter, the same mechanism the Maps site uses for "search this area", and it overrides the `gl` country default (verified empirically before writing the fix: same query/token/`gl=us`, US chains without the block, Alghero gelaterie with it). `build_search_pb` takes an optional `near` coordinate and prepends the block; `search()` routes a `lat,lng` `--near` into the pb and keeps it out of the query text; a place-name `--near` keeps riding the query text, which works.

**Country inference:** the agent process exports `TZ` from its configured timezone (`main.py`), so every maps invocation inherits the user's IANA zone, kept fresh fleet-wide by the user-timezone notification flow. tzdata's `zone.tab` maps the zone to one country offline. Chain: `--country` flag > `TZ` (falling back to the `/etc/localtime` symlink) > `zone.tab` > today's `us`. `zone.tab` over `zone1970.tab` is deliberate: devices report per-country zone names (`Europe/Oslo`) that `zone1970.tab` folds into multi-country lines (`DE,DK,NO,SE,SJ -> Europe/Berlin`), while `zone.tab` keeps one country per reported name (`Europe/Oslo -> NO`, `Europe/Vatican -> VA`).

SKILL.md/SETUP.md state the new behavior: pass `--country` only for a query about another country, `--locale` for language.

## Verification

- TDD throughout, every behavior test watched failing first. 58 CLI tests pass; ruff check + format clean.
- Live A: the exact incident command `maps search "frozen yogurt" --near "40.5589,8.3138"` went from Oklahoma Menchie's to Alghero gelaterie.
- Live B: `TZ=Europe/Rome maps search "gelateria"` with no flags returns Rome results.
- Real-`zone.tab` spot check: Rome->it, Oslo->no, Stockholm->se, Vatican->va, New_York->us, UTC->us (fallback).

### DIFF
--- a/agent/skills/maps/SETUP.md
+++ b/agent/skills/maps/SETUP.md
@@ -26,8 +26,10 @@ exits 0 when all pass and 1 with the failing checks named otherwise.
 ## How it works
 
 The skill replays Google Maps' own public web endpoints over plain HTTP (no browser, no account,
-no API key). It reads the country and language from `--locale`/`--country`, so pass the user's
-own values for results and formatting that match where they are.
+no API key). The country defaults from the box's timezone (`TZ`, which the agent's configured
+timezone sets, mapped through tzdata's `zone.tab`), so results and formatting follow where the
+user lives with no flags; `--country` overrides it for a query about another country, and
+`--locale` sets the language.
 
 `search`, `show`, and `itinerary` print a human table by default and structured JSON with
 `--json` / `--json-pretty`; the other commands always print JSON.
--- a/agent/skills/maps/SKILL.md
+++ b/agent/skills/maps/SKILL.md
@@ -25,7 +25,8 @@ answer at request time; null when the place publishes no hours), and a `links` b
 opens the exact place; `directions_url` is a ready directions link). The table shows each
 result's `cid`, which you pass to `show` or `directions`, and marks a currently closed place.
 Filter with `--min-rating`, `--category`, and `--sort rating`.
-`--near` takes an address or `lat,lng`; `--radius-km` and `--sort distance` measure from `--near`,
+`--near` takes an address or `lat,lng`; a coordinate centres the search on that point, so pass
+the user's position when you have it. `--radius-km` and `--sort distance` measure from `--near`,
 so they need it as `lat,lng` (a coordinate, not a place name).
 
 Send the user the `place_url` for a single pick. It opens the exact place on the web and the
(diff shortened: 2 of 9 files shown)


elyxlz:
**Fixes:** nothing linked. Solves a real first-fleet bug: a bare `lat,lng --near` was appended as query text Google drops, so `maps search "frozen yogurt" --near "40.5589,8.3138"` returned US chains instead of Alghero. Worth carrying.

**Blocking:** none.

**Non-blocking**
- `pb.py:96` the viewport is a fixed scale (`_VIEWPORT_SCALE=3000`, `!4f13.1`) independent of `--radius-km`, so a large-radius coordinate search may only surface places inside a town-sized box before the haversine filter runs, returning fewer distant hits than the radius allows.
  Proof: none, read only; live Google ranking not checkable here.
- The empirical claim that prepending `!4m12…` to the pb template overrides `gl=` is the load-bearing fact and cannot be re-verified offline; the two new client tests only pin URL routing (coordinate rides `&pb=`, not `q=`), not that Google honors it.
  Proof: reverted the fix → `test_coordinate_near_becomes_a_viewport_not_query_text` + `test_build_search_pb_near_prepends_viewport` go red; full suite 51 passed.

Untouched paths held: `near=None` returns the template verbatim, place-name `--near` still rides query text (no regression), and lat/lng are not swapped (`!2d{lng}!3d{lat}` ← `_near_latlng` → `(lat,lng)`).

**Verdict:** MERGE — attacked the coordinate/place-name split, the no-near path, and lat/lng ordering; all held. The one uncheckable part is Google's live behavior, which the author verified empirically.

---

PR #2191, opened 2026-08-23
feat(voice): shared voice session model, mobile parity, hands-free plumbing

emipasca:
## What this does

Voice on web and mobile ran on two separate implementations, and mobile had drifted behind: it ignored the `auto_send` setting, never barged in on TTS despite `interrupt_tts` defaulting to true, cut consecutive replies off mid-sentence because it had no TTS queue, and left the audio session in record mode after a session ended. This PR extracts one platform-neutral voice session model into `@vesta/core`, migrates both apps onto it, adds the native plumbing a hands-free (car or headset Bluetooth) session needs, and ships the hands-free conversation UI on mobile. Master is merged in (including the #2126 mobile restructure).

## The shared model (`apps/core/src/voice/`)

- `stt-session.ts`: the /stt/listen socket lifecycle, the Deepgram TurnInfo protocol, transcript and hold accumulation, and a capped pre-open audio buffer. Apps inject only `buildUrl`, a socket constructor, and an `AudioCapture` port.
- `tts-queue.ts`: ordered playback over an injected `SpeechPlayer` (prepare then play), with epoch invalidation on stop and a prefetch cache warmed during the typing-pacing delay.
- `voice-session.ts`: composes the two and owns the behavior decisions: barge-in per `interrupt_tts`, send-vs-draft routing per `auto_send`, hold-to-talk accumulation with one send on release, and the idle auto-stop.
- `status.ts`: the voice service status wire shapes plus the enabled/setting readers, now shared by both apps' settings UIs.

35 new unit tests cover queue order, epoch invalidation, prefetch reuse and fallback, buffering and overflow, turn parsing, accumulation, barge-in gating, auto-send both ways, hold, idle timeout, and error paths.

## Web

`use-voice.ts` becomes a thin zustand view over the core session; `lib/voice.ts` keeps the HTTP fetchers and now exposes the browser ports (worklet capture, WebSocket adapter, prepared-speech streaming). Behavior is unchanged.

## Mobile parity fixes

- `auto_send` off now drafts into the composer instead of always sending.
- A user turn now stops TTS playback when `interrupt_tts` is on.
- Consecutive agent replies queue instead of cutting each other off.
- TTS is prefetched during the typing-pacing delay, so playback starts immediately on reveal.
- The audio session is restored to a playback-only mode when listening ends.

## The hands-free conversation UI (mobile)

A pulse button in the composer (shown when both STT and TTS are ready) opens a docked panel over the composer, about a third of the screen, with the chat still visible above. While it is open:

- One continuous full-duplex session runs: end-of-turn detection splits speech into turns, every turn sends as a normal chat message, and replies speak the moment they arrive (typing pacing is suspended for the panel's lifetime).
- Barge-in is forced on, so speaking over a reply stops it.
- The panel shows the session state (connecting, listening, speaking, paused), the live transcript, a mute toggle, and a large End button, sized for use at arm's length in a car.
- Opening the panel activates the iOS `playAndRecord`/`voiceChat` audio session (echo cancellation plus Bluetooth routing; `UIBackgroundModes: audio` keeps it alive with the screen locked) and the Android microphone-type foreground service, the only way Android keeps a locked-screen mic alive. Ending the panel tears both down.
- A session with no user turn for ten minutes ends itself, bounding battery and transcription spend.

## Verification

- `check.sh guards`, `app-core` (412 tests), `app-web` (293), `app-desktop`, `app-mobile` (300 tests plus the clean-prebuild verify, which exercises the new native module and config through `expo prebuild`) all pass locally after the master merge.
- Not verified on hardware: I could not run a simulator or device from this machine, so the panel's look, the Bluetooth routing, the Android foreground service, and the echo-cancellation behavior need a manual device pass. A visual-qa scenario for the panel needs a voice-socket harness fixture and a capture host, left as a follow-up.
- Echo risk is now mitigated on iOS in conversation mode (the `voiceChat` session mode enables voice processing); Android conversation mode still captures without echo cancellation, so a loud speaker may self-interrupt there. Plain dictation is unchanged.


emipasca:
Superseded. The shared voice session model landed in #2302 (extracted into @vesta/core with today's dictation/conversation semantics and web migrated onto it), and the mobile hands-free conversation mode plus native plumbing is in #2303 (open, awaiting a device pass). The mobile TTS playback-error path this PR's review flagged is fixed in #2303. Thanks for the groundwork here.

PR #2203, opened 2026-08-24
fix(agent): restart prompt ends the boot turn before resuming work

emipasca:
## Problem

Observed on agent axel (2026-08-24): the user restarted the agent mid-build to change the model. The agent woke into its resumed session (whose fresh compaction summary said "NEXT TASK: continue the build") and resumed the build **inside the restart-greeting boot turn**. That turn is non-interruptible and counted in `boot_complete`, so for 6+ minutes:

- the roster showed `booting` ("waking up...") while the logs showed real work,
- `send_preempt` refused delivery (`noninterruptible_turn_active`), so the user's WhatsApp messages queued behind the turn, locking out the user at the exact moment they most want contact (they just restarted the agent).

Nothing was lost (messages queue durably, and the killed pre-restart turn's notification file survives to re-deliver the work), but the greeting turn's length was unbounded and the prompt never told the agent the turn blocks the user.

## Fix

`agent/core/prompts/restart.md` now states the mechanism and the constraint per the prompt guide (mechanism + why, no changelog): the wake-up turn is a boot turn during which the app shows "waking up" and messages queue; recover daemons, orient, end the turn; do not resume in-progress work inside it, because unfinished work re-arrives on its own right after boot.

Core prompts ship inside the engine mount, so every agent gets this on the next release with no migration.

## Tests

No test pins `restart.md` content (`test_processor.py` uses a stand-in). `test_processor.py` + `test_boot_turns.py` pass (41 tests).


elyxlz:
**Fixes:** nothing linked. Solves a real incident (agent axel, 2026-08-24): a user restart mid-build made the agent resume the build inside the non-interruptible restart-greeting boot turn, so for 6+ minutes the roster showed "waking up" and `send_preempt` refused every user message.

**Blocking:** none.

**Non-blocking**
- `agent/core/prompts/restart.md:3` tells the agent "anything unfinished reaches you again on its own right after boot (queued messages, notification files, your task list)". True for message/notification-driven work, but self-driven multi-turn work carried only by the compaction summary ("NEXT TASK: continue the build") has no re-delivery path once the boot turn ends, so a strict reading could drop it. It's a nudge to a capable model, not a hard control, and the cited incident does have a surviving notification file.
  Proof: `clear_notifications` runs only after a turn completes (`loops.py:359`), so an intentional restart mid-turn leaves the file for `monitor_loop` to re-queue; but a build not backed by a live file has no such carrier.

Claims verified: the greeting turn (which carries this prompt via `build_restart_context`, `helpers.py:38`) is queued `interruptible=False` (`main.py:204`) and counted in boot_complete; `send_preempt` refuses while `noninterruptible_turn_active` (`client.py:109`). Prose is mechanism+why with no changelog narration, prompt-guide compliant. No test pins restart.md content.

**Verdict:** MERGE, an accurate, low-risk prompt clarification that matches the runtime mechanism it describes.

---

emipasca:
Folded in the non-blocking finding: work whose only record is the agent's own context (a compaction summary's next step) now gets an explicit carrier, "save it to your task list before you end the turn", instead of the over-claim that everything re-arrives on its own.

PR #2215, opened 2026-08-25
feat(app): web and desktop polish with one owner for user notifications

emipasca:
## Summary

Epic polish pass across the web and desktop apps, built on a gateway-side overhaul that gives user-facing notifications a single delivery owner.

### vestad: one owner for user notifications
- New `user_notifications.rs`: every producer routes through `UserNotifier::notify`, which fans the `user_notification` delta to `/sync` clients, appends to a durable log, and Expo-pushes backgrounded mobile per the `MOBILE_PUSH_ROUTES` table (the whole difference between kinds).
- New `user_notification_log.rs`: an append-only `user-notifications.jsonl` beside `settings.json`, held in memory, paged newest-first; feeds the client notification feed.
- New kinds: agent-injected `needs_user` and `task` (closed set on the agent endpoint), gateway-minted `update_available` (once per discovered release), `agent_status` (stopped/died/recovered from stable status transitions via `notify_status_transition`), and `device_connected` (first sighting of a device id). The last-stable-status map moves from `mobile_app.rs` into `AgentStatusCache` in `agent_status.rs`, and `mobile_app.rs` shrinks to the push facet.
- Agent side: the tasks CLI fires a `task` user notification on the agent's own create/complete (fire-and-forget), core injects `needs_user` on a rate limit, and the `user-notification` script carries the kind through.

### Web app polish
- Navbar `NotificationsPill` backed by the durable feed (`@vesta/core` `notifications-pill` model + `useNotificationsPill`), and a `ToastPill` + toast store replacing sonner.
- Shared orb model in `@vesta/core` (`orb/orb.ts`), consumed by both the web `Orb` and mobile `AgentOrb` so the two render the same visual state.
- Chat refinements: measured content width, collapsible chat state, virtualization polish, markdown bare-URL linkification.
- Agent log streaming moves behind `AgentLogStreamProvider` with session grouping (`log-session.ts`); auto-hiding scrollbars; a shared `Scrim` with hold semantics; agent-subpage routing helper.

### Desktop
- Icons are generated from `icon.png` by `scripts/make-icons.py` (`npm run make-icons`) instead of committing binary `.icns`/`.ico`.

### Contract notes
- All `user_notification` changes are additive (`kind` stays a plain string, unknown kinds render), so `min_supported` does not move.
- `AGENTS.md` updated to describe the new delivery owner and kind routing.

## Test plan
- `./check.sh guards`, `./check.sh web` (core, web, desktop, visual, mobile) green locally.
- Agent: tasks CLI suite (149 passed, incl. new `test_user_notify.py`) and `test_vestad_client.py` green; remaining local failures are macOS-environmental (bash 3.2, skill daemons) and fail identically on master.
- vestad suites are Linux-gated locally; CI runs clippy + unit + integration.


emipasca:
Polish pass on this PR (commit c2d816f): review, simplify, and docs sweep. One commit on top; no author commits touched.

## Pass 1, code review (high, fixed in tree)
- **Dashboard skill build was broken fleet-wide**: the synced ui primitives import `@/hooks/use-scrim-hold`, which did not exist in the dashboard app, and deleting `ui/sonner.tsx` orphaned the dashboard's own `App.tsx` import. Added a signature-identical no-op `use-scrim-hold.ts` shim and a dashboard-local `components/sonner.tsx` wrapper (both outside the strict `ui/` mirror); the dashboard app builds again.
- **Bare-URL linkify matched inside longer TLDs**: `vesta.community` linked to `vesta.com`. Added a `(?![a-z0-9-])` boundary after the TLD alternation in `markdown.tsx`.
- **Pill rotation starved under a steady stream**: the advance timer was keyed on the whole queue, so every enqueue reset the shown item's 2.5s. Now keyed on the head id.
- **`URLSearchParams.size` is not implemented on Hermes**: the feed's query-string gate silently dropped `before`/`limit` for a future mobile consumer; now gates on `toString()`.
- **History error caption**: a failed "load older" no longer captions an already-loaded list with "couldn't load notifications".

Skipped as product/intent calls: `announce_available` not gated on `auto_update` (can notify about a release the nightly auto-update will install anyway; author's call), the ChatComposer "add attachment" button with no handler (reads as a deliberate placeholder).

## Pass 2, simplify (4 lenses: reuse, simplification, efficiency, altitude)
- `Dialog` now uses `useOverlayScrim` (given an `enabled` gate) instead of re-implementing its open tracking inline; popover/dropdown/dialog all share one owner (mirrored to the dashboard by sync).
- Deleted `ToastPillProvider` entirely: the morph width lives beside the queue in the toast store (module scope already owns toast state), removing a provider, a context file, and the local-width fallback.
- One owner for the agent-injectable notification kinds: `agent_injectable_kind` moved into `user_notifications.rs` beside the kind consts and `MOBILE_PUSH_ROUTES`; serve.rs delegates, and the closed-set test moved with it.
- `device_connected` copy moved out of the sync socket handler into `UserNotifier::notify_device_connected`, beside the other gateway-minted producers.
- `observe_transitions` no longer returns the previous status no production caller read.
- NotificationsPill day separators now delegate to the chat's `chat-day-stamp` owner (one rule for "same year drops the year"), computed once per row, with a hoisted `Intl.DateTimeFormat`.
- AgentsCarousel's nearest-card scan extracted once (`nearestCardIndex`), shared by the per-frame pass and the wheel settle.
- Bubble corner radii + tail override extracted to `Chat/bubble-radius.ts`, shared by the real bubbles and the skeletons.
- Chat composer inset now uses the existing `useMeasuredHeight` hook instead of a hand-rolled measure + observer.
- `use-auto-hide-scrollbars` runs the cheap overflow size test before `getComputedStyle` on the hot `pointerover` path.
- `useNotificationsPill` consolidates four ref-mirror pairs into one options ref and returns a stable `dismiss`; the pill provider memoizes its feed, callbacks, and context value so roster deltas stop re-rendering every consumer.
- Trimmed `@vesta/core`'s public index to what apps consume (icons + types); deleted the dead `useNotificationsPillState` export; dropped the web Orb's caller-less `glowSpreadScale`/`glowGradientFade` props; the gradient angle is a precomputed `ORB_GRADIENT_ANGLE_DEG` const.
- `make-icons.py` reads the orb colors from `design/tokens.json` (verified byte-identical to the old literals) instead of re-hardcoding them.
- The pre-rename `rate_limited` compat branches (web, mobile, icon table) now carry `LEGACY(remove-when: ...)` markers per AGENTS.md; comments claiming a mobile pill consumer that does not exist were corrected to "view-independent".

Deliberately left for the author:
- The hidden split-pane `Chat` staying mounted under fullscreen chat (two live chat instances): unmounting it would drop that pane's local draft/scroll, a behavior change.
- `update_available_notified` as separate AppState memory when the durable log could answer "already notified" (and close the restart re-announce hole): a design change, not polish.
- The `push_notifications` per-kind override surface currently has zero client consumers (settings UI presumably next); left as intended API.
- ToastPill/NotificationsPill each hand-roll the same morphing-shell mechanics; real differences exist (tint crossfade, right-anchoring), so no shared shell yet: a third pill should force it out.
- `useMeasuredWidth` duplicates `useMeasuredHeight` line-for-line, and the feed's tiny `record()` helper duplicates `protocol/parse.ts`'s private one: folding either means touching files outside this PR's diff.
- `sonner` + `next-themes` are now unused in `apps/web/package.json` (dashboard app keeps its own copies): dependency removal + lockfile churn left out of the polish commit.
- Subpage routing is decided both in `router.tsx` and `agent-subpage.ts`; deriving the pane from the router would delete the helper but rewrites the author's structure.

## Pass 3, prompt sweep
No prompt surfaces in the diff (no SKILL.md, core prompts, migrations, or MEMORY.md); no-op.

## Pass 4, AGENTS.md freshness
- Corrected the delivery owner's name (`UserNotifier::notify`, not `user_notifications::notify_user`) and completed the sentence with what the PR actually shipped: the durable log (`user_notification_log.rs`, `user-notifications.jsonl`, id-cursored `GET /notifications` behind the clients' notification feed) and the per-kind `push_notifications` overrides in the gateway settings.
- Flagged, not written: the `@vesta/core` bullet could mention the shared orb visual model and notifications-pill model the PR moved into core; borderline altitude, author's call.

## Verify
`app-core` (388), `app-web` (327), `app-desktop` (30), `app-mobile`, and `guards` all green. Agent side: tasks CLI suite 149 passed, `test_vestad_client.py` 5 passed (remaining local agent failures are the known macOS-environmental set, identical on master). vestad checked natively via the revert-safe mac patch: `cargo clippy -p vestad -- -D warnings` clean, `cargo test -p vestad --bins` 479 passed with only the known macOS-only `upstream::tests` bash-3.2 failures; CI covers the real Linux run.

PR #2236, opened 2026-08-27
ci(mobile): build and deliver mobile on GitHub runners

emipasca:
## Why

The release pipeline built iOS + Android on **EAS cloud**, which hit the account's **free-plan monthly iOS build quota** and auto-reverted a release. Cloud builds are also invisible in the GitHub Actions UI (they live on expo.dev), so a failure means digging through a separate dashboard.

## What

Move both mobile builds onto **GitHub-hosted runners** with `eas build --local`, driven by one shared script so a developer runs the identical process locally:

- **New** `apps/mobile/scripts/deliver-mobile.sh` — `eas build --local` for a platform/profile, iOS optionally submits to TestFlight. Build chatter goes to stderr; stdout is only the artifact path (repo output-contract convention).
- **`release-pipeline.yml`**: the single `mobile-delivery` job becomes two parallel jobs — `mobile-ios` on `macos-latest` (build + TestFlight submit) and `mobile-android` on `ubuntu-latest` (build + attach `vesta-android.apk`). `revert-failed-release` now depends on both.
- **Removed** the now-orphaned `apps/mobile/.eas/workflows/mobile-prerelease.yml`.

Signing is unchanged: `eas build --local` fetches the distribution cert + provisioning profile (and the Android keystore) from **EAS remote credentials** via `EXPO_TOKEN`. No p12 or secrets added.

## Effect

- No EAS cloud build quota on the release path (the runner does the compute; free on this public repo).
- Both mobile builds show as normal GitHub Actions jobs, triggered from the Mac via `release.sh` as before.
- EAS Update (OTA) is untouched; only EAS **Build** compute moves.

## Validation

- Spiked `eas build --local` for iOS locally on macOS: auth, remote-credential fetch, prebuild, pods, compile, archive, and code-signing all succeeded. The only local failure was Homebrew's `rsync` shadowing the system `rsync` at the final packaging step (`--extended-attributes` unsupported), which does not occur on GitHub-hosted runners.
- `./check.sh guards` (shellcheck + conventions) passes; script is shellcheck-clean.
- Real end-to-end test is the retriggered `0.2.11` release.


elyxlz:
**Fixes:** nothing linked. Solves the EAS cloud free-plan iOS build quota that auto-reverted a release: moves both mobile builds off EAS cloud onto GitHub-hosted runners via `eas build --local`, keeping remote-credential signing through `EXPO_TOKEN`.

**Blocking:** none. Script is shellcheck-clean, honors the stdout-path/stderr-chatter output contract, the orphaned `mobile-prerelease.yml` is fully removed with no dangling refs, `revert-failed-release` gains both new jobs, and `permissions: contents: write` correctly stays only on the Android job that uploads the release asset.

**Non-blocking**
- The Android `eas build --local` path was never spiked (the PR's validation only exercised iOS on macOS), and `release-pipeline.yml` runs on a published prerelease, so no PR check reaches it. A missing NDK/gradle expectation on `ubuntu-latest` would surface only on the real `0.2.11` release.
  Proof: none, read only; PR body Validation section lists an iOS spike alone.
- [`deliver-mobile.sh:44`](https://github.com/elyxlz/vesta/blob/87f94f810b997dbd40ff4f07b54b76a3af2c3911/apps/mobile/scripts/deliver-mobile.sh#L44) pins `eas-cli@latest`, so each build pulls a different compiler version; matches the prior `eas-version: latest`, so no regression.
  Proof: none, read only.

**Verdict:** MERGE, a sound infra move; the only real risk (un-spiked Android local build) is untestable until the release fires.

CI: release-pipeline jobs do not run on PRs; only the Actions-syntax validator covers this diff. Other checks pending, none caused by this change.

---

PR #2237, opened 2026-08-27
feat(desktop): make app updates manual, in App Settings

emipasca:
## Why

The desktop app auto-updated up-only on launch (silent `autoDownload` + `autoInstallOnAppQuit`). That could push the app **ahead** of the gateway on its own, which flips the `/sync` handshake to the blocking `GatewayBehindScreen` and effectively forces the user to update their gateway, a deliberate action they never chose. Make desktop self-update happen only on an explicit click.

## What

**Desktop main process**
- `updater.ts`: remove the silent launch auto-update. Expose `getAppUpdate()` (check only, no download), `downloadAppUpdate(onProgress)`, and `quitAndInstallUpdate()`. Wired as IPC in `main.ts`, dual-declared in `preload.ts` and `native/types.ts`.
- Signing/feed unchanged: electron-updater still reads the GitHub release feed (stable, up only); only the trigger moves from launch to a click. No EAS, no new secrets.

**Client**
- New `appUpdate` native bridge (electron impl, browser no-op) and a shared `useAppUpdate` hook (`{ supported, status, phase, percent, check, start, relaunch }`).

**Web UI**
- Remove `UpdatePill` from `StatusPill`, so no update indicator lives in the navbar (by request).
- New **Updates** card in App Settings: a "Vesta Desktop" row (desktop only, inline download progress then "relaunch to finish") and a "gateway" row (reuses the existing `UpdatePill`), plus a combined "check for updates" action. `CheckForUpdates` is removed from the navbar and deleted (its role folds into the card).
- `AppBehindScreen`: adds an "update Vesta Desktop" button (it previously relied on the silent background update). `GatewayBehindScreen` keeps its inline "update gateway" action as the guided step after the app lands ahead.

**Discoverability:** none in the chrome, by design. Updates are found in App Settings, or forced by a mismatch screen.

## Checks

`./check.sh app-desktop` and `./check.sh app-web` pass locally (lint, tsc, 32 + 357 tests). The updater test now covers the manual `getAppUpdate` (available / not / fail-closed) instead of the old background flow.


elyxlz:
**Fixes:** nothing linked. Makes desktop self-update manual (click-driven, App Settings "Updates" card + AppBehindScreen) instead of the silent launch auto-download, so the app can no longer push itself ahead of the gateway into the blocking `GatewayBehindScreen`. Coherent problem, worth carrying.

**Blocking:** none. Traced the singleton-across-IPC assumption (module + electron-updater cache persist download→install state), the fail-closed check paths, the preload/`types.ts` wire-contract parity (4 new methods match), and confirmed no dangling refs to the removed `checkForAppUpdate`/`CheckForUpdates`; every `GatewayProvider` field `UpdatesCard` reads exists.

**Non-blocking**
- [`updater.ts`](https://github.com/elyxlz/vesta/blob/8f27835dcd3265cf31d3f5511f0ab66a41105d07/apps/desktop/src/updater.ts#L84-L214) new `downloadAppUpdate`/`quitAndInstallUpdate` and the linux `(received/total)*100` progress math have no test; `updater.test.ts` covers only `getAppUpdate`.
  Proof: none, read only; couldn't run vitest (node v18 here, rolldown needs 20+).
- `useAppUpdate` checks on every mount (AppBehindScreen + UpdatesCard), vs the prior launch-only check. Minor.

**Verdict:** MERGE, well-scoped desktop/web change; attacked the IPC-state and contract-parity seams and it held.

---

PR #2239, opened 2026-08-27
fix(ci): drop the Enterprise-only TestFlight changelog from mobile delivery

emipasca:
## Why

Both v0.2.11 delivery attempts reverted on the iOS job. The build succeeded every time; `eas submit` then died with the generic `Error: GraphQL request failed.`. Reproducing the submit with `EXPO_DEBUG=1` surfaced the real server error:

> `[GraphQL] This submission specified the 'changelog' parameter. Changelog submission is currently available for Enterprise plan only.`

`deliver-mobile.sh` passed `--what-to-test "$CHANGELOG"`, which sets TestFlight's "What to test" through EAS Submit — an Enterprise-plan feature. An A/B submit of the same archive with and without the flag confirms it: without → scheduled; with → this exact failure. The previous EAS cloud workflow passed the changelog only as the *build* message (a dashboard annotation), never as `--what-to-test`, so nothing is lost by dropping it.

## What

- `deliver-mobile.sh`: remove the `--changelog` flag and the `--what-to-test` submit argument (a comment records the constraint so it does not come back).
- `release-pipeline.yml` `mobile-ios`: drop the release-body changelog extraction; call the script without it.

### DIFF
--- a/.github/workflows/release-pipeline.yml
+++ b/.github/workflows/release-pipeline.yml
@@ -308,10 +308,7 @@ jobs:
         working-directory: apps/mobile
         env:
           EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
-          RELEASE_BODY: ${{ github.event.release.body }}
-        run: |
-          CHANGELOG=$(printf '%s\n' "$RELEASE_BODY" | sed -n '/<!-- whats-new -->/,/<!-- \/whats-new -->/p' | sed '1d;$d')
-          ./scripts/deliver-mobile.sh --platform ios --profile production --changelog "$CHANGELOG" --submit
+        run: ./scripts/deliver-mobile.sh --platform ios --profile production --submit
 
   mobile-android:
     name: Mobile · Build and attach Android APK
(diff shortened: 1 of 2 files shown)


elyxlz:
**Fixes:** nothing linked, unblocks mobile delivery: `eas submit --what-to-test` is Enterprise-only and failed the iOS TestFlight submit, reverting v0.2.11 three times (`git log`: bump/revert cycles `bff88d0f`, `cb06c6be`, `16a4dda2` each reverted).

**Blocking:** none.

**Non-blocking**
- Dropping `--what-to-test` removes the "What to test" note testers see in TestFlight. The PR states the prior EAS cloud workflow never set it either, so no regression, just no gain.
  Proof: `git show 263692e:apps/mobile/scripts/deliver-mobile.sh` submit block passes only `--platform/--path/--profile/--non-interactive`; no changelog path remains.

Sole `--changelog` caller was `release-pipeline.yml:314`, updated in the same PR; the android caller (`:342`) never passed it, and the removed `RELEASE_BODY` env was used only for that extraction. Head script has no dangling `CHANGELOG`; comment is 2 lines, usage string updated.

**Verdict:** MERGE, surgical CI fix that removes the exact flag the reverts trace to, no other caller affected.

---

PR #2240, opened 2026-08-27
ci(mobile): give the Android build a real budget and a gradle cache

emipasca:
## Why

v0.2.11 attempt 4: iOS built and submitted to TestFlight (the #2239 fix works), but the Android job hit its **60-minute `timeout-minutes` mid-gradle** — a cold RN release build overruns it on the hosted runner. The cancellation ended the run as `cancelled` (so the revert was skipped and the release survived), but `vesta-android.apk` never got attached.

## What

- `release-pipeline.yml` `mobile-android`: `timeout-minutes` 60 → 120, plus an `actions/cache` on `~/.gradle` (keyed on `apps/package-lock.json`) so warm builds land far under the old budget.
- `mobile-build.yml` (manual dispatch): reworked onto the same local-build path — `deliver-mobile.sh` on a hosted runner (macOS for iOS, ubuntu for Android), with a `ref` input and an `attach_to_release` input so one dispatch can rebuild a platform for an existing release. The old job queued an EAS **cloud** build, which the plan's quota blocks.

## Follow-up once merged

Dispatch `mobile-build.yml` with `platform=android, profile=android-internal, ref=v0.2.11, attach_to_release=v0.2.11` to attach the missing APK to the live prerelease.

### DIFF
--- a/.github/workflows/mobile-build.yml
+++ b/.github/workflows/mobile-build.yml
@@ -1,5 +1,11 @@
 name: Mobile build
 
+# Manual mobile delivery on GitHub-hosted runners through the same
+# apps/mobile/scripts/deliver-mobile.sh the release pipeline runs (eas build
+# --local, signing from EAS remote credentials). Dispatch with a tag ref plus
+# attach_to_release to rebuild one platform for an existing release, e.g. after
+# a delivery job died mid-pipeline.
+
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


elyxlz:
**Fixes:** nothing linked. Unblocks Android APK delivery for the v0.2.11 release: the `mobile-android` release job hit its 60-min cap mid-gradle and never attached `vesta-android.apk`. This raises the cap to 120 and adds a `~/.gradle` cache. Real problem, right layer.

**Blocking:** none.

**Non-blocking**
- `mobile-build.yml` build step: `[ "${{ inputs.submit }}" = "true" ] && args+=(--submit)` looked like a `set -e` trap for the false default, but it is not the last line so the step survives.
  Proof: `bash -eo pipefail` with `[ false = true ] && args+=(--submit); echo reached` → prints `reached`, exit 0.
- Both cache keys use `hashFiles('apps/package-lock.json')`, a weak proxy for gradle deps (npm lock ≠ gradle deps), but the `gradle-${{ runner.os }}-` `restore-keys` prefix makes any prior cache a warm start, so correctness holds; only freshness is approximate.
  Proof: none, read only.
- Whether a cold build fits under 120 min is unverifiable here; the doubling plus cache is a reasonable bet, not a guarantee.
  Proof: none, read only.

Ground truth: `deliver-mobile.sh` writes chatter to stderr and the artifact path alone to stdout, so `artifact=$(...)` is clean; `android-internal` exists in `eas.json`; head YAML parses. Master still has timeout 60 + no cache, so not already fixed.

**Verdict:** MERGE — targeted CI budget/cache fix, `deliver-mobile.sh` reuse mirrors the proven release-pipeline pattern, no defect found.

CI: only the workflow-relevant jobs run (Validate GitHub Actions, lint, merge-gate-ci all pass); the rest are path-skipped. `mergeStateStatus` BLOCKED is branch protection, not the diff.

---

PR #2241, opened 2026-08-27
ci(release): revert on a cancelled delivery job too

emipasca:
## Why

The release pipeline's atomicity has a hole: `revert-failed-release` is guarded by `if: ${{ failure() }}`, but a job that overruns its `timeout-minutes` concludes **cancelled**, not failed — `cancelled()` becomes true, `failure()` does not. v0.2.11 attempt 4 hit exactly this: the Android job timed out mid-gradle, the revert was skipped, and a half-delivered release stayed up (recovered forward by hand).

## What

Guard the revert with `failure() || cancelled()`. Jobs guarded on `cancelled()` still run after a run cancellation (GitHub's documented cleanup pattern), so both a timed-out delivery job and a manually cancelled release run now revert cleanly, restoring "no release unless every job is good" with no exceptions.

### DIFF
--- a/.github/workflows/release-pipeline.yml
+++ b/.github/workflows/release-pipeline.yml
@@ -364,7 +364,9 @@ jobs:
         mobile-ios,
         mobile-android,
       ]
-    if: ${{ failure() }}
+    # cancelled() too: a job that overruns timeout-minutes concludes cancelled,
+    # not failed, and a half-delivered release must revert either way.
+    if: ${{ failure() || cancelled() }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
     permissions:


elyxlz:
**Fixes:** nothing linked. Closes the release-pipeline hole where a delivery job overrunning `timeout-minutes` concludes *cancelled*, not *failed*, so the old `if: failure()` skipped the revert and left a half-delivered release up (the v0.2.11 attempt-4 incident).

**Blocking:** none.

**Non-blocking**
- [`release-pipeline.yml:369`](https://github.com/elyxlz/vesta/blob/9685488a5d67b61ce25ee4bc6c183057f47c9b62/.github/workflows/release-pipeline.yml#L369): whether this fixes the exact reported case rests on `cancelled()` returning true downstream for a *per-job* timeout. GitHub docs define it only as "the workflow was canceled" and say nothing about job-level timeouts; I can't run Actions to confirm. The change is strictly safe (additive, no path now wrongly skips revert) and the idiom is well-chosen: `!success()` would misfire on an intentionally mobile-skipped release, this does not. Real confirmation is the next timeout.
  Proof: docs say `cancelled()` = "workflow was canceled"; per-job-timeout propagation undocumented; read only.
- Behavior change: manually cancelling a release run now auto-reverts (tag + version bump). Author states this is intended.
  Proof: `failure() || cancelled()` is true on run cancellation.

**Verdict:** MERGE, strictly-safe additive guard with the correct idiom; only residual is `cancelled()`-on-timeout behavior GitHub does not document and I cannot test here.

---

PR #2243, opened 2026-08-27
feat(web): autofocus the chat composer when chat is visible

emipasca:
## What

On the web and desktop app, the chat composer now takes focus as soon as the chat is the visible surface, so you can type right away without clicking into the field.

## How

One effect in `Chat` (`apps/web/src/components/Chat/index.tsx`) focuses `textareaRef`:

- **On mount**: initial load, agent switch, expanding the collapsed chat panel, opening fullscreen chat.
- **On subpage close**: returning from logs or settings. Those panes stay mounted (hidden), so a mount-time `autoFocus` never refires; the effect re-runs on the `pathname` change instead.
- **Mobile is excluded**: autofocus there would raise the keyboard on every open.
- **`preventScroll: true`**: no list jump when focus lands.

A fullscreen `Chat` and a split-panel `Chat` can both be mounted; the hidden one sits under `visibility: hidden`, where `focus()` is a browser no-op, so only the on-screen instance takes focus. A signed-out disabled textarea silently refuses focus.

## Testing

`./check.sh app-web` green (tsc, eslint, prettier, 356 tests).

Note: the composer now shows its focus ring on chat open, so the visual QA chat gallery shots pick up the ring on next capture. No drive assertions change.

### DIFF
--- a/apps/web/src/components/Chat/index.tsx
+++ b/apps/web/src/components/Chat/index.tsx
@@ -8,6 +8,7 @@ import {
   type ChangeEvent,
   type KeyboardEvent,
 } from "react";
+import { useLocation } from "react-router-dom";
 import { ArrowDown } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import { Card } from "@/components/ui/card";
@@ -25,6 +26,7 @@ import { ChatComposer } from "./ChatComposer";
 import { ChatHeaderActions } from "./ChatHeaderActions";
 import { ChatMessageArea, type ChatScrollHandle } from "./ChatMessageArea";
 import { useChatKeyboardFocus } from "./use-chat-keyboard-focus";
+import { agentSubpage } from "@/lib/agent-subpage";
 import { TRIM_HISTORY_SETTLE_MS, agentNeedsUser } from "@vesta/core";
 
 // Breathing room between the last bubble and the floating composer, folded into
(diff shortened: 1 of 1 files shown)


elyxlz:
**Fixes:** nothing linked. Adds one behavior: focus the web/desktop chat composer whenever chat becomes the visible surface (mount, agent switch, subpage close), skipped on mobile.

**Blocking:** none.

**Non-blocking**
- The real change is 12 lines in [`Chat/index.tsx:95-99`](https://github.com/elyxlz/vesta/blob/2201e8d3f9169166657cab89f8942028db43b8da/apps/web/src/components/Chat/index.tsx#L95-L99). The other ~840 diff lines (ChatMessageArea rewrite, `scroll.ts`, `use-chat-scroll.ts`, `chat-stream-model.ts`, mobile ChatPage, package-lock) are #2228 (`b0b2e619`) already on master: the branch was cut from `203d4048`, before it. On head those files are byte-identical to master.
  Proof: `git diff origin/master head -- <each>` → SAME as master for all five; only `Chat/index.tsx` differs.
- Not mergeable: one real conflict, in `Chat/index.tsx`, because #2228 rewrote the same file. A rebase onto master drops the phantom bloat and leaves only the autofocus hunk to graft.
  Proof: `git merge --no-ff head` on origin/master → `CONFLICT (content): apps/web/src/components/Chat/index.tsx`, sole conflicted path.
- No test covers the new effect (focus-on-visible, mobile skip, subpage-close refire); the 356 green tests are unchanged behavior.
  Proof: none, read only.

**Verdict:** MERGE, the autofocus logic is sound (deps `[isMobile,pathname,name]` complete, `agentSubpage`/`isMobile`/`name` all in scope, `focus()` is a no-op on the hidden instance), but it must be rebased onto master first to clear the one conflict and strip the #2228 duplication.

---

PR #2244, opened 2026-08-27
perf(web): window the log viewers to bound the DOM

emipasca:
## Problem

The web log viewers loaded far too many rows into the DOM. The agent **Console** renders every buffered line as a `<div dangerouslySetInnerHTML>`, up to the 5000-line scrollback, so opening the logs mounts thousands of nodes at once. The **GatewayLogsViewer** is worse: it appended every line with no cap, so its buffer grew without bound while following the tail.

## Fix

Render only a tail window, mirroring chat's no-virtualization strategy. Because logs have no server cursor (the lines are already fully in the client buffer), "load an older page" collapses to "render more of the buffer I already hold" — synchronous, no backend change.

- **`lib/log-window.ts`** (new, unit-tested): pure window decisions (at-bottom, grow trigger, grow step) and constants — base/step 300 lines, 3-screen preload, 80px at-bottom, 30s settle.
- **`lib/use-log-window.ts`** (new): shared hook owning the window size, scroll handler, follow-on-append, and the settle-trim timer. Both viewers use it.
- **`Console`**: renders `lines.slice(-visibleCount)`; removed its bespoke pin/follow logic.
- **`GatewayLogsViewer`**: also fixed its uncapped buffer (added the shared 5000-line cap) and gave rows stable ids.

## Design note

The scrollers leave the browser's native `overflow-anchor` enabled (the Console already did), so a window grow, a trim, and the rolling-cap front-drop all lean on native scroll anchoring — no manual restore/latch machinery needed. The 3-screen preload keeps the grow away from `scrollTop: 0`, where native anchoring stops holding.

Mobile is unaffected: it already windows through `FlatList`.

## Tradeoff

A text-selection drag now spans only mounted rows. Grow-ahead-on-scroll-up plus trim-only-after-bottom-settle keep rows mounted during an upward drag, and the Console's Copy reads the buffer (not the DOM), so copy stays whole.

## Testing

`./check.sh app-web` green: format, types, lint (0 errors), 364 tests pass (incl. 7 new for `log-window`).

### DIFF
--- a/apps/web/src/components/Console/index.tsx
+++ b/apps/web/src/components/Console/index.tsx
@@ -1,17 +1,9 @@
-import {
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useRef,
-  useSyncExternalStore,
-} from "react";
+import { useEffect, useRef, useSyncExternalStore } from "react";
 import { useLayout } from "@/stores/use-layout";
 import type { LogStreamState } from "@/lib/log-session";
+import { useLogWindow } from "@/lib/use-log-window";
 import { useAgentLogSession } from "@/providers/AgentLogStreamProvider";
 import { cn } from "@/lib/utils";
-
-// How close to the bottom still counts as pinned for follow-on-append.
-const AT_BOTTOM_THRESHOLD_PX = 80;
 
 const STREAM_NOTICE: Record<Exclude<LogStreamState, "live">, string> = {
   stopped: "— agent stopped —",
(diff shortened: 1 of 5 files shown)


elyxlz:
**Fixes:** nothing linked. Bounds the DOM of the two web log viewers: the Console mounted up to the full 5000-line scrollback, and `GatewayLogsViewer` grew its `lines` buffer without any cap. Real problem, real fix.

**Blocking:** none.

**Non-blocking**
- [`lib/log-window.ts:41`](https://github.com/elyxlz/vesta/blob/25e71bca43e5d1798b1660cf1e21808ae522495a/apps/web/src/lib/log-window.ts#L34-L41): growth is driven only by `onScroll`, so a fast fling to `scrollTop: 0` can outrun the 3-screen preload; a grow then prepends 300 rows at the top where `overflow-anchor` no longer holds, and the view jumps. The design note already concedes this window; acceptable, worth knowing.
  Proof: none, read only (no browser here).
- Buffer cap and window are independent bounds (`LOG_SCROLLBACK_LINES` 5000 in memory, `visibleCount` base 300 in DOM); both hold, goal reached end to end.
  Proof: [`GatewayLogsViewer/index.tsx:77`](https://github.com/elyxlz/vesta/blob/25e71bca43e5d1798b1660cf1e21808ae522495a/apps/web/src/components/GatewayLogsViewer/index.tsx#L74-L80) slices to the cap on every append; [`use-log-window.ts:64`](https://github.com/elyxlz/vesta/blob/25e71bca43e5d1798b1660cf1e21808ae522495a/apps/web/src/lib/use-log-window.ts#L57-L68) grows only while unpinned so it never fights an upward scroll.

Traced: follow-on-append writes scroll only when `atBottomRef` is true; grow uses a functional `setVisibleCount` so batched scroll events accumulate; trim timer isn't reset by appends; gateway rows now key on `line.id` not array index. No orphaned refs to the removed constants/refs.

CI: could not run `check.sh app-web` here (local Node too old for this vitest — unrelated `styleText` import error); author reports it green with 7 new tests. Watch that job on the PR page.

**Verdict:** MERGE — attacked the windowing and follow/trim logic, it held; only a documented fast-fling jump remains, and Copy reads the buffer so no data is lost.

---

PR #2246, opened 2026-08-28
refactor(web): remove tooltips from home page agent orbs

emipasca:
## What

Remove the hover tooltip from the agent orbs on the home page.

Each orb on the home page (`AgentCard`, rendered through `Home/AgentsCarousel`) was wrapped in a `Tooltip` that showed the agent status label on hover. The orb already exposes that same information through its `label` aria attribute, and the card renders the agent name below it, so the tooltip added no information.

## Change

- Drop the `Tooltip`/`TooltipTrigger`/`TooltipContent` wrapper around the `Orb` in `AgentCard`.
- Keep the orb's accessible `label` and the surrounding `span`, so the accessibility surface is unchanged.

No other orb usages (agent island, navbar) are touched.

## Verification

- `./check.sh app-web` passes (lint, format, tsc, 357 vitest tests).

https://claude.ai/code/session_01Un1w1HZP5ST7YfgizjUh7E

### DIFF
--- a/apps/web/src/components/AgentCard/index.tsx
+++ b/apps/web/src/components/AgentCard/index.tsx
@@ -6,11 +6,6 @@ import {
   CardTitle,
 } from "@/components/ui/card";
 import { Spinner } from "@/components/ui/spinner";
-import {
-  Tooltip,
-  TooltipContent,
-  TooltipTrigger,
-} from "@/components/ui/tooltip";
 import { Orb } from "@/components/Orb";
 import type { AgentRow } from "@/lib/types";
 import { useNavigate } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


elyxlz:
**Fixes:** nothing linked. Removes the hover tooltip that duplicated the orb status already carried by the orb's aria `label` and the visible card name.

**Blocking:** none.

**Non-blocking:**
- Idle status text now only exists in the orb `aria-label` ([`Orb/index.tsx:109`](https://github.com/elyxlz/vesta/blob/a108a73bff99ed5fbb66ad3f97aeb74eb657a355/apps/web/src/components/Orb/index.tsx#L109)); a sighted mouse user loses the on-hover status label while idle. This is the intended trade, not a defect.
  Proof: none, read only.

**Verdict:** MERGE. Wrapper removal is complete, `label` is still consumed by the orb and the inflight/error description ([`AgentCard/index.tsx`](https://github.com/elyxlz/vesta/blob/888363d03ae3b1d3005b56e8e4f36476c8950019/apps/web/src/components/AgentCard/index.tsx)), no test asserted the tooltip, and accessibility is preserved by the orb aria label.

---

PR #2247, opened 2026-08-28
fix(reminders): give reminder notifications a meaningful type

emipasca:
## The smell

Reminder notifications were written with a constant `type="reminder"` on `source="reminders"`. The `type` just restated the `source` and carried no information. Meanwhile the one real distinction, a reminder firing on schedule against a one-shot replayed late (because its time passed while the daemon was down), was smuggled into a `missed:true` payload flag.

The tasks daemon already set the precedent the other way: it recently moved off `type=reminder` to `type=task_due` (migration `2026-08-task-due-notification-type.md`).

## The fix

Name the event in the `type`:
- On-schedule fire: `type=reminder_due`
- Late replay: `type=reminder_missed` (the `missed:true` flag is gone; the type says it)

`source=reminders` notifications now carry a `type` that earns its place, and a `notifications` rule can route on-schedule and missed reminders separately.

## Fleet compatibility

`type=reminder` is now written by no skill (tasks moved to `task_due` earlier; reminders move here). A prompt migration (`2026-08-reminder-notification-type.md`) restarts the reminders daemon, re-points user notification rules keyed on the old pairing, and sweeps agent notes. The old catch-all `type=reminder` caught both on-schedule and missed reminders, so the migration explains that preserving that behavior means one rule for each new type.

## Tests

- Reminders CLI suite updated for the new type strings and filename globs; **167 passed** locally.
- The one local `test_status_rejects_a_reused_pid[reminders]` failure is a pre-existing macOS artifact (no `/proc`, so the pid record has no starttime field); it fails identically on clean `master` and passes on CI's Linux. Unrelated to this change.

### DIFF
--- a/agent/core/migrations/2026-08-reminder-notification-type.md
+++ b/agent/core/migrations/2026-08-reminder-notification-type.md
@@ -1,0 +1,40 @@
+The reminders daemon emits two notification types with `source=reminders`: `type=reminder_due`
+when a reminder fires on schedule, and `type=reminder_missed` when a one-shot is replayed on
+restart because its time passed while the daemon was down. Skill code that ran before this boot
+emitted both as `type=reminder`, so anything on this box keyed to `source=reminders` with
+`type=reminder` now points at a type the daemon no longer writes. If the reminders skill is not
+active on this box, every step below is a no-op.
+
+### 1. Restart the reminders daemon
+
+The daemon loaded its code when it started, so until it restarts it keeps emitting the old type:
+
+```bash
+reminders daemon restart
+```
+
+### 2. Update notification rules that target the old pairing
+
+```bash
+notifications list
+```
+
+A rule with `type=reminder` that catches the reminders skill (either `source=reminders` or no
+`source`) no longer matches anything. The old `type=reminder` caught every reminder, on-schedule
+and missed alike, so decide what the rule was for and re-create it with `notifications remove <id>`
+then `notifications add` (same action and conditions, `--before`/`--after` to keep its position):
+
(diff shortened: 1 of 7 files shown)


elyxlz:
**Fixes:** nothing linked. Replaces the do-nothing `type=reminder` (which just restated `source=reminders`) with `type=reminder_due` on-schedule and `type=reminder_missed` on late replay, retiring the `missed:true` flag; mirrors the tasks skill's earlier `task_due` move.

**Blocking:** none.

**Non-blocking**
- Both `write_reminder_notification` call sites verified: on-schedule fire ([`commands.py:294`](https://github.com/elyxlz/vesta/blob/1c510e342decac18935518463b069636ef31f696/agent/skills/reminders/cli/src/reminders_cli/commands.py#L294)) takes the `reminder_due` default, missed replay ([`commands.py:349`](https://github.com/elyxlz/vesta/blob/1c510e342decac18935518463b069636ef31f696/agent/skills/reminders/cli/src/reminders_cli/commands.py#L349)) passes `reminder_missed`. No core code or other skill keys on the old `type`/`missed`; `missed` in `notification.py:63` is only a docstring naming example.
  Proof: `git grep -niE "reminders-reminder|type=reminder\b|missed"` over `agent/` minus the changed CLI returns nothing that reads it.
- Migration is a new file (append-only), gated on the reminders skill being active, and its step-3 grep `reminder\b` correctly excludes the new `reminder_due`/`reminder_missed` types.
  Proof: `uv run pytest tests/ -q` on head `1c510e3` → `167 passed`.

**Verdict:** MERGE, meaningful type replaces a redundant one, both writers correct, fleet migration present, suite green.

---

PR #2250, opened 2026-08-28
fix(web): dismiss open overlays when clicking the titlebar drag region

emipasca:
## Problem

In the desktop app, clicking the navbar while the agent island (or any popover/dropdown/dialog) is open drags the window instead of dismissing the component.

Two holes in the drag-region rule in `apps/web/src/index.css`:

1. The navbar nests sibling `data-drag-region` divs, and `[data-drag-region]:has([aria-expanded="true"])` only un-drags **ancestors** of the open control. The left/right half divs (most of the navbar surface) stayed draggable, and a drag region consumes the mousedown at the window level, so the page never received the pointerdown that outside-click dismissal listens for.
2. Portaled overlays never matched at all: the island modals are controlled dialogs with no trigger in the navbar, and Radix portals content to `document.body`, outside every drag region.

## Fix

Root-scope the rule: while anything dismissable is open, **every** drag region turns no-drag, so the titlebar click reaches the page and the normal outside-click dismissal fires. Conditions:

- `[data-drag-region] [aria-expanded="true"]`: the expanded island or a menu opened from the titlebar
- `body > [data-state="open"]`: an open Radix dialog/alert-dialog/drawer (overlay and content portal as direct body children and unmount on close)
- `body > [data-radix-popper-content-wrapper]`: an open popover/dropdown/select (the wrapper only exists while open)

Once the component closes, the navbar drags again, matching the native macOS titlebar-click-dismisses-popover behavior. Deliberate side effects: a modal dialog blocks navbar dragging while open (the click closes it via the overlay instead), and the delete confirm still ignores outside clicks because alert dialogs never dismiss on outside click.

## Testing

- `./check.sh app-web` green (lint, format, tsc, 357 tests)
- Drag regions only exist in the real Electron window, so please confirm the dismiss behavior in the desktop app

https://claude.ai/code/session_01BXZyXa2W3Ruv7wDux2XSvE

### DIFF
--- a/agent/skills/dashboard/app/src/index.css
+++ b/agent/skills/dashboard/app/src/index.css
@@ -46,10 +46,18 @@ html[data-platform="android"] {
 .desktop [data-drag-region] [role="button"] {
   -webkit-app-region: no-drag;
 }
-/* While a popover-like control in the titlebar is open (the expanded agent
-   island), the navbar stops dragging so an outside click on it dismisses,
-   the way a native popover closes when you click the title bar. */
-.desktop [data-drag-region]:has([aria-expanded="true"]) {
+/* While anything dismissable is open (the expanded agent island or another
+   control in the titlebar, or any portaled overlay: dialogs put a
+   data-state="open" element directly under body, popovers/dropdowns a popper
+   wrapper), every drag region stops dragging so a click on the titlebar
+   reaches the page and dismisses it, the way a native popover closes when
+   you click the title bar. */
+.desktop:has(
+    [data-drag-region] [aria-expanded="true"],
+    body > [data-state="open"],
+    body > [data-radix-popper-content-wrapper]
+  )
+  [data-drag-region] {
   -webkit-app-region: no-drag;
 }
 
--- a/apps/web/src/drag-regions.test.tsx
+++ b/apps/web/src/drag-regions.test.tsx
@@ -1,0 +1,41 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { Dialog, DialogContent, DialogTitle } from "@/components/ui/dialog";
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+} from "@/components/ui/dropdown-menu";
+
+// The desktop drag-region dismissal rule in index.css matches open overlays
+// with `body > [data-state="open"]` and `body > [data-radix-popper-content-wrapper]`,
+// relying on Radix portals rendering them as direct body children (their
+// internal Portal uses asChild). Pin that DOM shape across Radix upgrades.
+describe("portaled overlay DOM shape (index.css drag-region rule)", () => {
+  it("open dialog puts data-state=open elements directly under body", () => {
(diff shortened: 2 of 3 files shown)


elyxlz:
**Fixes:** nothing linked. Solves navbar clicks dragging the window instead of dismissing the open agent island.

**Blocking**
- [`apps/web/src/index.css:54-55`](https://github.com/elyxlz/vesta/blob/1221f3bb49135edbf8a2341d196f7aca8d11ec62/apps/web/src/index.css#L52-L58) `body > [data-state="open"]` and `body > [data-radix-popper-content-wrapper]` never match: Radix `Portal` wraps its children in a `Primitive.div` inside `body`, so the overlay/content and the popper wrapper are **grandchildren** of `body`, not direct children. Both new conditions are dead, so dialogs and page-level popovers/dropdowns do not dismiss on titlebar click, contrary to the description and the "Testing" note.
  Proof: rendering `<Portal><div data-state="open"></Portal>` (radix-ui 1.6.7) → `document.querySelectorAll('body > [data-state="open"]').length` = 0, `[data-state="open"]` = 1, chain `DIV < DIV(portal wrapper) < BODY`. The claim "overlay and content portal as direct body children" is false.

**Non-blocking**
- The primary case works: the island sets `aria-expanded` ([`AgentIsland/index.tsx:88`](https://github.com/elyxlz/vesta/blob/1221f3bb49135edbf8a2341d196f7aca8d11ec62/apps/web/src/components/AgentIsland/index.tsx#L88)), so clause 1 root-scoped un-drags every navbar half and dismisses it. Titlebar-triggered dropdowns also work via that clause. Only the portaled cases fall through.
  Proof: `.desktop:has([data-drag-region] [aria-expanded="true"]) [data-drag-region]` matches all halves when expanded.
- Fix is to drop the `>` (descendant `body [data-state="open"]`), but that also matches inline open accordions/collapsibles; scope it deliberately.
  Proof: none, read only.

**Verdict:** BUGGED, the island fix (clause 1) holds, but the two `body > ...` overlay clauses are dead selectors, so the dialog/dropdown dismissal the PR claims never fires.

emipasca:
Verified the blocking finding against this repo's actual components, and it does not hold: the two `body > ...` clauses are live.

The repro rendered the bare `<Portal>` primitive, which does wrap its children in a `Primitive.div`. But Radix's `DialogPortal` and `MenuPortal` render `<PortalPrimitive asChild>` internally, so the dialog overlay/content (with `data-state="open"`) and the popper wrapper are portaled as **direct** children of `body`, no wrapper div.

Proof now in the PR as a regression test (`apps/web/src/drag-regions.test.tsx`, commit 5c5e97aed): renders this repo's `Dialog` and `DropdownMenu` open and asserts `body > [data-state="open"]` and `body > [data-radix-popper-content-wrapper]` each match. Both pass, and the test pins the DOM shape the CSS depends on across Radix upgrades.

The suggested descendant-selector fix (`body [data-state="open"]`) is what would actually break: it matches inline open accordions/collapsibles, which is why the child combinator is scoped to body-level portals in the first place.

PR #2252, opened 2026-08-28
fix(web): reserve the composer height before a reopened chat's first paint

emipasca:
## Problem

Reopening the chat from the navbar (after collapsing it) makes the message bubbles render in a wrong position for one frame, then snap up to their final position.

On desktop, collapsing fully **unmounts** `<Chat>` (`DesktopPanelView.tsx:74`, `{!chatCollapsed && ...}`). Reopening remounts it fresh, with the message history already loaded in the provider. The scroller then positions the bubbles in two passes:

1. **Frame 1** — the initial jump to bottom runs with `composerInset = 0` (the composer measures via an async `ResizeObserver`), so the last bubble lands where the floating composer will sit.
2. **Frame 2** — the composer's height lands, `bottomInset` grows, the list's `paddingBottom` grows, and the re-pin effect (`use-chat-scroll.ts:100-103`) snaps the bubbles up to clear the composer.

It's invisible on a normal first load because the history is still loading (skeleton showing) while the composer measures. On reopen the bubbles paint immediately and race the measurement.

React runs child layout effects before parent ones, so `ChatMessageArea` jumps before `Chat` ever measures the composer: the correct fix is to make the composer's reserved height known at the remounted chat's **first render**.

## Fix

Seed `composerInset`/`composerHeight` from a per-layout baseline instead of `0`:

- A **warm** remount (composer mounted once this session) reads the exact measured height cached in the `use-layout` store → pixel-perfect, no re-pin.
- A **cold** mount (chat kept collapsed across a reload) falls back to `COMPOSER_BASELINE_PX` (a CSS-deterministic constant: 36px button row + 50px pill padding/border + the wrapper's `pb-3`/`pb-4`/`pb-1`) → close enough to avoid a visible snap, and the real measurement overwrites the cache.

One slot per variant (`panel` / `fullscreen` / `mobile`) since their bottom gaps differ.

## Testing

- `tsc` clean, `eslint` 0 errors, 357/357 web unit tests pass.
- Verified live: collapse → reopen no longer snaps the bubbles.

The paint-timing itself isn't observable in jsdom, so there's no unit test for the frame race; the fix is verified visually. The hardcoded fallbacks are a magic-number tradeoff (they can drift if the composer's padding/button size changes), but the cache self-corrects every warm mount regardless.

### DIFF
--- a/apps/web/src/components/Chat/index.tsx
+++ b/apps/web/src/components/Chat/index.tsx
@@ -34,10 +34,11 @@ import { TRIM_HISTORY_SETTLE_MS, agentNeedsUser } from "@vesta/core";
 const COMPOSER_GAP_FULLSCREEN_PX = 12;
 const COMPOSER_GAP_PANEL_PX = 8;
 
-// The empty composer's height per layout, a CSS-deterministic baseline: the 36px
-// button row plus the pill's fixed padding and border (50px), plus the wrapper's
-// bottom gap (pb-3 / pb-4 / pb-1). Seeds a cold remount so the list reserves the
-// composer's space on its first render; the real measurement then overwrites it.
+// The empty composer's height per layout, a CSS-deterministic baseline: the 50px
+// pill (a 36px button row plus its 6+6px padding and 2px border) plus the
+// wrapper's bottom gap (pb-3 12 / pb-4 16 / pb-1 4). Seeds a cold remount so the
+// list reserves the composer's space on its first render; the real measurement
+// then overwrites it.
 const COMPOSER_BASELINE_PX: Record<ComposerVariant, number> = {
   panel: 62,
   fullscreen: 66,
--- a/apps/web/src/components/Chat/index.tsx
+++ b/apps/web/src/components/Chat/index.tsx
@@ -14,7 +14,7 @@ import { Button } from "@/components/ui/button";
 import { Card } from "@/components/ui/card";
 import { CHAT_CONTENT_WIDTH } from "./content-width";
 import { useToast } from "@/stores/use-toast";
-import { useLayout } from "@/stores/use-layout";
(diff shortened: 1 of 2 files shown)


elyxlz:
**Fixes:** nothing linked. Removes the one-frame bubble snap when a desktop chat is reopened after collapse (`DesktopPanelView` unmounts `<Chat>`, so reopen remounts with history already loaded and the composer's reserved height lands a frame late).

**Blocking:** none.

**Non-blocking**
- `Chat/index.tsx:38-42` the cold-mount fallbacks (`panel:62`/`fullscreen:66`/`mobile:54`) are hand-tuned magic numbers that drift if the composer padding or button size changes, and the header comment's own arithmetic ("36px + 50px + pb-3/4/1") does not sum to them. Author acknowledges the tradeoff; the warm cache self-corrects on the first mount this session, so the only cost is a small snap on a chat kept collapsed across a full reload. Same class as today's bug, smaller.
  Proof: `62 != 36 + 50 + 12`; read only, the paint race is not observable in jsdom so no unit test added.

The mechanism holds: a warm remount seeds `composerInset`/`composerHeight` from the exact cached height, the async measure reports the same value, React bails on the identical state, so the `[bottomInset]` re-pin effect (`use-chat-scroll.ts:99-103`) never fires and the initial end-jump already used the correct `scrollHeight`. `isMobile` (line 45) is defined before the variant is computed, the `|| baseline` fallback treats an unset `0` cache correctly (baseline is only written for `height > 0`), and `panel`/`fullscreen` keep separate cache slots so two mounted chats never cross-contaminate. No test references the changed symbols. `MERGEABLE`, one concern.

**Verdict:** MERGE, additive seed that reserves the composer height at first render; attacked the re-pin path and it holds.

---

PR #2258, opened 2026-08-28
fix(desktop): install the app update on relaunch on macOS

emipasca:
## Problem

On macOS, clicking **Relaunch to finish** for an app update just hid the window: the app stayed in the dock and neither relaunched nor quit, so the downloaded update never installed.

## Root cause

The macOS keep-in-dock behavior turns a window close into a hide unless a `quitting` flag is set (`main.ts`):

```ts
mainWindow.on("close", (event) => {
  if (process.platform === "darwin" && !quitting) {
    event.preventDefault();
    mainWindow?.hide();
  }
});
```

That flag flipped **only** on `before-quit`. But `updater.quitAndInstall()` (Electron's native Squirrel.Mac path) emits `before-quit-for-update`, **not** `before-quit`, then closes windows and quits once they are gone. So `quitting` stayed `false`, the close guard hid the window instead of closing it, and the quit-after-windows-closed never completed. Result: window hides, app stays in dock, update never installs.

macOS only: Windows (NSIS) and Linux (`app.relaunch()` + `app.quit()`) have no darwin close guard, so they were unaffected.

## Fix

Track quit intent from both `before-quit` and `before-quit-for-update`, extracted into a small testable `trackQuitIntent` helper (`main.ts` has no unit coverage as the Electron entry point).

## Testing

- `./check.sh app-desktop` green (lint + tsc + vitest, 35 tests, 3 new).
- Windows/Linux verified by code analysis only (I only have a Mac); both paths bypass the darwin close guard, so behavior is unchanged there.

https://claude.ai/code/session_015qdsHPLyP6moGXtPuwdHsu

### DIFF
--- a/apps/desktop/src/lifecycle.test.ts
+++ b/apps/desktop/src/lifecycle.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from "vitest";
+import { trackQuitIntent } from "./lifecycle";
+
+function fakeApp() {
+  const handlers: Record<string, () => void> = {};
+  return {
+    on(event: string, listener: () => void): void {
+      handlers[event] = listener;
+    },
+    emit(event: string): void {
+      handlers[event]?.();
+    },
+  };
+}
+
+describe("trackQuitIntent", () => {
+  it("starts not quitting", () => {
+    const app = fakeApp();
+    const isQuitting = trackQuitIntent(app);
+    expect(isQuitting()).toBe(false);
+  });
+
+  it("flips on before-quit", () => {
+    const app = fakeApp();
+    const isQuitting = trackQuitIntent(app);
+    app.emit("before-quit");
+    expect(isQuitting()).toBe(true);
+  });
+
+  it("flips on before-quit-for-update, so an update relaunch lifts the macOS close guard", () => {
+    const app = fakeApp();
+    const isQuitting = trackQuitIntent(app);
+    app.emit("before-quit-for-update");
+    expect(isQuitting()).toBe(true);
+  });
+});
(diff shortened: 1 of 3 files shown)


elyxlz:
**Fixes:** nothing linked. Solves: macOS "Relaunch to finish" hid the window instead of installing the update, because the darwin close guard only lifted on `before-quit` while `quitAndInstall()` emits `before-quit-for-update`.

**Blocking:** none.

**Non-blocking**
- The fix rests on `before-quit-for-update` firing *before* the update path closes the window, so the flag is set in time to lift the guard at [`main.ts:109`](https://github.com/elyxlz/vesta/blob/94c2a26a9ef6302c89c9e3d06f3cd57fb7f3d9cc/apps/desktop/src/main.ts#L109). That ordering is Electron/Squirrel.Mac native behavior and the documented fix for this class of hide-to-dock-blocks-update bug; it holds on master (still `before-quit`-only at [`main.ts:89`](https://github.com/elyxlz/vesta/blob/2236723e/apps/desktop/src/main.ts#L89)).
  Proof: none, read only. Needs macOS + Electron runtime to exercise; not runnable here (no node deps in tree).
- All three `quitting` readers on master (declare/set/read) are replaced; no dangling reference left; Linux keeps its `app.quit()` path which emits `before-quit`.
  Proof: `git grep quitting origin/master -- apps/desktop/src` → only `main.ts` L15/89/90/112, all covered by the diff.

**Verdict:** MERGE. Small, correct fix for a real macOS update path; the removed `let quitting` and `before-quit` handler map cleanly onto the extracted helper, and the new helper is unit-covered. Confidence limited only by not being able to run Electron on macOS here.

---

PR #2259, opened 2026-08-28
chore(desktop): move the app bundle id into the vesta namespace

emipasca:
## Summary

`appId` in `apps/desktop/electron-builder.yml` moves from `com.vestarun.app` to `com.vesta.desktop`, matching the mobile app's `com.vesta.mobile` namespace.

## Impact on existing installs

- **Auto-update keeps working.** `electron-updater` keys its manifest on `version`, `files`, and `sha512`, never the bundle id, so current installs receive the next release normally.
- **Saved connection survives.** `userData` is keyed on `productName` ("Vesta"), so `connection.json` stays where it is.
- **One-time permission re-grant on macOS.** TCC grants (location, microphone) and the Location Services entry are keyed on the bundle id, so users re-grant them once after this ships.
- **Possible one-time side-by-side install on Windows.** electron-builder derives the NSIS upgrade GUID from `appId` when no explicit `nsis.guid` is set (none is), so the release carrying this may install beside the old copy with a second uninstall entry instead of upgrading in place. Same one-time cost class as the macOS re-grant; `connection.json` survives. Pinning `nsis.guid` to the old derived value would avoid it, at the cost of a config line whose only job is masking the rename; not done here.
- Linux (deb/rpm) does not key on `appId`.

No code reads the bundle id; this is a config-only change.

## Sequencing

Merge this **before** the desktop location PR (#2260): its bundled CoreLocation helper carries an embedded `Info.plist` stamped with `com.vesta.desktop`, and CoreLocation only attributes the helper's grant to the app when the two ids match.

https://claude.ai/code/session_01LuohZkpAcVv1rZjS5CGoVT

### DIFF
--- a/apps/desktop/electron-builder.yml
+++ b/apps/desktop/electron-builder.yml
@@ -1,4 +1,4 @@
-appId: com.vestarun.app
+appId: com.vesta.desktop
 productName: Vesta
 artifactName: "Vesta_${version}_${arch}.${ext}"
 directories:


elyxlz:
**Fixes:** nothing linked. Renames the desktop `appId` from `com.vestarun.app` to `com.vesta.desktop` so it shares the `com.vesta.*` namespace; the real driver is the follow-up location PR, whose CoreLocation helper embeds an `Info.plist` stamped `com.vesta.desktop` and only gets its grant attributed when the two ids match.

**Blocking:** none. `com.vestarun.app` appears nowhere but this line; OAuth is loopback (`127.0.0.1`, `src/oauth-loopback.test.ts`) not a bundle-id protocol, `APP_SCHEME` is an internal SPA scheme, and no entitlements/plist reference the id. The macOS TCC re-grant and updater/userData claims all hold (userData is keyed on `productName` "Vesta").

**Non-blocking**
- [`electron-builder.yml:27`](https://github.com/elyxlz/vesta/blob/e960cf8418e4b610be47598492d8b631f7dea268/apps/desktop/electron-builder.yml#L27) the body says "Windows (NSIS) do not key on `appId`", but electron-builder derives the NSIS upgrade GUID from `appId` when no `guid` is set (none is here), so on the release carrying this change existing Windows installs may land a side-by-side copy / new uninstall entry instead of an in-place upgrade, the same one-time cost class as the disclosed macOS re-grant. `connection.json` still survives (userData keyed on `productName`).
  Proof: none, read only; no explicit `nsis.guid` in the yml, reasoning from electron-builder defaults, could not run a real Windows update.

**Verdict:** MERGE, one-line namespace alignment, nothing reads the old id and the macOS impact is disclosed; only gap is the undisclosed Windows re-install possibility, non-blocking. CI: desktop macOS/Windows/Linux builds still pending; watch those.

---

PR #2260, opened 2026-08-28
fix(desktop): resolve macOS location through a bundled CoreLocation helper

emipasca:
## Problem

No desktop client ever saves a position to the gateway. The macOS cause is confirmed on a packaged, signed build: the renderer's `navigator.geolocation.getCurrentPosition` hangs with no callback at all (its own `timeout` option never fires). vestad and the client frame path are symmetric across platforms and store whatever position a client sends, so the loss is the fix read itself.

Why the renderer path is a dead end on macOS, from Electron's docs and the Chromium source:
- Electron's geolocation defaults to Google's network provider and **ships no API key**, so that provider fails.
- Chromium's built-in CoreLocation provider needs its `GeolocationSystemPermissionManager` to drive the macOS authorization prompt, which Electron does not implement. The prompt never appears and the request stalls. Tracked, unresolved: electron/electron#45290, #46013.

## Fix

macOS now resolves in the **main process**, the same shape Windows (WinRT) and Linux (GeoClue2) already use in `geolocation.ts`:
- `native/vesta-location.swift`: a one-shot CoreLocation helper. It calls `requestWhenInUseAuthorization()` itself (raising the macOS prompt), reads a fix, prints the same invariant `lat|lon|accuracy` line the Windows script does, and exits non-zero with a reason (`denied`, `timeout`, `error: …`) on failure. `parseWindowsFix` covers it unchanged.
- `native/vesta-location.plist`: embedded `Info.plist` stamping the helper with the app's bundle id and usage string, so CoreLocation prompts and grants **as Vesta**, not as a separate entry.
- `scripts/build-native.mjs`: electron-builder `beforePack` hook compiling the helper (macOS builds only, arm64 or universal per build arch) into `Resources`, where electron-builder signs it with the app. Runs on every packaging path, so CI needs no workflow change.
- `geolocation.ts` gains the `darwin` branch; `main.ts` resolves the helper in `Resources` when packaged, `native/` in dev.

## Failures now explain themselves

`runWithTimeout` captures the provider's stderr and `resolveNativeFix` re-raises it instead of collapsing every failure to `null`: the helper's exit reason on macOS, PowerShell's exception on Windows, gdbus's D-Bus error on Linux. `null` now means only "no provider on this platform". The renderer's `readFix` already catches to `null`, so normal behavior is unchanged. Verified with a throwaway Electron app that the message survives the IPC boundary intact.

## TEMPORARY diagnostic, for the release test

Toggling *share this device's location* on runs a probe of both paths and renders the exact per-platform outcome under the toggle (grep `TEMPORARY`). Windows also reports no position, but its code path never touches the renderer (it is WinRT in the main process), so its cause is separate and still unobserved. The probe's `native provider: threw <reason>` line will name it. Remove the probe once both platforms are confirmed.

## Validation

The one piece not verifiable locally is whether macOS attributes the helper's grant to the app: that needs a real Developer-ID-signed build, since ad-hoc signing gives TCC no stable identity. Standalone runs of the helper time out by design (a headless CLI has no app identity to prompt from). Validated on the next official release on macOS and Windows.

Locally: `./check.sh app-desktop` green (37 tests); web `tsc`, `eslint`, `prettier` green; a `--dir` build confirmed the hook compiles the helper and lands it in `Contents/Resources/vesta-location` with the embedded plist.

## Sequencing

**Depends on #2259** (bundle id → `com.vesta.desktop`): the helper's plist carries that id, and CoreLocation only attributes the grant to the app when the two match. Merge #2259 first.

https://claude.ai/code/session_01LuohZkpAcVv1rZjS5CGoVT

### DIFF
--- a/apps/.gitignore
+++ b/apps/.gitignore
@@ -4,6 +4,8 @@ dist/
 
 # Electron build artifacts
 desktop/dist-electron/
+desktop/native/vesta-location
+desktop/native/vesta-location.*64
 
 # Expo mobile build artifacts
 mobile/.expo/
--- a/apps/desktop/electron-builder.yml
+++ b/apps/desktop/electron-builder.yml
@@ -10,7 +10,11 @@ files:
 extraResources:
   - from: ../web/dist
     to: web
+beforePack: scripts/build-native.mjs
 mac:
+  extraResources:
+    - from: native/vesta-location
+      to: vesta-location
   category: public.app-category.productivity
   target:
     - target: dmg
--- a/apps/desktop/native/vesta-location.plist
+++ b/apps/desktop/native/vesta-location.plist
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <key>CFBundleIdentifier</key>
+  <string>com.vesta.desktop</string>
+  <key>CFBundleName</key>
+  <string>Vesta</string>
+  <key>NSLocationWhenInUseUsageDescription</key>
+  <string>Vesta shares this device's location with your agents so they follow your travel.</string>
+</dict>
+</plist>
--- a/apps/desktop/native/vesta-location.swift
+++ b/apps/desktop/native/vesta-location.swift
@@ -7,7 +7,9 @@ import Foundation
 // WinRT on Windows and GeoClue2 on Linux. Any failure exits non-zero with a reason on stderr; the
 // caller falls back to timezone-only.
 
-let TIMEOUT_SECONDS: TimeInterval = 15
+// Shorter than the caller's exec budget (NATIVE_FIX_TIMEOUT_MS in geolocation.ts), so a stall
+// ends with this reason on stderr rather than the caller's SIGTERM.
+let TIMEOUT_SECONDS: TimeInterval = 12
 
 func fail(_ reason: String) -> Never {
   FileHandle.standardError.write(Data("\(reason)\n".utf8))
--- a/apps/desktop/native/vesta-location.swift
+++ b/apps/desktop/native/vesta-location.swift
@@ -1,0 +1,57 @@
+import CoreLocation
+import Foundation
+
+// A one-shot CoreLocation fix for the main process, printed as culture-invariant "lat|lon|accuracy"
+// on stdout (Swift's Double description is always dot-decimal). macOS geolocation in Electron's
+// renderer hangs with no callback, so the app resolves a fix here instead, the same way it shells
+// WinRT on Windows and GeoClue2 on Linux. Any failure exits non-zero with a reason on stderr; the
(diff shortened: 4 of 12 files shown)


elyxlz:
**Fixes:** nothing linked. Solves: desktop clients never persist a position because Electron's macOS renderer `getCurrentPosition` hangs; this moves the fix into a main-process CoreLocation helper, mirroring the existing Windows/Linux pattern. Author states it cannot be confirmed without a Developer-ID-signed build, so the grant-attribution claim is unverified until release.

**Blocking:** none.

**Non-blocking**
- Merge order, not a defect: `native/vesta-location.plist:5` hardcodes `com.vesta.desktop`, but master's `electron-builder.yml` `appId` is `com.vestarun.app`. Per the PR's own premise CoreLocation only attributes the grant when the two match, so merging this before #2259 ships a helper macOS ignores.
  Proof: master `appId: com.vestarun.app`; #2259 diff sets `+appId: com.vesta.desktop`; plist string is `com.vesta.desktop`.
- `geolocation.ts` macOS helper failure (`denied`/`timeout`) rejects, is caught to null in `device-context.ts:60` `readFix`, then falls through to `browserFix()`, the renderer path this PR documents as hanging on macOS. No worse than today, but a clean denial re-enters the hang instead of returning null.
  Proof: `readFix` `.catch(() => null)` then `return browserFix()`; darwin has no bounded resolve there.
- TEMPORARY probe ships to production: duplicate geolocation path + raw per-platform error UI, and toggling on fires both the real report and the probe, so macOS spawns the helper (and its prompt) twice concurrently. Flagged for removal; remove before this is permanent.
  Proof: `DevicesCard/index.tsx` `onToggle` calls `setEnabled(checked)` and `probeGeolocation()`.
- Helper internal `TIMEOUT_SECONDS=15` equals the exec `NATIVE_FIX_TIMEOUT_MS=15000`, so on a stall the exec's SIGTERM likely kills the helper before it prints its clean `timeout` reason, defeating that diagnostic line. Unrelated `main.ts` import/handler reformatting also rides along (CI green).
  Proof: none, read only.

**Verdict:** MERGE, sound and mirrors the Windows/Linux main-process pattern with tests green, but gate the merge on #2259 and strip the TEMPORARY probe before it becomes permanent. `polish-pr` could drop the unrelated `main.ts` churn.

---

emipasca:
Addressed the review in the follow-up commit:

- **Denied re-enters the hang:** fixed at the design level. The renderer's `navigator.geolocation` is no longer a fallback in the desktop app at all: a null or refusing native provider yields timezone-only. Only a plain browser asks `navigator.geolocation`. Pinned by two new tests in `device-context.test.ts` (null bridge and rejecting bridge both never call `getCurrentPosition`).
- **Helper/exec timeout collision:** the helper's internal timeout is now 12s, below the caller's 15s exec budget, so a stall ends with the helper's own `timeout` reason on stderr rather than the caller's SIGTERM.
- **`main.ts` churn:** it is `prettier --write` output, and `check.sh app-desktop` runs `prettier --check` on the file, so reverting it would fail lint. Format-only, left in.
- **TEMPORARY probe / double spawn:** acknowledged and disclosed; both go when the probe is removed after release validation.

Merge order on #2259 confirmed in the PR body.

PR #2269, opened 2026-08-29
feat: secure saved gateway switching

emipasca:
## Summary

- adds a connected mobile gateway switcher for iOS and Android, reachable from Settings
- marks the current gateway and prevents forgetting or reconnecting to the active entry
- encrypts desktop active-session and saved-gateway credentials with Electron safeStorage
- migrates legacy desktop connection files and renderer localStorage records into encrypted main-process storage
- keeps browser storage behavior behind the same native bridge contract

## Validation

- `./check.sh app-desktop`
- `./check.sh app-web`
- `./check.sh app-mobile`
- `./check.sh app-visual guards`

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -10,7 +10,14 @@ import path from "node:path";
 import { readNativeGeolocation } from "./geolocation";
 import { trackQuitIntent } from "./lifecycle";
 import { cancelLoopback, startLoopback } from "./oauth-loopback";
-import { clearConnection, readConnection, writeConnection } from "./store";
+import {
+  clearConnection,
+  clearRecentGateways,
+  readConnection,
+  readRecentGateways,
+  writeConnection,
+  writeRecentGateways,
+} from "./store";
 import { createMainWindow, registerAppScheme, showMainWindow } from "./window";
 
 registerAppScheme();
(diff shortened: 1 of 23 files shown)


elyxlz:
**Fixes:** nothing linked. Feature PR: encrypt desktop connection + saved-gateway credentials with Electron `safeStorage`, add a mobile "Switch gateway" sheet that marks and locks the active entry, and route web recent-gateway storage through the native bridge (async).

**Blocking:** none.

**Non-blocking**
- `apps/web/src/api/server.ts:39` `connectToServer` now `await`s `rememberGateway`, which on desktop reaches `store.ts` `encryptedPayload`, which throws `"secure credential storage is unavailable"` when `safeStorage.isEncryptionAvailable()` is false (Linux desktops with no keyring backend). Previously this wrote plaintext and never threw, so a connect that formerly succeeded now rejects after the session is already set, surfacing a false "connection failed". `connectionStore.write` shares the throw but is fire-and-forget (`void`), so it only degrades to non-persistence.
  Proof: none, read only. Could not confirm `isEncryptionAvailable()` is ever false on a supported desktop (macOS/Windows always true).
- `apps/mobile/app/recent-gateways.tsx:106` `if (status === "connected") router.dismissTo("/")` reads `status` captured at closure-creation, not post-await; harmless here because the switch sheet only opens while already connected, but fragile if reused.
  Proof: none, read only.

**Verdict:** MERGE. Callers all updated to async, preload/`types.ts` parity held, legacy plaintext→encrypted migration and the `LEGACY(remove-when: MIN_SUPPORTED_CLIENT_VERSION exceeds 0.2.13)` marker are sound (repo is 0.2.13); the one caveat is an unproven Linux-keyring edge on an awaited path.

CI: not read here, check the app-desktop/app-web/app-mobile jobs on the PR page.

---

emipasca:
Addressed the review caveat in 6e31a73c8:

- Authentication success no longer depends on saving the active or recent gateway.
- Active-store writes now catch synchronous browser failures and rejected Electron IPC writes.
- Linux desktop storage fails closed unless Electron selected libsecret or a KWallet backend; basic_text and unknown backends are rejected.
- The mobile post-connect dismissal decision is captured explicitly before the async switch.
- Regression tests cover unavailable encryption, Linux backend policy, synchronous storage exceptions, and rejected writes.

Local checks: app-desktop (47 tests), app-web (378 tests), app-mobile (301 tests plus clean prebuild), and guards all pass.

PR #2272, opened 2026-08-29
feat(mobile): add agent rename

emipasca:
## Summary

- add the authenticated agent rename mutation to the shared iOS/Android app
- expose Name from Agent settings and add a native General-screen rename form
- preview the gateway-canonical name, enforce reserved-name rules, block lifecycle conflicts, and follow the returned identity after rename
- add endpoint/model tests and a deterministic rename visual state

## Validation

- ./check.sh app-mobile app-visual guards
- npm -w @vesta/mobile test -- src/agent/settings/agent-name.test.ts src/api/endpoints.test.ts
- npm run mobile:visual:capture -- --device "iPhone 17" --gentle (17/17 flows, 150 screenshots)
- inspected ios/agent-rename.png at original resolution

No version bump.

### DIFF
--- a/apps/mobile/app/agent/[name]/settings.tsx
+++ b/apps/mobile/app/agent/[name]/settings.tsx
@@ -121,6 +121,7 @@ function AgentSettingsContent() {
       />
       <FormGroup>
         <FormSection title="Agent">
+          <FormRow label="Name" value={name} onPress={() => open("general")} />
           <FormRow
             label={sectionTitle("provider")}
             onPress={() => open("provider")}
--- a/apps/mobile/maestro/visual/agent-sections.yml
+++ b/apps/mobile/maestro/visual/agent-sections.yml
@@ -60,6 +60,17 @@ tags:
     file: capture-screenshot.js
     env:
       SCREENSHOT: agent-general.png
+- tapOn:
+    text: "New agent name"
+- eraseText: 4
+- inputText: "Nova_Prime"
+- waitForAnimationToEnd:
+    timeout: 3000
+- takeScreenshot: agent-rename
+- runScript:
+    file: capture-screenshot.js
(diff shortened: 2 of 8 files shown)


elyxlz:
**Fixes:** nothing linked. Adds mobile parity for the agent-rename feature web already has (`RenameCard`); solves a real gap, mobile had no rename.

**Blocking:** none.

**Non-blocking**
- `agent-name.ts:5` reimplements vestad's `normalize_name` in TS with no cross-check, so a change to the Rust normalizer silently diverges the mobile preview from the name the server actually creates.
  Proof: `renameAgent` sends the client-normalized string; vestad re-normalizes at [serve.rs:915](https://github.com/elyxlz/vesta/blob/ea6a1827ae229639317d71eb9485715085061f1f/vestad/src/serve.rs#L915). No test pins the two together. Read only.
- The `!"vesta"` reserved rule is a client-only brand guard; vestad's `validate_name` does not enforce it, so it is UX consistency with web, not a real constraint.
  Proof: [docker.rs:556-580](https://github.com/elyxlz/vesta/blob/ea6a1827ae229639317d71eb9485715085061f1f/vestad/src/docker.rs#L556-L580) has no reserved-word check.

Verified: `PATCH /agents/{name}` body `{new_name}` → `{name}` matches the handler; `agentStatusKind`, `agent.operation`, `Field`/`FormSection`/`Button` props, and the `details/[section]` route all exist. Normalization traced against the four test inputs and vestad, results agree.

**Verdict:** MERGE, a faithful mobile mirror of the existing web rename with matching normalization, lifecycle guard, and follow-the-identity navigation; it held under attack.

---
Maintainers: mention \`@vestabot\` to re-check, run \`polish-pr\`, or fix red CI.

PR #2283, opened 2026-08-29
fix(agent): make the hello the last act of the birth turn

emipasca:
## Problem

On a brand-new agent, birth runs as one non-interruptible boot turn (`collect_boot_turns`, `agent/core/main.py`), and every message the user sends during a boot turn queues until it ends. The birth skill had the agent say hi early (step 4 of "come online") and then run the housekeeping (workspace attach, six skill setups, MEMORY.md edits, seed context) "between replies", inside that same turn. So the user answered the hello and waited, sometimes minutes, while the app showed "waking up".

## Fix

Prompt-only; the boot turn stays non-interruptible.

- `agent/skills/birth/SKILL.md`: the housekeeping moves into the silent "come online" list as step 4, before the hello. Step 5 is the hello, and it ends the turn ("do nothing after the hello"). A lead sentence states the mechanism: nothing the user sends reaches the agent until the turn ends, so the hello is last. "Then meet them" now reads "across the turns their replies open". The "between replies" housekeeping paragraph is gone (it is step 4 now), and the never-reply section drops its duplicate mention.
- `agent/core/prompts/birth.md`: states the boot-turn mechanism the way `restart.md` does: set up in silence, say hi last, end the turn right after.

Per the prompt guide, both texts describe the mechanism and the constraint, never what changed.

## Test plan

- [x] `./check.sh guards`
- No code change, so no unit test; the behavior is the agent's ordering, verified on the next fresh agent's birth.

https://claude.ai/code/session_01FE72XNhoW6qFTDLZ5Hvjtu

### DIFF
--- a/agent/core/prompts/birth.md
+++ b/agent/core/prompts/birth.md
@@ -1,1 +1,3 @@
 This is your first wake. Use the `birth` skill now and follow it exactly.
+
+This first wake is a boot turn: while it runs, the app shows you as still waking up, and every message the user sends queues instead of reaching you. So set up in silence, say hi as the very last thing you do, and end the turn right after the hello. Their reply reaches you as its own turn the moment this one ends.
--- a/agent/skills/birth/SKILL.md
+++ b/agent/skills/birth/SKILL.md
@@ -7,21 +7,22 @@ description: Vesta's first-wake setup and onboarding, run once on a brand-new ag
 
 Hello world. First wake.
 
+This first wake is a boot turn: nothing the user sends reaches you until it ends. So the hello is the last thing you do in it, after every piece of setup below, and you end the turn right after saying hi. Their reply arrives as its own turn.
+
 Come online first, silently, in order:
 1. Read `/run/vestad-env` for ports and token (already exported as env vars). Your name is `$AGENT_NAME`.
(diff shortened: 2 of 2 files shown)


elyxlz:
**Fixes:** nothing linked. Solves a real birth-UX defect: the user's first reply queues behind the whole non-interruptible birth turn, so the app shows "waking up" and the reply waits.

**Blocking:** none.

**Non-blocking**
- `agent/core/main.py:176` docstring says "The one interruptible boot turn is birth", but after-sync migrations can also queue interruptible (`migrations.py` builds an `interruptible=True` batch). Pre-existing simplification the PR extends, not a behavior bug.
  Proof: `git grep -n "interruptible=interruptible" agent/core/migrations.py` → batch turn built with the interruptible flag.
- `agent/core/loops.py:342` `is_notification = interruptible and not is_user` now flips True for birth, so the greeting body logs under `[NOTIFICATION]` instead of `[SYSTEM]`. Log label only; `process_message` gets the same text.
  Proof: none, read only.

**Verdict:** MERGE. Minimal correct change: with `interruptible=first_start`, `noninterruptible_turn_active` is False during birth so `send_preempt` (`client.py:109`) delivers the reply, `boot_turns_pending` excludes birth (`main.py:101`), restart greeting stays non-interruptible. Test pins it: reverting to `interruptible=False` → `test_first_start_birth_turn_is_interruptible` FAILED; 29 passed as-is.

---

PR #2284, opened 2026-08-30
feat(web): dictation and conversation voice modes in the chat composer

emipasca:
## What this does

Polishes the chat composer's voice controls and introduces two explicit voice modes.

**Composer**
- The row is always `[+] [input] [mic] [conversation | send]`. Nothing mounts or unmounts on press, so no button moves under the pointer. The `motion` layout animation and its counter-scaled pill radius are removed (that was the distortion).
- **Mic** starts a dictation. While dictating the two slots become ✕ (discard) and ✓ (confirm and send). The activation setting in Settings > keybinds picks the trigger: `toggle` (click, then ✓) or `hold` (press, release anywhere confirms). Space mirrors the mic.
- **Conversation** (primary) is duplex: each turn sends as it ends, replies are spoken the moment they land (chat pacing is suspended), speaking always cuts a reply, and it ends on the button or after fifteen minutes with no user turn, announced by a toast. Send takes the slot once a draft exists.
- Starting either mode drops typed text. The composer border turns red while recording. The placeholder reads `connecting...`, `listening...`, or `speaking...`.
- The stop-playback button leaves the composer for the chat header, as a mute toggle beside expand and collapse (alone in fullscreen). No animation.

**Store (`use-voice.ts`)**
- `startVoice(mode)`, `stopVoice()` (confirm dictation or end conversation), `cancelVoice()` (discard), `recordingMode: "dictation" | "conversation" | null`, `listening`. The mode is set at press time so a release that beats the microphone prompt still ends the session.
- `registerChat(send, clearInput)` replaces `registerChatCallbacks(send, draft)`; the draft path and the idle timer are gone.

**Transcriber (`lib/voice.ts`)**
- Captures before dialing and buffers up to ~2 s of audio (`MAX_PENDING_AUDIO_BYTES`) until `/stt/listen` opens, so the first words after the press are kept.
- `stop()` during one of `start()`'s awaits now releases what was acquired instead of wiring it up.

**Voice skill (`agent/skills/voice`)**
- `auto_send` is removed at its owner: the setting, its setter, the `/stt/set-auto-send` route, and the legacy `voice_auto_send` migration. Its nested end-of-turn knobs are top-level STT settings now.

Carries over from #2191 (which stays open for its `@vesta/core` model and mobile plumbing, to follow as separate PRs): pacing suspension, forced barge-in, session-state vocabulary, and the pre-open audio buffer.

## Verification

- `./check.sh app-web` (315 tests, 10 new: mode routing, discard, barge-in, inactivity toast, pacing bypass, pre-open buffer flush and cap, right-slot decision)
- `./check.sh guards`
- voice skill CLI pytest (23 passed)


elyxlz:
**Fixes:** nothing linked. Not a `fixes #N` PR: it references open PR #2191 as the broader model it slices from. It polishes the chat composer's voice controls and replaces the `auto_send` toggle with explicit dictation/conversation modes.

**Blocking:** none. Attacked the recording state machine and the `auto_send` removal; both hold. `eot_threshold`/`eot_timeout_ms` read straight from `stt.<key>` (`deepgram.py:157-166` at head@adf8daca), independent of the removed `auto_send`, so hoisting them out of its nested schema strands no stored value; dropping the `voice_auto_send` migration leaves only a harmless orphan key. No mobile or core reader of `auto_send`/the `/stt/set-auto-send` route remains.

**Non-blocking**
- `apps/web/visual/drives/agent-settings.ts:233` still mocks the removed `auto_send` STT setting, so visual QA renders a control that no longer exists.
  Proof: `deepgram.py` schema drops `auto_send` (diff), the drive still lists `key: "auto_send"`.
- `stores/use-voice.ts:254` `startVoice`'s `.catch` sets `transcriber = null` / `recordingMode: null` without the `transcriber !== stream` guard its `.then` has, so a stale start that rejects after a newer session began clobbers the live one. Narrow race (needs mic-denied plus rapid cancel+restart).
  Proof: none, read only.
- Overlaps open #2191 on `use-voice.ts`, `lib/voice.ts`, `voice.test.ts`; whichever lands second needs a rebase.
  Proof: `comm -12` of the two file lists → those three paths.

**CI:** web vitest not run here (Node 18 present, this vitest needs Node 20+); watch `test-app-web`. Voice CLI pytest passed locally (23).

**Verdict:** MERGE, the mode routing, discard, barge-in, and pre-open buffer logic hold and the `auto_send` removal strands no reader; only a stale visual mock and a narrow start-race remain.

---

emipasca:
Polish pass landed as 03e602584 (one commit on top of the author's).

**Pass 1, review (`/code-review high --fix`)**
- `apps/web/src/stores/use-voice.ts`: a transcriber error or a failed `start()` now tears the session down through `cancelVoice` (so `listening`, `liveTranscript`, and the armed inactivity timer reset) and only when that session is still the current one, so a stale session's late failure never clobbers a newer live one. Two tests pin it.
- `apps/web/src/lib/voice.ts`: `Transcriber.stop()` drops an in-flight partial instead of reporting it through `onTurnEnd`, so `cancelVoice`, `_cleanup`, and the inactivity timeout never send half a turn in a conversation. Test added. An em dash in an error string became a colon.
- `apps/web/src/providers/KeybindProvider/index.tsx`: the global Space keydown yields to a focused composer button while a voice mode runs (Space on "discard dictation" was sending), and the hold-mode keyup honors editable targets again.
- `apps/web/src/components/Chat/ChatHeaderActions/index.tsx`: in fullscreen the actions sit below the navbar (`navbarHeight`) instead of under it.

**Pass 2, simplify**
- `right-slot.ts` loses its dead `sttAvailable` branch (`VoiceButtons` already returns the send button there); its test trimmed; the composer's `recording`/`useLiveTranscript` double alias collapsed.
- The inactivity toast derives its minutes from one `CONVERSATION_INACTIVITY_MINUTES` constant.

**Pass 3, prompt sweep**: the changed `agent/` lines are deletions and a flattened settings list with unchanged labels; nothing describes a previous design. No edit.

**Pass 4, AGENTS.md**: nothing the PR changed is described there (only the STT socket and TTS stream in the auth section, both untouched). No edit.

**Reported, not changed (outside the diff or design calls, for the author)**
- The `config`/`config_label` nesting has no producer left after `auto_send` went, but its consumers survive in `base.py`, `handlers.py` (`_merge_setting_values`), `lib/voice.ts`, `VoiceSection` (`SubSettingsCollapsible`), and mobile `types.ts`. Removing it spans files outside the PR.
- `apps/web/visual/drives/agent-settings.ts` still ships the removed `auto_send` setting and lacks the top-level eot numbers.
- Hold-mode release: on iOS Safari touch events keep targeting the detached mic node, so the window `pointerup` may not fire once the mic unmounts. Verify on a device.
- `use-agent-socket.ts` reads the voice store for the duplex flush, a second owner of the pacing decision beside `use-chat-pacing`; mobile's drain has no equivalent.
- `SpeechButton` and `TtsCard` each spell out the optimistic TTS-enable write, and only the header one stops speech.

**Verify**
- `./check.sh app-web`: green (43 files, 317 tests).
- `./check.sh guards`: green.
- `./check.sh agent`: `ty` green; pytest has 82 failures that are all macOS-environmental (no `/proc` for daemon starttime records, bash 3.2 lacking `mapfile`, macOS `df` in the reality check), none in PR files. Substituted `uv run --project skills/voice/cli pytest skills/voice/cli/tests/`: 23 passed.

PR #2285, opened 2026-08-30
feat(web): instant notification popover, fuller dialog, honest bell dot

emipasca:
## Why

Three things about the bell were wrong.

1. **Opening the popover was slow.** The first open started a `GET /notifications` fetch *and* held its skeletons for a minimum 500ms, so the surface took most of a second to show anything.
2. **Both surfaces showed the wrong amount.** The popover rendered the entire unseen set; the "see all" dialog truncated every notification to a single line.
3. **The dot lied.** It was derived from the gateway's two watermark scalars alone, so a message from the agent whose page you have open raised it, even though the pill deliberately hides that message because the chat renders it. The dot then stayed up until you opened a history surface.

## What

**Instant open.** `useNotificationFeed` prefetches the newest page as soon as the controller exists, before any surface opens, so the popover renders from memory. With a prefetch in place the `minLoadingMs` hold has nothing left to protect against (skeletons now show only while a fetch is genuinely in flight), so it is deleted along with its `Promise.allSettled` dance and the web provider's constant.

**The popover is a taster.** At most four of the newest unseen rows, with "see all" carrying the rest. It no longer needs its own scroll region.

**The dialog reads in full.** The title on its own line, the body wrapping below it up to three lines, more row spacing, and the alternating row bands the backups list uses.

**The dot counts only what is truly new.** The feed records, at arrival, which rows the user watched land in the chat (`notificationReadInPlace`: a `message` from the viewed agent, on a focused window), and the new `feedUnseen` discounts them. The decision is taken at arrival, not at render, so navigating away does not resurrect the dot. Anything the client cannot account for keeps the dot: an unloaded feed, or a gateway stamp above every row it holds. Nothing leaves the history, and the gateway's watermark is untouched, so no other notification can be masked.

## Verification

- `./check.sh app-core`, `app-web`, `app-desktop`, `app-visual`, `guards`: green.
- New tests: the prefetch (`use-notification-feed.test.tsx`), the read-in-place rule at arrival including the unfocused and other-page cases, and four `feedUnseen` cases covering the discount, the alongside-unread case, and both cannot-account-for fallbacks.
- Two new visual scenarios, `notifications-popover` and `notifications-dialog`, capture both surfaces on every web platform; captured and inspected on `web` and `web-narrow`.

### DIFF
--- a/apps/core/src/index.ts
+++ b/apps/core/src/index.ts
@@ -57,6 +57,7 @@ export {
   EMPTY_FEED,
   feedNeedsMarkSeen,
   feedSections,
+  feedUnseen,
   feedView,
   reduceFeed,
   type FeedAction,
--- a/apps/core/src/notifications-pill/notification-feed.test.ts
+++ b/apps/core/src/notifications-pill/notification-feed.test.ts
@@ -3,6 +3,7 @@ import {
   EMPTY_FEED,
   feedNeedsMarkSeen,
   feedSections,
+  feedUnseen,
   feedView,
   reduceFeed,
   type FeedAction,
@@ -27,9 +28,9 @@ function ids(feed: NotificationFeed): number[] {
 describe("reduceFeed: the live edge", () => {
   it("lands a live arrival at the top, once, and remembers it as live", () => {
     const feed = run([
-      { type: "arrived", entry: entry(5) },
-      { type: "arrived", entry: entry(5) },
-      { type: "arrived", entry: entry(6) },
+      { type: "arrived", readInPlace: false, entry: entry(5) },
+      { type: "arrived", readInPlace: false, entry: entry(5) },
+      { type: "arrived", readInPlace: false, entry: entry(6) },
     ])
     expect(ids(feed)).toEqual([6, 5])
     expect(feed.liveIds).toEqual([5, 6])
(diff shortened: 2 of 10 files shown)


elyxlz:
**Fixes:** nothing linked. Removes the popover's slow first open (a `GET /notifications` plus a forced 500ms skeleton hold), caps the popover at four rows with "see all", and gives the dialog a full title+body layout.

**Blocking:** none.

**Non-blocking**
- The prefetch (`use-notification-feed.ts:71`) fires a `GET /notifications` for every user on mount, even those who never open the bell. Intended trade for instant open, and it's one small page, but worth a maintainer's nod.
  Proof: none, read only.
- `NotificationsPill/index.tsx:210` `archiveExtendsBeyond` returns true whenever `sections` exists, so with unseen>0 but no seen rows and older exhausted, "see all" shows a dialog identical to the popover. Pre-existing on master (same `if (sections) return true`), not introduced here.
  Proof: `git show origin/master:apps/web/src/components/Navbar/NotificationsPill/index.tsx` has the same branch.

**Verdict:** MERGE. Prefetch, cap, and dialog layout all check out; the new prefetch test genuinely pins the behavior.
Proof: disabled the prefetch effect in a head worktree, `npx vitest run use-notification-feed.test.tsx` → 1 failed ("prefetches the newest page before any surface opens"); restored → 6 passed, notifications-pill suite 27 passed. No other reader of the removed `minLoadingMs`.

---

PR #2295, opened 2026-08-30
feat: app chat attachments

emipasca:
Users can send any file to their agent through app chat, WhatsApp-style, and the agent can send files back. Web and desktop ship here; the wire contract and the client engine are shared so mobile reuses everything except the view layer.

Design spec: `docs/superpowers/specs/2026-08-30-app-chat-attachments-design.md` (committed on this branch). Built as five reviewed commits, one per surface:

## app-chat skill (`e7d9399d`)
The skill owns the whole attachment plane. An offset-addressed chunked upload (create session with mime validation, `PUT data?offset=` with 409 resync, a status probe, idempotent complete), a Range-capable streamed serve route (inline for media mimes only, CSP sandbox, RFC 5987 dispositions, a 410 removed state), attachment metadata embedded on `user`/`chat` events, the notification `attachments` line carrying absolute paths and sizes the agent opens directly, `app-chat send --attach`, an `app-chat attachments list|rm` disk-management CLI, and an activity-aged periodic GC sweep. Zero `agent/core/` changes, zero vestad changes, no `min_supported` bump.

## @vesta/core (`1999a75d`)
`ChatAttachment` wire metadata, kind derivation and byte formatting with one owner, and the chunked upload engine built for spotty connections: adaptive chunk sizing, 409 resync via the status probe, unbounded classified retries with backoff, offline parking on the connectivity edge, per-chunk stall timeout, and run-level abort. Plus the pure draft reducer and the extended send-message/optimistic-bubble seams.

## Web composer (`6d5d477f`)
The Plus button's attach popover, paste, and a contained per-chat drag-drop overlay feed per-agent draft chips (progress ring, waiting-for-network, error with retry, name and size always visible, totals footer). Send gates on every draft being uploaded; a bubble parked in the retryable state re-posts itself once per reconnect edge.

## Bubbles, viewer, downloads (`0ef06f35`)
Typed bubbles per kind (pre-sized images with skeleton/broken/removed states, inline video with an expand handoff, audio, download tiles with determinate progress), the portal-less in-chat viewer (contained scrim, cursor-anchored wheel/pinch zoom with a tested clamp model, double-click toggle), header-authed blob downloads, and the desktop `will-download` handler.

## Review pass (`898ace09`)
Seven finder angles over the diff; every confirmed finding fixed (stale-token retry loop, eviction-orphaned uploads, voice sends dropping visible chips, markdown spacing, probe efficiency, and a set of one-owner consolidations).

The merge commit re-weaves the composer surfaces onto #2284's dictation/conversation row (AttachMenu in the add slot, chips as drafts for `rightSlot`, voice sends through the drafts wrapper).

Gallery scenarios cover the menu, all chip states, the drop overlay, every bubble kind, the removed/broken states, and the viewer fitted and zoomed, captured and inspected across `web`, `desktop`, and `web-narrow` in both themes.


elyxlz:
**Fixes:** nothing linked. Adds app-chat file attachments (send + receive) on web/desktop, shared engine so mobile reuses all but the view. No `agent/core` or `vestad` changes, so the additive `attachments` field on the chat `StoredEvent` is covered by ignore-unknown and needs no `min_supported` bump: correct.

**Blocking**
- [`attachments.py:65`](https://github.com/elyxlz/vesta/blob/0166a47e2a575642132622c91faf5dad700d9cc3/agent/skills/app-chat/cli/src/app_chat_cli/attachments.py#L60-L68) `sanitize_filename` runs `lstrip(".")` then `.strip()`, so a leading "space-dot" re-exposes a dot and defeats the documented anti-clobber invariant (lines 21-22). `finalize` then renames the upload onto a control file: a file named `. .meta.json` gets overwritten by the metadata JSON (served under the declared mime), and `. .session.json` gets unlinked and serves 410. The user's uploaded bytes are corrupted or lost.
  Proof: ran the real store, `create_session(…, '. .meta.json', 5)` + append `HELLO` + `finalize` → `blob_path` reads `b'{"id":…,"name":".meta.json"…}'`, not `HELLO`; the `. .session.json` variant → `is_removed` True. Name is client input on `POST /attachments`.

**Non-blocking**
- [`test_attachments.py:200`](https://github.com/elyxlz/vesta/blob/0166a47e2a575642132622c91faf5dad700d9cc3/agent/skills/app-chat/cli/tests/test_attachments.py#L200-L210) asserts the anti-clobber invariant but only tests names a single `lstrip` already fixes, so all 117 tests pass while the whitespace variant corrupts. Add the `. .meta.json` case alongside the fix (strip whitespace and dots to a fixed point).
  Proof: test loop covers `".meta.json"`/`".session.json"`, never `". .meta.json"`.

**Verdict:** BUGGED, a proven data-loss path in filename sanitization contradicts its own invariant; two-line fix (loop strip-then-lstrip until stable) plus the missing test case. Everything else held: TS upload/retry/abort/zoom/blob-lifecycle and the serve-route inline-mime/CSP/traversal/size boundaries all survived attack.

---

PR #2302, opened 2026-08-30
refactor(core,web): extract the voice session model into @vesta/core

emipasca:
## What this does

Extracts the platform-neutral voice model into `@vesta/core` so mobile can share it, migrating web onto it with no behavior change. This is the shared-model half of the follow-up to #2284 (the composer voice modes), sliced from #2191 with today's dictation/conversation semantics.

## The shared model (`apps/core/src/voice/`)

- `stt-session.ts`: the `/stt/listen` socket lifecycle, the Deepgram TurnInfo protocol, transcript accumulation, and the pre-open audio buffer (~2 s, kept until the socket opens then flushed in order). Clients inject only `buildUrl`, a socket constructor, and an `AudioCapture` port.
- `tts-queue.ts`: ordered playback over an injected `SpeechPlayer` (prepare then play), with epoch invalidation on stop and a prefetch cache warmed during the typing-pacing delay.
- `voice-session.ts`: composes the two and owns the mode decisions. **Dictation** accumulates turns and sends one message on confirm (keeping an in-flight partial the user sees), or sends nothing on cancel. **Conversation** sends each turn as it ends, always cuts a reply on a user turn (barge-in forced), and ends on stop or an inactivity window; dictation honors `interrupt_tts` and arms no timer.

21 core unit tests cover queue order, epoch invalidation, prefetch reuse, the buffer window and cap, turn parsing, accumulation, per-mode barge-in, dictation confirm vs cancel, conversation per-turn send, inactivity, listening state, and error paths.

## Web

`use-voice.ts` becomes a thin zustand view over `createVoiceSession` (about 110 fewer lines); `lib/voice.ts` keeps the HTTP fetchers and now exposes the browser ports (`browserCapture`, `browserSocket`, `browserPlayer`, `voiceWsUrl`), losing the hand-rolled `Transcriber` class and TTS drain loop. The store's integration test still drives the whole path through fake `Audio`/`WebSocket`/`AudioContext` globals; the Transcriber buffering unit tests moved to core's `stt-session.test.ts`.

## Verification

- `./check.sh app-core` (21 new voice tests), `./check.sh app-web` (store integration suite green), `./check.sh app-desktop`, `./check.sh guards`

Follow-up: mobile conversation mode plus hands-free native plumbing, migrated onto this model, fixing the mobile TTS playback-error path flagged on #2191.


elyxlz:
**Fixes:** nothing linked. This is a refactor: it extracts the platform-neutral voice session model (STT socket, TTS queue, mode decisions) into `@vesta/core` so mobile can share it, migrating web onto it with no intended behavior change.

**Blocking:** none.

**Non-blocking**
- [`apps/web/src/stores/use-voice.ts:254`](https://github.com/elyxlz/vesta/blob/64bdafec39ced752028df11da7e340a083914f3a/apps/web/src/stores/use-voice.ts#L254) `export { prepareSpeech }` re-exports a symbol no file imports; on master it was used in the body, now it only exists to keep the import from going unused. Dead re-export, drop the import instead.
  Proof: `grep -rn prepareSpeech apps/web apps/desktop` → only the def, the in-lib call, and this import+re-export; nothing consumes it from the store.
- [`use-voice.ts:164`](https://github.com/elyxlz/vesta/blob/64bdafec39ced752028df11da7e340a083914f3a/apps/web/src/stores/use-voice.ts#L164) `speak`/`prefetch` dropped the `!agentName` guard; a null agent now reaches `browserPlayer.prepare`, which rejects and surfaces "Voice playback failed" instead of a silent return. Practically unreachable (`speechEnabled` implies a fetched per-agent `ttsStatus`), so noted not blocked.
  Proof: none, read only.

**Verdict:** MERGE. Careful port with the ownership boundary held (generation/epoch guards, `stt === session` re-entrancy checks sound); core 21 + web store 17 tests pass, and both suites drive the real session layer through fakes.
Proof: `vitest run src/voice` → 21 passed; web `vitest run src/stores/use-voice` → 17 passed, on head@64bdafec.

**CI:** app-core / app-web / app-desktop pending on the PR page; nothing in the diff should fail them.

---

PR #2303, opened 2026-08-30
feat(mobile): dictation and hands-free conversation voice modes

emipasca:
## What this does

Brings mobile voice to parity with web (#2284) on the shared core model (#2302): the mic dictates into the composer with confirm/discard, and a conversation button opens a hands-free duplex session. Ports the native plumbing and the panel from #2191, adapted to the current core API and with the playback-error wedge fixed.

## Adapter (`useLiveVoice.ts`)

Rewritten as the mobile adapter over `createVoiceSession`, with expo-audio `useAudioStream` capture and `useAudioPlayer` playback as the ports and one TTS queue (consecutive replies play back to back; a user turn cuts the reply). Replaces the hand-rolled STT socket and the separate `useSpeechPlayer`.

- **Dictation**: the mic accumulates turns into the composer; confirm sends one message, discard drops it.
- **Conversation**: each turn sends as it ends, barge-in is forced, and the session holds the native hands-free setup around the mic (below). Ends on the panel's End button or after fifteen minutes of silence, announced to the user.

## Native hands-free (`mobile/modules/`)

- **iOS** (`vesta-audio-session`, Swift): a `playAndRecord` / `voiceChat` audio session (echo cancellation, Bluetooth A2DP routing), kept alive with the screen locked via `UIBackgroundModes: ["audio"]`.
- **Android** (`vesta-voice-service`, Kotlin): a `microphone`-type foreground service, the only way Android keeps a locked-screen mic alive, with the `FOREGROUND_SERVICE`/`FOREGROUND_SERVICE_MICROPHONE` permissions. Both activate only in conversation mode and tear down when it ends.

## UI

- The composer's right cluster: mic (dictation) plus send when there is a draft or a conversation button when there is not; dictation swaps the row for discard and confirm.
- A docked `VoiceConversationPanel` (state orb, live transcript, large End) replaces the composer while a conversation runs.

## The playback-error fix

The mobile play port settles the TTS queue's `await` on a stream that finishes, aborts, or drops out of the loaded state, so one failed reply never leaves the queue wedged (the blocking issue flagged on #2191).

## Verification

- `./check.sh app-mobile`: tsc, 298 vitest tests, lint, and the clean-prebuild verify (both `ios` and `android` native projects generate with the local modules autolinked and manifests merged).
- `./check.sh guards`.
- Android/iOS native compile runs in CI (`mobile-android` / `mobile-ios`); this machine lacks a JDK for the local Gradle step, so the Android prebuild + manifest merge was verified locally and the Kotlin compile is left to CI.

**Needs a device pass before merge** (cannot be exercised here): the panel look, iOS Bluetooth routing and echo cancellation, the Android foreground service and locked-screen mic, and mic permission prompts. Android conversation captures without echo cancellation, so a loud speaker may self-interrupt there.

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -56,6 +56,8 @@ const config: ExpoConfig = {
       ITSAppUsesNonExemptEncryption: false,
       NSCameraUsageDescription: "Scan a Vesta connection QR code.",
       NSMicrophoneUsageDescription: "Talk to your Vesta agent.",
+      // Keeps a hands-free voice conversation alive with the screen locked.
+      UIBackgroundModes: ["audio"],
     },
   },
   android: {
@@ -70,6 +72,10 @@ const config: ExpoConfig = {
       "android.permission.CAMERA",
       "android.permission.RECORD_AUDIO",
       "android.permission.POST_NOTIFICATIONS",
+      // The microphone foreground service that keeps a hands-free voice conversation alive
+      // while the app is backgrounded or the screen is locked.
+      "android.permission.FOREGROUND_SERVICE",
+      "android.permission.FOREGROUND_SERVICE_MICROPHONE",
     ],
     intentFilters: [
       {
--- a/apps/mobile/modules/vesta-audio-session/ios/VestaAudioSessionModule.swift
+++ b/apps/mobile/modules/vesta-audio-session/ios/VestaAudioSessionModule.swift
@@ -13,6 +13,23 @@ public final class VestaAudioSessionModule: Module {
         .setAllowHapticsAndSystemSoundsDuringRecording(enabled)
     }
 
+    // Hands-free voice: playAndRecord with voice processing (echo cancellation)
+    // and Bluetooth routing, so a session works over a car or headset and
+    // playback does not re-enter the microphone as a fake user turn.
+    AsyncFunction("setHandsFreeSessionActiveAsync") { (active: Bool) in
+      let session = AVAudioSession.sharedInstance()
+      if active {
+        try session.setCategory(
+          .playAndRecord,
+          mode: .voiceChat,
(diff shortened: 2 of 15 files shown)


elyxlz:
**Fixes:** nothing linked. Brings mobile voice to parity with web: mic dictation into the composer (confirm/discard) plus a hands-free duplex conversation, over the shared `@vesta/core` voice session. Also carries the #2191 playback-wedge fix.

**Blocking:** none. No dangling refs to the removed `useSpeechPlayer`/`ComposerActionButton`/`voice.active`, teardown is idempotent, native modules and manifests are well-formed, no banned TS patterns.

**Non-blocking**
- [`useLiveVoice.ts:179`](https://github.com/elyxlz/vesta/blob/1696addda91543eb66e5889a73e98fbd800e8e11/apps/mobile/src/voice/useLiveVoice.ts#L166-L188) settles the TTS queue await only on `didJustFinish` or `played && !isLoaded`. A stream that errors before it ever loads keeps `played=false`, so neither branch fires and that reply wedges the queue. Narrow (prepare already returned an id), but a wider miss than the "loaded-then-dropped" case it targets.
  Proof: none, read only.
- Dictation confirm sends the STT capture via core `onSend`, not `input`, so a manual edit to the composer during dictation is dropped on send.
  Proof: none, read only.
- Native and locked-screen paths (iOS Bluetooth/echo-cancel, Android mic foreground service, `startForegroundService` needs minSdk≥26) can't be exercised here; the author flags a device pass is required.
  Proof: none, read only.

**CI:** not run here. The real gate is `mobile-android`/`mobile-ios` native compile and `app-mobile` vitest (author verified prebuild locally, left Kotlin compile to CI); watch those.

**Verdict:** MERGE, the adapter/UI logic held under review and the wiring is clean, but treat the acknowledged device QA and the native CI jobs as the merge gate, not this read.

---

PR #2305, opened 2026-08-30
fix(desktop): grant microphone under the macOS hardened runtime

emipasca:
## What this does

Voice input never captured on the packaged macOS desktop app. Root cause: `electron-builder.yml` sets `hardenedRuntime: true` but ships no custom entitlements, so electron-builder's default entitlements apply, which contain only the JIT and library-validation keys, **not** `com.apple.security.device.audio-input`. Under the hardened runtime, macOS blocks the microphone without that entitlement, whatever the renderer permission handler (already allows `media`) or the Info.plist `NSMicrophoneUsageDescription` say.

## The fix

- Add `apps/desktop/build/entitlements.mac.plist`: the three Electron hardened-runtime defaults plus `com.apple.security.device.audio-input` and `com.apple.security.device.camera` (the connection QR scanner). Point `mac.entitlements` and `mac.entitlementsInherit` at it.
- In the main process, request the OS microphone grant via `systemPreferences.askForMediaAccess("microphone")` when the renderer asks for `media`, so the TCC prompt appears when voice is first used. The permission decision is a pure `rendererPermissionDecision` helper with a unit test.

## Verification

- `./check.sh app-desktop` (tsc, lint, vitest incl. the new permission-decision test) and `./check.sh guards`.
- Ran the dev Electron app: voice captures, and the media path/permission handler are exercised.
- Confirmed the entitlement embeds by codesigning the packed app with the new file: `codesign -d --entitlements` shows `com.apple.security.device.audio-input` and `camera`. electron-builder applies `mac.entitlements` the same way in a signed build.
- The full end-to-end gate is a microphone test on the next **signed + notarized** release build (the entitlement only takes effect there; dev runs unsigned).

### DIFF
--- a/apps/desktop/build/entitlements.mac.plist
+++ b/apps/desktop/build/entitlements.mac.plist
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <!-- Electron's hardened-runtime defaults (JIT + library validation). -->
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
+    <!-- Voice input: the hardened runtime blocks the microphone without this, whatever the
+         TCC grant or the renderer permission handler say. -->
+    <key>com.apple.security.device.audio-input</key>
+    <true/>
+    <!-- The connection QR scanner reads the camera. -->
+    <key>com.apple.security.device.camera</key>
+    <true/>
+  </dict>
+</plist>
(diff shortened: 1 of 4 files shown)


elyxlz:
**Fixes:** nothing linked. Solves voice input not captured on the packaged macOS desktop app: under `hardenedRuntime: true` the app shipped no `com.apple.security.device.audio-input` entitlement, so macOS blocked the mic.

**Blocking:** none.

**Non-blocking**
- `window.ts:88` treats every `media` request as microphone: on darwin it calls `askForMediaAccess("microphone")` and gates the grant on the mic result, ignoring `details.mediaTypes`. A camera (video) `media` request would prompt for the wrong device and be denied when only the mic is refused.
  Proof: only `apps/web/src/lib/voice.ts:279` calls getUserMedia (`audio` only); no video getUserMedia exists in the desktop renderer, so this is latent, not hit today. Read only.

**Verdict:** MERGE. Entitlements file matches electron-builder's installed default template (`app-builder-lib/templates/entitlements.mac.plist`, exactly the 3 keys the PR calls out) plus the two device keys; `entitlements`/`entitlementsInherit` both point at it, matching the default which uses one template for both. The `rendererPermissionDecision` mapping matches its test and the standing "media + geolocation only" comment. The real gate (mic under a signed+notarized build) cannot run here; dev is unsigned so the entitlement is a no-op, and the author flags this honestly.

CI: pending on the PR page; not evaluated here.

---

emipasca:
Folded into #2308 (single voice PR with the web mute/TTS-gate/space-to-dictate changes). The desktop mic commit is included there.

PR #2308, opened 2026-08-30
feat(web): voice mute, TTS-gated conversation, space-to-dictate + macOS desktop mic

emipasca:
## What this does

A batch of voice refinements in one PR (web chat voice + the macOS desktop mic fix).

### Web chat voice

- **Local mute** (header speaker button): now a per-device mute of spoken replies, not a provider toggle. Shown only when text-to-speech is enabled, so turning TTS off in settings hides the control (previously nothing changed in the chat UI). Red while muted. A conversation always speaks back and ignores the mute; the mute only silences the ambient read-aloud of replies outside a conversation.
- **TTS-gated conversation**: starting a conversation needs text-to-speech, so the conversation button now refuses when TTS is off or unconfigured and toasts the fix, "enable text-to-speech in the settings" or "set up text-to-speech in the settings", instead of opening a silent session. The button still shows either way.
- **Space to dictate from the empty composer**: pressing Space in the composer when it is empty starts dictation without typing a space; a composer with text, or any other focused field, types the space as before. The composer textarea is marked so this never fires in unrelated inputs.

### macOS desktop microphone

Folded in from the desktop mic fix (the packaged app never captured audio): a custom `entitlements.mac.plist` adds `com.apple.security.device.audio-input` (and camera) under the hardened runtime, wired via `mac.entitlements`/`entitlementsInherit`, plus `systemPreferences.askForMediaAccess` for the OS grant. Verified the entitlement embeds under codesign.

## Verification

- `./check.sh app-web` (voice store tests updated; new tests: the two conversation TTS-gate toasts, and mute silencing an ambient reply while a conversation speaks anyway), `./check.sh app-desktop`, `./check.sh guards`.
- Desktop dev-app run confirmed mic capture; entitlement embedding confirmed via `codesign -d --entitlements`. The desktop end-to-end gate remains a mic test on the next signed release.

Supersedes #2305 (the desktop mic fix is included here). Mobile voice parity stays separate in #2303, which is blocked on a device pass.

### DIFF
--- a/apps/desktop/build/entitlements.mac.plist
+++ b/apps/desktop/build/entitlements.mac.plist
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <!-- Electron's hardened-runtime defaults (JIT + library validation). -->
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
+    <!-- Voice input: the hardened runtime blocks the microphone without this, whatever the
+         TCC grant or the renderer permission handler say. -->
+    <key>com.apple.security.device.audio-input</key>
+    <true/>
+    <!-- The connection QR scanner reads the camera. -->
+    <key>com.apple.security.device.camera</key>
+    <true/>
+  </dict>
+</plist>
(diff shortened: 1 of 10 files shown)


elyxlz:
**Fixes:** nothing linked (supersedes #2305, mobile parity #2303 stays separate). Solves two real problems: the packaged macOS app never got the mic TCC grant under the hardened runtime, and the chat voice control conflated a per-device mute with the provider TTS toggle.

**Blocking:** none.

**Non-blocking**
- [`window.ts:88`](https://github.com/elyxlz/vesta/blob/3daac3c3fc7b0ccfdadf4f867b9b8762bdf05efa/apps/desktop/src/window.ts#L88) always calls `askForMediaAccess("microphone")` for any `media` request, and the new camera entitlement ships with no `NSCameraUsageDescription`; if the QR/camera path is ever wired on desktop it prompts for the wrong device and crashes on macOS for the missing usage string.
  Proof: none, read only; no `getUserMedia` camera call exists in `apps/web/src`, so the path is currently unreachable.
- Custom `entitlements.mac.plist` correctly re-adds the three electron-builder JIT defaults a custom file would otherwise drop.
  Proof: none, read only.

**Verdict:** MERGE, the mute/TTS-gate/space-dictate logic and the mic entitlement hold up under trace; Apps·web and Apps·desktop CI green.

---

PR #2310, opened 2026-08-30
test(web): lock that muted TTS fires no pipeline

emipasca:
## What this does

Confirms and locks the behavior you asked about: when the user mutes text-to-speech, the pipeline does not fire and get dropped at the speaker, it never fires at all.

The store already gates `speak` and `prefetch` before the core TTS queue (`speakable` returns false when muted outside a conversation), and that queue is the only caller of `player.prepare`/`player.play`. So a muted reply reaches neither the `/tts/prepare` POST nor a streamed `<audio>`.

This pins it: the strengthened test asserts the mocked prepare call is **not** made and **no** audio element is created while muted, so a future change cannot regress it into fire-but-silent. A conversation still speaks while muted (kept as its own case).

No production code changes; behavior was already correct.

## Verification

- `./check.sh app-web`, `./check.sh guards`.

### DIFF
--- a/apps/web/src/stores/use-voice.test.ts
+++ b/apps/web/src/stores/use-voice.test.ts
@@ -452,12 +452,20 @@ describe("recording modes", () => {
     useToastStore.getState().dismiss();
   });
 
-  it("mute silences an ambient reply but a conversation speaks anyway", async () => {
+  it("mute fires no TTS pipeline at all: no prepare, no stream", async () => {
     useVoice.setState({ muted: true });
+    apiJsonMock.mockClear();
+    useVoice.getState().prefetch("would warm");
     useVoice.getState().speak("silent while idle");
     await new Promise((resolve) => setTimeout(resolve, 0));
+    // The gate is at the store's entry, before the queue: a muted reply reaches neither the
+    // prepare POST nor a streamed <audio>, rather than preparing and staying silent.
+    expect(apiJsonMock).not.toHaveBeenCalled();
     expect(FakeAudio.created).toHaveLength(0);
+  });
 
+  it("a conversation speaks even while muted", async () => {
+    useVoice.setState({ muted: true });
     await startRecording("conversation");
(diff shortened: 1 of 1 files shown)


elyxlz:
**Fixes:** nothing linked. Strengthens an existing test so a future change cannot regress muted TTS from "never fires" into "fires but stays silent at the speaker".

**Blocking:** none.

**Non-blocking**
- `use-voice.test.ts:445` (head) `apiJsonMock.mockClear()` is dead: nothing in this test calls `apiJson` before it, and the top-level `beforeEach` already `mockReset()`s it. Harmless, drop it in a polish pass.
  Proof: [head@ef4ad3d](https://github.com/elyxlz/vesta/blob/ef4ad3d14b9827f0d4720b9bfc4269b2f72d8b7e/apps/web/src/stores/use-voice.test.ts#L445-L446), no `apiJson` caller precedes the clear.
- Could not run the suite here (Node 18.19.1; vitest/rolldown needs `node:util.styleText` from Node 20.12+). CI `test-app-web` is the check to watch.
  Proof: none, read only.

The gate the test pins is real: `speakable` returns false when muted outside a conversation, and both `prefetch` and `speak` early-return on it before `session.*` (which is the only path to the prepare POST / `<audio>`). The sibling happy-path and "conversation speaks while muted" tests prove the mock wiring, so the new `not.toHaveBeenCalled()` genuinely locks the guarantee.
Proof: gate at [use-voice.ts:123-129](https://github.com/elyxlz/vesta/blob/ef4ad3d14b9827f0d4720b9bfc4269b2f72d8b7e/apps/web/src/stores/use-voice.ts#L123-L129), callers at [L229-L235](https://github.com/elyxlz/vesta/blob/ef4ad3d14b9827f0d4720b9bfc4269b2f72d8b7e/apps/web/src/stores/use-voice.ts#L229-L235).

**Verdict:** MERGE, test-only change that correctly locks a real behavior; description is honest that no production code moved.

---

PR #2311, opened 2026-08-31
feat(web,core,mobile): conversation drawer UI and yield-to-user turn gate

emipasca:
## What

Two halves of one conversation-mode overhaul:

### Conversation drawer UI (web)
- The composer card morphs into a conversation surface: a Vesta orb driven by the real agent status with extra motion configs (minimal listening pulse, obvious talking pulse), and an action bar of mic-mute, a state pill carrying the live transcript (2 lines max, latest word bold), and an icon exit button.
- The chat recedes behind a scrim with a perspective transform, locks to the bottom, trims history, and skips the fake typing delay while a conversation runs; the bottom-inset reservation rides a CSS-variable transform during the morph so the whole transition stays off the layout path (120fps-clean).
- Agent settings gain a Voice section (dictation + conversation cards), `interrupt_tts` moves out of the provider card, and the provider gets its own tab.

### Yield-to-user turn gate (core + agent + mobile)
Vesta never talks over the user, while the user can still barge in:
- **Client** (`@vesta/core` voice-session): a yielding conversation reports the user's live turn (`onUserSpeakingChange`) and holds replies off the speaker until the turn ends; the STT session now closes empty turns so the gate can never stick.
- **Wire** (chat socket): a `{"type": "speaking", "active": bool}` inbound frame, cached and replayed on reconnect. Old daemons ignore it, old clients send none — fully additive.
- **Agent** (app-chat skill): the daemon ties the flag to the connection that set it (no timers; a dropped client clears it) and refuses `app-chat send` mid-turn with "the user is talking right now: drop this reply, wait for their next message, then answer the whole thought", so Vesta answers the full thought once instead of half of it. SKILL.md documents the contract.
- **Settings**: "never interrupt you" toggle in the conversation card, per device, on by default. Mobile is wired through the same core callback (always yields; its toggle lands with the mobile polish pass).

## Checks
- `./check.sh app-core` 466 ✓, `app-web` 319 ✓, `app-desktop` ✓, `app-mobile` ✓ (incl. clean-prebuild), `guards` ✓
- `./check.sh app-chat` 58/58 ✓ (six daemon tests need a short pytest basetemp on macOS due to the AF_UNIX path limit — pre-existing, they fail identically on master locally and pass on CI's Linux)

### DIFF
--- a/agent/skills/app-chat/SKILL.md
+++ b/agent/skills/app-chat/SKILL.md
@@ -65,6 +65,7 @@ app-chat attachments rm <id> [<id>...] # frees the bytes, keeps the chat history
 
 ## Notes
 - Always reply to app messages using `app-chat send`, not through any other channel
+- In a live voice conversation, `send` is refused with `the user is talking right now: ...` while the user is mid-sentence. Do what the error says: drop that reply, wait for the next `source=app-chat` notification (it arrives when they finish), and answer their whole thought in one fresh reply
 - `send` enforces short-bubble texting: a wall (over ~220 chars, or any text after a full stop) is rejected so you re-send as several short calls, one thought each. Don't use full stops at all: a `.`, `!` or `?` may only close a bubble, never carry text after it. Ellipses stay free, they're a beat rather than a stop. For genuine reference material the user asked for (a brief, a code block, a list), pass `--longform` to bypass
 - A numbered or bulleted list is fine to send as one message (each item is one short thought); a line-leading marker like `1.` or `2)` is not a full stop, so a list does not need `--longform`
 - Send multiple short messages instead of one long one (like texting)
--- a/agent/skills/app-chat/cli/src/app_chat_cli/daemon.py
+++ b/agent/skills/app-chat/cli/src/app_chat_cli/daemon.py
@@ -246,12 +246,12 @@ async def _ingest_attachments(state: DaemonState, attach: list[str]) -> tuple[li
 
 async def _handle_send(state: DaemonState, message: str, attach: list[str]) -> tuple[dict[str, object], str | None]:
     """One validated send: gate on the user's live turn, ingest attachments, persist + emit.
-    Returns the ack and the toast text (None when nothing went out)."""
+    Returns the response envelope (ok or error) and the toast text (None when nothing went out)."""
     if not message and not attach:
         return {"error": "empty message"}, None
-    if state.service.user_speaking():
-        # A reply now would talk over the user; the error tells the model the next move.
-        return {"error": "the user is talking right now: drop this reply, wait for their next message, then answer the whole thought"}, None
+    refusal = state.service.refuse_send_while_speaking()
+    if refusal is not None:
+        return {"error": refusal}, None
     metas, ingest_error = await _ingest_attachments(state, attach)
     if ingest_error is not None:
         return {"error": ingest_error}, None
--- a/agent/skills/app-chat/cli/src/app_chat_cli/daemon.py
+++ b/agent/skills/app-chat/cli/src/app_chat_cli/daemon.py
@@ -244,6 +244,26 @@ async def _ingest_attachments(state: DaemonState, attach: list[str]) -> tuple[li
     return metas, None
 
 
+async def _handle_send(state: DaemonState, message: str, attach: list[str]) -> tuple[dict[str, object], str | None]:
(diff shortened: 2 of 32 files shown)


elyxlz:
**Fixes:** nothing linked. Ships the conversation-drawer UI plus a yield-to-user gate: the app-chat daemon refuses `app-chat send` while the user is mid-turn, and the client holds/replays the speaking flag. Additive wire frame, old daemons/clients degrade cleanly.

**Blocking:** none. The gate logic held: reconnect replay only re-sends a live `true`, disconnect clears the flag daemon-side, empty turns still close the floor, and held replies drop on session end.

**Non-blocking**
- PR is CONFLICTING (DIRTY). It branched from `5f20cc8`, before [`d6f1d18` feat: app chat attachments (#2295)](https://github.com/elyxlz/vesta/commit/d6f1d18d) landed on master. Its [`ChatComposer`](https://github.com/elyxlz/vesta/blob/d485e78312fe174d5cb11c1e89899b44088bad30/apps/web/src/components/Chat/ChatComposer/index.tsx#L340-L358) still carries the pre-#2295 dead `<Plus/>` placeholder (not introduced here, just relocated); a take-ours conflict resolution silently reverts the attachment feature. Rebase and keep master's `<AttachMenu onFiles={attachments.addFiles}/>`.
  Proof: `git merge origin/master` → CONFLICT in ChatComposer/index.tsx, Chat/index.tsx, mobile useAgentSocket.ts; merge-base ChatComposer has the bare `add attachment` button, `origin/master:296` has `<AttachMenu .../>`.
- Turn end reports `speaking:false` over the chat WS, then sends the transcript via HTTP intake; if intake beats the WS frame the agent's first `send` is refused once and retried. No data loss (message durable).
  Proof: none, read only.
- CI mostly pending (guards green); nothing red yet. Watch Apps · web/core/mobile and Agent typecheck.

**Verdict:** MERGE, the gate and reconnect logic survived attack (dropping the daemon gate reddens `test_send_is_refused_while_the_user_is_talking`); only a rebase onto #2295 stands between it and merge, and the resolver must not let the stale ChatComposer clobber attachments.

---

emipasca:
Polish pass (review → simplify → prompt sweep → docs freshness):

**Review** (confirmed + fixed)
- Switching agents mid-conversation now cancels the live voice session, so transcripts and speaking reports can never route to the wrong agent's daemon (+ regression test).
- A muted mic streams zeroed frames instead of dropping them, so Deepgram still closes the turn and the speaking gate can't wedge.
- The daemon repays a refused reply: if the blocking turn produces no message (empty transcript, dropped socket), the floor clearing writes a `user_finished_talking` interrupt notification, keeping the refusal's "a notification follows" promise on every path.
- Vestigial `morphed` machinery deleted (its enter delay was 0, so it sequenced nothing); comments corrected to the real mechanism.
- `sheetEase` extracted to `lib/motion`; `aria-pressed` on the activation segmented control; `_handle_send` docstring fixed.

**Simplify**
- One `_emit_notification` writer owns the notification-file contract; one `refuse_send_while_speaking` method owns the whole send gate (refusal text, marker, re-wake) in `service.py`.
- `morphing` React state collapsed to a ref + `onMorphSettled` callback (two dead re-renders per morph gone).
- FLIP measurement gated on the wrap flip (`layoutDependency`), so typing renders skip motion's layout pass.
- Shared `CONVERSATION_FADE` and `recedeTransition` constants; `RECORDING_BUTTON` hoisted next to the panel.

**Prompt sweep**: the SKILL.md refusal bullet checked against final behavior; accurate, no edit.

**Flagged, not changed**
- Permanent `will-change` layers on the receding chat: deliberate pre-promotion tradeoff, worth author eyes.
- Morph-stick knife-edge (measured height exactly 264 at the flip skips the settle callback): low-probability, self-healing, restructuring risks the hand-tuned timing.
- `DictationCard` hand-rolls its `interrupt_tts` write beside `DynamicSettings`' generic path: unifying is a settings-surface refactor beyond polish.
- AGENTS.md: whether the yield-to-user gate deserves a sentence in the Message flow paragraph is borderline; nothing existing became false, so left to the author.

Verify: app-core, app-web, app-desktop, app-mobile, guards, app-chat CLI green locally (one macOS-only `/tmp`-symlink failure reproduced on master); the agent suite's Linux-runtime tiers fail identically on master here and pass in CI.

PR #2304, opened 2026-08-30
refactor: move provider setup into the agent

emipasca:
## Summary

- Move provider catalog data, OAuth/setup sessions, model discovery, and key validation into the agent.
- Fold the catalog into `GET /provider` and expose personality metadata independently at `GET /personalities`.
- Remove provider-specific code and `/manifest` from vestad, leaving generic agent-scoped provider and personality relays.
- Create the agent shell before provider setup, then keep onboarding behind dedicated loading states until the selected agent is explicitly `alive` with `booting: false`.
- Invalidate cached boot completion whenever the agent tap disconnects, preventing stale readiness from exposing chat or dashboard during a new boot.
- Move the breaking compatibility floor to `0.2.16`.

## Design

Provider setup now lives with the runtime and catalog it configures. vestad remains responsible for lifecycle, authentication, routing, and fleet state, without knowing provider-specific OAuth or model contracts.

Personality selection is a separate agent-owned resource. Preset frontmatter is the source of truth, including the single default preset.

The web flow is shell-first:

1. Create the named agent.
2. Wait for its API surface.
3. Complete provider and personality setup through that agent.
4. Apply configuration and restart.
5. Enter the normal agent home only after exact current-boot readiness.

The navigation guard applies only to the agent currently being onboarded, so existing agents remain usable.

## Validation

- `./check.sh guards`
- `./check.sh vestad`: 495 passed, 35 ignored
- `./check.sh app-core`: 456 passed
- `./check.sh app-web`: 324 passed
- `./check.sh app-mobile`: 300 passed, clean Expo prebuild
- `./check.sh app-desktop`: 57 passed
- `./check.sh app-visual`: 58 passed
- `./check.sh integration`: 6 multi-user, 4 OAuth, and 66 server tests passed
- `./check.sh agent`: 1,172 passed, 87 skipped, plus every shipped skill suite

PR #2320, opened 2026-08-31
test(vestad): stop the harness pooling keep-alive connections

emipasca:
## Problem

The release pipeline for v0.2.16 flaked twice on the vestad e2e test `auth::provider_setup_and_catalogs_are_owned_by_the_named_agent` with `io: Peer disconnected`, at two different lines across the two attempts.

## Root cause

The integration harness `Client` (`vestad/tests-integration/src/client.rs`) uses one pooled `ureq::Agent` shared across every call in a test. ureq keeps idle keep-alive connections for a default 15s and reuses them. When it reuses a connection the gateway (hyper/axum) has meanwhile closed, it surfaces `io: Peer disconnected` instead of transparently reconnecting; for a POST it does not retry, so the test panics.

#2304 rewrote this test from two trivial direct gateway calls into one that creates an agent, polls `wait_until_running` for up to 180s, then fires ~8 relay calls on the shared pooled agent. That made it the first test with enough calls and wall-clock to hit the stale-reuse window.

It is a latent harness flake, not a relay regression: one of the two failures landed on `GET /manifest` with no auth, a route that 404s purely in the gateway and never reaches the agent, which pins the failure to the client-to-gateway pooled connection.

## Fix

Disable idle connection pooling in the test client (`max_idle_connections_per_host(0)`). The harness makes sequential blocking calls, so pooling buys nothing but this race; a fresh connection per request removes it at the layer that owns it. Test-only change, no product code touched.

Verified: `cargo check -p vesta-tests` passes. The full `server` suite is Docker+Linux-gated, so behavioral verification is CI-only.

### DIFF
--- a/vestad/tests-integration/src/client.rs
+++ b/vestad/tests-integration/src/client.rs
@@ -124,6 +124,10 @@ impl Client {
         };
         let agent = ureq::Agent::config_builder()
             .http_status_as_error(false)
+            // No keep-alive reuse: the harness makes sequential blocking calls, so a pooled
+            // connection the gateway has since closed only surfaces as `io: Peer disconnected`
+            // on the next reuse. A fresh connection per request removes that race.
+            .max_idle_connections_per_host(0)
             .tls_config(tls_config)
             .build()
             .new_agent();


elyxlz:
**Fixes:** nothing linked. Solves a real release-gate flake: v0.2.16 flaked twice on `auth::provider_setup_and_catalogs_are_owned_by_the_named_agent` with `io: Peer disconnected`.

**Blocking:** none.

**Non-blocking**
- Behavioral fix is unverifiable here: the `server` suite is Docker+Linux-gated, so I cannot reproduce the stale-reuse race or confirm the 20x-green stability the pipeline needs. The root cause is sound though: `1c0731e8` (#2304) rewrote this test to `wait_until_running` for 180s then fire ~8 proxy calls on the one shared pooled `ureq::Agent`, and one recorded failure hit `proxy_get("/manifest", None)`, a gateway-only 404 route that never reaches the agent, pinning the fault to the client↔gateway pooled connection.
  Proof: none, read only; CI (`test-vestad-docker`) is the settling signal to watch.

Fix is at the layer that owns the pool, `max_idle_connections_per_host(0)` is a valid ureq 3.4.0 knob (default 3), no test depends on keep-alive reuse, and sequential blocking calls lose nothing from a fresh connection each. Comment block is 3 lines, no product code touched.

**Verdict:** MERGE — correct layer, minimal, single concern; watch `test-vestad-docker` for the behavioral confirmation this diff can only get in CI.

---

PR #2321, opened 2026-08-31
fix(app-chat): stop the reply waterfall when the user starts talking

emipasca:
## What

In a live voice conversation, a stray bubble still landed **right as the user stopped speaking**, because the agent's reply went out as many independent `app-chat send` calls paced by shell `sleep`. The daemon's speaking gate is checked per call and reopens the instant the user stops, so a later bubble of an already-interrupted reply slipped through: the trailing bubble in the window after the floor cleared but before the agent's fresh reply, or a bubble that landed during a brief pause. The gate could refuse one bubble but never recall the in-flight shell loop.

Found by retro'ing `eleven`'s logs on a real conversation: the first bubble of a reply was refused (`the user is talking right now`), then a later bubble of the same reply still persisted and was spoken.

## Fix

Move the waterfall into the CLI so it can be stopped as a unit:

- **`app-chat send` takes the whole reply** as repeated `-m` bubbles. The CLI sends them in order with a beat (`--gap` overrides the default), and the moment one is refused because the user is talking it drops the rest and returns `stopped_for_user`. The reply drops as a unit, so no stray bubble lands as the user stops. The daemon re-wakes the agent with the existing `user_finished_talking` notification when the floor clears.
- **Pre-lint all bubbles up front**, so a malformed bubble never leaves a half-sent reply.
- **Daemon:** one additive line, the refusal envelope carries `user_speaking` so the CLI distinguishes a yielded floor from a real error. The per-call gate and the turn-end re-wake are unchanged.
- Single `-m` still works, so nothing regresses. No client change, no wire-contract change.

No reply-id needed: sequential send + stop-on-refusal means there is no independent tail to poison.

## Testing

- `./check.sh app-chat` → 125 passed (4 new: pacing+order, stop-on-talk, pre-lint-all, refusal marker).
- `pytest tests/test_daemon_contract.py -k app-chat` → 12 passed (real daemon lifecycle intact).
- `./check.sh guards` → all checks passed.

## Out of scope

The client still flushes held bubbles at `onTurnEnd` (`voice-session.ts`) instead of dropping them. With this fix the residual is tiny (a single bubble emitted in the sub-second before the daemon sees `speaking:true`); a follow-up can flip flush to drop if wanted.

### DIFF
--- a/agent/skills/app-chat/SKILL.md
+++ b/agent/skills/app-chat/SKILL.md
@@ -29,6 +29,7 @@ server (intake, history, and the live `/ws` chat socket).
 ```bash
 app-chat daemon status
 app-chat send --message 'Hello!'
+app-chat send -m "hey" -m "one thought per -m" -m "they go out as separate bubbles, paced like texting"
 app-chat send --attach ~/out/chart.png --message 'here it is!'
 app-chat history --search 'query'
 app-chat history --limit 20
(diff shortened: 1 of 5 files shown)


elyxlz:
**Fixes:** nothing linked (no closing keyword). Solves the stray-voice-bubble race: a reply's later bubbles slipping through the per-call speaking gate after the user interrupts. The move from shell-paced independent `send` calls to one interruptible CLI waterfall does close that window as claimed (residual sub-second bubble honestly noted as out of scope).

**Blocking:** none.

**Non-blocking**
- A multi-bubble reply is now a single foreground `app-chat send` that blocks the agent turn for `gap`*(n-1) seconds (default 2.5s each), so a non-voice text preempt only lands after the whole waterfall drains, where separate calls used to yield between bubbles. By design for texting pace, and the daemon gate still stops the voice case; noting the latency trade only.
  Proof: none, read only.

**Verdict:** MERGE, attacked the stop path and it held. `_send_bubbles` returns `stopped_for_user` on the `user_speaking` envelope and drops the tail; reverting that line reddens `test_cmd_send_stops_the_waterfall_when_the_user_starts_talking` (`1 failed`), and `uv run pytest` in the head worktree → `125 passed`. `--message -` heredoc reply path and the notification `reply_command` still work; no wire-contract reader of the changed stdout shape.

---

